PR #20, opened 2018-10-18
xmpp: add support for authentication

saghul:


saghul:
@damencho can you PTAL?

PR #23, opened 2018-10-29
Make JVB DAEMON_OPTS Configurable

lsiv568:
This PR updates the `DAEMON_OPTS` passed the the JVB configurable by the user via the environment variable `JVB_DAEMON_OPTS`.  

If `JVB_DAEMON_OPTS` environment variable is present, it will be used, otherwise, the default opts will be used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,7 @@ Variable | Description | Default value
 `JVB_STUN_SERVERS` | STUN servers used to discover the server's public IP | stun.l.google.com:19302, stun1.l.google.com:19302, stun2.l.google.com:19302
 `JVB_PORT` | Port for media used by Jitsi Videobridge | 10000
 `JVB_BREWERY_MUC` | MUC name for the JVB pool | jvbbrewery
+`JVB_ENABLE_APIS` | Comma separated list of JVB APIs to enable | none
 `JIGASI_XMPP_USER` | XMPP user for Jigasi MUC client connections | jigasi
 `JIGASI_XMPP_PASSWORD` | XMPP password for Jigasi MUC client connections | passw0rd
 `JIGASI_BREWERY_MUC` | MUC name for the Jigasi pool | jigasibrewery
--- a/env.example
+++ b/env.example
@@ -77,6 +77,10 @@ JVB_STUN_SERVERS=stun.l.google.com:19302,stun1.l.google.com:19302,stun2.l.google
 # Media port for the Jitsi Videobridge
 JVB_PORT=10000
 
+# A comma separated list of APIs to enable when the JVB is started. The default is none.
+# See https://github.com/jitsi/jitsi-videobridge/blob/master/doc/rest.md for more information
+#JVB_ENABLE_APIS=rest,colibri
+
 # XMPP component password for Jicofo.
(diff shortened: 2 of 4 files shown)


saghul:
Hi @lsiv568 thanks for the patch! What are you trying to configure exactly? (I'm trying to understand if this is the right approach or we can do things in a different way).

Also, once we agree on the approach, I'd need you to update the README and add some documentation on this.

lsiv568:
@saghul I'm trying ti enable the private REST API.  The reason I want to do this is so I can communicate with the colibri REST API.  In order to do this, I need to supply the the `--apis=rest` option.  

Instead of hardcoding this, I thought it would be ideal to make the value dynamic/configurable so one could add the following to the JVB environment section within `docker-compose.yml`:
`- JVB_DAEMON_OPTS=--apis=rest`.

If we agree on this approach, I'd be more than happy to update the documentation to explain this.

saghul:
I think I'd go for a explicit option instead: JVB_ENABLE_REST_API or maybe JVB_ENABLE_APIS with a comma separated list of them (default to none).

lsiv568:
For this particular use case a more explicit option makes sense. I was attempting to provide a bit more flexibility/future proofing in the event additional opts were ever wanted.  However, that's not always the best approach. I will update the PR to reflect this approach.

lsiv568:
@saghul on second thought, if a user want's to supply any other options such as `--host, --secret` etc.. the explict option approach doesn't scale that well with respect to implementation.  Thoughts?

lsiv568:
Updated this PR to introduce a `JVB_APIS_TO_ENABLE` environment variable which can be used to enable apis when the JVB is started.  For example, the REST api as documented here https://github.com/jitsi/jitsi-videobridge/blob/master/doc/rest.md

README.md
```diff
@@ -140,6 +140,7 @@ Variable | Description | Default value
 `JVB_STUN_SERVERS` | STUN servers used to discover the server's public IP | stun.l.google.com:19302, stun1.l.google.com:19302, stun2.l.google.com:19302
 `JVB_PORT` | Port for media used by Jitsi Videobridge | 10000
 `JVB_BREWERY_MUC` | MUC name for the JVB pool | jvbbrewery
+`JVB_APIS_TO_ENABLE` | Comma separated list of JVB APIs to enable | --apis=none
```
saghul: Please let's call this JVB_ENABLE_APIS.

env.example
```diff
@@ -77,6 +77,10 @@ JVB_STUN_SERVERS=stun.l.google.com:19302,stun1.l.google.com:19302,stun2.l.google
 # Media port for the Jitsi Videobridge
 JVB_PORT=10000
 
+# A comma separated list of APIs to enable when the JVB is started. The default is none.
+# See https://github.com/jitsi/jitsi-videobridge/blob/master/doc/rest.md for more information
+#JVB_APIS_TO_ENABLE=--apis=rest,colibri
```
saghul: I don't want the user to specify the `--apis=` part, they should just say `none` or `rest,colibri`.

saghul:
Sorry for the delay. I left some comments in the PR.

lsiv568:
@saghul I have updated this PR to address your comments.

Issue #16, opened 2018-10-05
[noob] Unable to get the docker image working


nailyk-fr:
Hi. 

Thanks for providing a docker image template. 
However, I am not able to make it working. 

# The setup: 
Running the image on a dedicated server with public ip on eth0. 
Having apache2 as proxy for both http(s) ports. 

After some `dock exec -it <id> cat/etc/passwd` I was able to set the right perms on some folder. For further references: 
```
root@khiksistigma :) # ls -aln 
total 32
drwxr-xr-x  8 1000 1000 4096 oct.   5 17:26 .
drwxrwxr-x 10 1000 1000 4096 oct.   5 16:26 ..
drwx------  2  999 1000 4096 oct.   5 16:51 jicofo
drwx------  2  999 1000 4096 oct.   5 17:38 jvb
drwxr-xr-x  4 1000 1000 4096 oct.   5 16:51 meet
drwxr-xr-x  5  101  102 4096 oct.   5 16:51 prosody
drwxr-xr-x  2  101  102 4096 oct.   5 17:31 prosody_var
drwxr-xr-x  4    0    0 4096 oct.   5 17:15 web
```
After reading other issues, am not running docker as root. 
[.env](https://github.com/jitsi/docker-jitsi-meet/files/2451518/env.txt)
[docker-compose.yml.txt](https://github.com/jitsi/docker-jitsi-meet/files/2451531/docker-compose.yml.txt)
As I had no idea which value tu setup on `DOCKER_HOST_ADDRESS` I put the public ip address. 

# the problem: 
After the `docker-compose up` everything seems to be working fine. Not so much errors in logs. Opening a webpage on jitsi.nailyk.fr seems to be working. However, when someone attempt to open the same link to join, the both webpages crashes and the log display `SEVERE: Can not invite participant -- no bridge available.`

Here is the list of opened ports: 
```
CONTAINER ID        IMAGE                       COMMAND                  CREATED             STATUS              PORTS                                                                                            NAMES
d97e2e1f4afa        jitsi/jvb                   "/init"                  5 seconds ago       Up 4 seconds        51.15.17.68:10000->10000/udp                                                                     docker-jitsi-meet_jvb_1_f812e25b8c76
a10386ce5886        jitsi/prosody               "/init"                  23 minutes ago      Up 3 seconds        5222/tcp, 5269/tcp, 5280-5281/tcp, 5347/tcp                                                      docker-jitsi-meet_prosody_1_740326e9447f
[...]

PR #25, opened 2018-11-03
README: Add chmod 777 notice

nailyk-fr:
* Next to https://github.com/jitsi/docker-jitsi-meet/issues/16 the problem was the base folder permissions. 
 * `chmod 777` on it solve the issue.


saghul:
Thanks for taking the time to write the PR, but this is not something I'd like to encourage in the README. The default Just Works, as long as you are not root.

pide2000:
For me, jitsi only works if i have set the whole config folder to 777. I don't want to do this, but I have no idea how to solve this problem.  @saghul , you say "the default just works" but what is "the default"? I dont think I make something special.

PR #28, opened 2018-11-17
Run cron only when HTTPS and Let's Encrypt are enabled

sapkra:


web/rootfs/etc/services.d/cron/run
```diff
@@ -1,3 +1,7 @@
 #!/usr/bin/with-contenv bash
 
-exec cron -f
+if [[ $DISABLE_HTTPS -ne 1 ]]; then
```
saghul: Let's simplify this and only check for Let's Encrypt being enabled. Nice catch!

sapkra: But then it's not consistent anymore. Right now I'm using the same conditions like we are using in `10-config` for copying the script into the `cron.monthly` directory.

PR #29, opened 2018-11-18
Add JWT support

sapkra:
With this PR I'm adding support for JWT tokens to the prosody container.
I don't wanted to copy the prosody modules out of the `jitsi/jitsi-meet` repository so that I added it as a submodule to have no duplicates and one single point of truth.

I hope I didn't missed something.

@amirasaran Thanks for your help.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -54,6 +54,11 @@ services:
             - JVB_AUTH_PASSWORD
             - JIGASI_XMPP_USER
             - JIGASI_XMPP_PASSWORD
+            - JWT_ENABLE_TOKEN_AUTH
+            - JWT_APP_ID
+            - JWT_APP_SECRET
+            - JWT_ACCEPTED_ISSUERS
+            - JWT_ACCEPTED_AUDIENCES
             - TZ
         networks:
             meet.jitsi:
--- a/prosody/Dockerfile
+++ b/prosody/Dockerfile
@@ -1,10 +1,24 @@
 FROM jitsi/base
 
+ADD https://raw.githubusercontent.com/jitsi/jitsi-meet/fc129d9849ca5e26245d54df6451931b6c179987/resources/prosody-plugins/token/util.lib.lua /prosody-plugins/token/util.lib.lua
+ADD https://raw.githubusercontent.com/jitsi/jitsi-meet/fc129d9849ca5e26245d54df6451931b6c179987/resources/prosody-plugins/mod_token_verification.lua /prosody-plugins/mod_token_verification.lua
+ADD https://raw.githubusercontent.com/jitsi/jitsi-meet/fc129d9849ca5e26245d54df6451931b6c179987/resources/prosody-plugins/mod_auth_token.lua /prosody-plugins/mod_auth_token.lua
+
+RUN sed -i s/hook/hook_global/g /prosody-plugins/mod_auth_token.lua
+
 RUN \
 	apt-dpkg-wrap apt-get update && \
+	apt-dpkg-wrap apt-get install -y lua5.2 liblua5.2-dev libssl1.0-dev lua-basexx luarocks gcc git && \
 	apt-dpkg-wrap apt-get install -t stretch-backports -y prosody && \
-	apt-cleanup && \
 	rm -rf /etc/prosody
+
+RUN \
+	luarocks install lua-cjson 2.1.0-1 && \
+	luarocks install luajwtjitsi
+
+RUN \
+	apt-dpkg-wrap apt-get remove -y liblua5.2-dev libssl1.0-dev gcc git && \
+	apt-cleanup
 
 COPY rootfs/ /
 
--- a/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
+++ b/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
@@ -1,13 +1,28 @@
 admins = { "{{ .Env.JICOFO_AUTH_USER }}@{{ .Env.XMPP_AUTH_DOMAIN }}" }
-plugin_paths = { "/prosody-plugins-custom" }
+plugin_paths = { "/prosody-plugins/", "/prosody-plugins-custom" }
 http_default_host = "{{ .Env.XMPP_DOMAIN }}"
 
+{{ if and (.Env.ENABLE_AUTH | default "0" | toBool) (.Env.JWT_ENABLE_TOKEN_AUTH | default "0" | toBool) .Env.JWT_ACCEPTED_ISSUERS }}
+asap_accepted_issuers = { "{{ join "\",\"" (splitList "," .Env.JWT_ACCEPTED_ISSUERS) }}" }
+{{ end }}
+
+{{ if and (.Env.ENABLE_AUTH | default "0" | toBool) (.Env.JWT_ENABLE_TOKEN_AUTH | default "0" | toBool) .Env.JWT_ACCEPTED_AUDIENCES }}
+asap_accepted_audiences = { "{{ join "\",\"" (splitList "," .Env.JWT_ACCEPTED_AUDIENCES) }}" }
+{{ end }}
+
 VirtualHost "{{ .Env.XMPP_DOMAIN }}"
-    {{ if .Env.ENABLE_AUTH | default "0" | toBool }}
-    authentication = "internal_plain"
+{{ if .Env.ENABLE_AUTH | default "0" | toBool }}
+    {{ if .Env.JWT_ENABLE_TOKEN_AUTH | default "0" | toBool }}
+    authentication = "token"
+    app_id = "{{ .Env.JWT_APP_ID }}"
+    app_secret = "{{ .Env.JWT_APP_SECRET }}"
+    allow_empty_token = false
     {{ else }}
-    authentication = "anonymous"
+    authentication = "internal_plain"
     {{ end }}
+{{ else }}
+    authentication = "anonymous"
+{{ end }}
     ssl = {
         key = "/config/certs/{{ .Env.XMPP_DOMAIN }}.key";
         certificate = "/config/certs/{{ .Env.XMPP_DOMAIN }}.crt";
@@ -52,8 +67,10 @@ Component "{{ .Env.XMPP_MUC_DOMAIN }}" "muc"
         {{ if .Env.XMPP_MUC_MODULES }}
         "{{ join "\";\n\"" (splitList "," .Env.XMPP_MUC_MODULES) }}";
         {{ end }}
+        {{ if .Env.JWT_ENABLE_TOKEN_AUTH | default "0" | toBool }}
+        "token_verification";
+        {{ end }}
     }
 
 Component "focus.{{ .Env.XMPP_DOMAIN }}"
     component_secret = "{{ .Env.JICOFO_COMPONENT_SECRET }}"
-
--- a/README.md
+++ b/README.md
@@ -140,6 +140,17 @@ Once in the container, run the following command to create a user:
 
 ``prosodyctl --config /config/prosody.cfg.lua register user meet.jitsi password``
 
+#### Authentication using JWT tokens
+You can also use JWT tokens to authenticate users. To enable it you have to enable authentication via both `ENABLE_AUTH` & `JWT_ENABLE_TOKEN_AUTH` environment variables and configure the settings you can see below.
+
+Variable | Description | Example
+--- | --- | ---
+`JWT_ENABLE_TOKEN_AUTH` | Enable authentication via JWT tokens | 1
+`JWT_APP_ID` | Application identifier | my_jitsi_app_id
+`JWT_APP_SECRET` | Application secret known only to your token | my_jitsi_app_secret
+`JWT_ACCEPTED_ISSUERS` | (Optional) Set asap_accepted_issuers as a comma separated list | my_web_client,my_app_client
+`JWT_ACCEPTED_AUDIENCES` | (Optional) Set asap_accepted_audiences as a comma separated list | my_server1,my_server2
+
 ### Advanced configuration
 
 These configuration options are already set and generally don't need to be changed.
--- a/env.example
+++ b/env.example
@@ -62,6 +62,21 @@ TZ=Europe/Amsterdam
 # Enable guest access.
 #ENABLE_GUESTS=1
 
+# Enable authentication via JWT tokens.
+#JWT_ENABLE_TOKEN_AUTH=1
+
+# Application identifier.
+#JWT_APP_ID=my_jitsi_app_id
+
+# Application secret known only to your token.
+#JWT_APP_SECRET=my_jitsi_app_secret
+
+# (Optional) Set asap_accepted_issuers as a comma separated list.
+#JWT_ACCEPTED_ISSUERS=my_web_client,my_app_client
+
+# (Optional) Set asap_accepted_audiences as a comma separated list.
+#JWT_ACCEPTED_AUDIENCES=my_server1,my_server2
+
 #
 # Advanced configuration options (you generally don't need to change these)
 #
--- a/prosody/rootfs/etc/cont-init.d/10-config
+++ b/prosody/rootfs/etc/cont-init.d/10-config
@@ -10,6 +10,14 @@ if [[ "$(stat -c %U /config)" != "prosody" ]]; then
     chown -R prosody /config
 fi
 
+if [[ "$(stat -c %U /prosody-plugins)" != "prosody" ]]; then
+    chown -R prosody /prosody-plugins
+fi
+
+if [[ "$(stat -c %U /prosody-plugins-custom)" != "prosody" ]]; then
+    chown -R prosody /prosody-plugins-custom
+fi
+
 if [[ ! -f $PROSODY_CFG ]]; then
     cp -r /defaults/* /config
     tpl /defaults/conf.d/jitsi-meet.cfg.lua > /config/conf.d/jitsi-meet.cfg.lua


prosody/Dockerfile
```diff
@@ -2,11 +2,19 @@ FROM jitsi/base
...
 
 COPY rootfs/ /
+COPY git-modules/jitsi-meet/resources/prosody-plugins /prosody-plugins
+
+RUN \
+	luarocks list --outdated && \
+	luarocks install luaossl  CRYPTO_LIBDIR=/usr/lib/x86_64-linux-gnu  CRYPTO_INCDIR=/usr/include OPENSSL_LIBDIR=/usr/lib/x86_64-linux-gnu OPENSSL_INCDIR=/usr/include && \
+	luarocks install luajwt CRYPTO_LIBDIR=/usr/lib/x86_64-linux-gnu  CRYPTO_INCDIR=/usr/include OPENSSL_LIBDIR=/usr/lib/x86_64-linux-gnu OPENSSL_INCDIR=/usr/include && \
+	luarocks install luajwtjitsi CRYPTO_LIBDIR=/usr/lib/x86_64-linux-gnu  CRYPTO_INCDIR=/usr/include OPENSSL_LIBDIR=/usr/lib/x86_64-linux-gnu OPENSSL_INCDIR=/usr/include && \
+	luarocks install jwt-jitsi CRYPTO_LIBDIR=/usr/lib/x86_64-linux-gnu  CRYPTO_INCDIR=/usr/include OPENSSL_LIBDIR=/usr/lib/x86_64-linux-gnu OPENSSL_INCDIR=/usr/include
```
sapkra: Do I really need to add `CRYPTO_LIBDIR`, `CRYPTO_INCDIR`, `OPENSSL_LIBDIR` and `OPENSSL_INCDIR `?
I saw it in the issue and just copied it.

prosody/Dockerfile
```diff
@@ -2,11 +2,19 @@ FROM jitsi/base
...
-	apt-dpkg-wrap apt-get install -t stretch-backports -y prosody && \
+	apt-dpkg-wrap apt-get install -t stretch-backports -y prosody luarocks gcc libssl1.1 libssl1.0-dev git && \
 	apt-cleanup && \
 	rm -rf /etc/prosody
 
 COPY rootfs/ /
+COPY git-modules/jitsi-meet/resources/prosody-plugins /prosody-plugins
+
+RUN \
+	luarocks list --outdated && \
```
sapkra: I think I can remove this line because there are always no rocks installed at this time. Right?
Sorry I'm new in lua.

prosody/Dockerfile
```diff
@@ -2,11 +2,19 @@ FROM jitsi/base
 
 RUN \
 	apt-dpkg-wrap apt-get update && \
-	apt-dpkg-wrap apt-get install -t stretch-backports -y prosody && \
+	apt-dpkg-wrap apt-get install -t stretch-backports -y prosody luarocks gcc libssl1.1 libssl1.0-dev git && \
```
saghul: do we want to install these new packages from backports too? why?

prosody/Dockerfile
saghul: I don't think it's needed in this setup.

prosody/Dockerfile
saghul: Please try without them and simplify the code if they are not needed.

aaronkvanmeerten: the OPENSSL and CRYPTO dirs are likely required when building the jwt code, as components of these libraries are used when verifying signatures on JWT tokens.

Rodrigora:
FYI I just tested this PR and worked fine 👍

prosody/Dockerfile
sapkra: Yeah I think we don't need to do that. I will fix it in my next update.

prosody/Dockerfile
ronanmcnulty: git-modules/jitsi-meet/resources/prosody-plugins is empty, the copy fails

damencho:
There is a patch about an event for bosh-session. Without it jwt will not work. And that change is available only in trunk.
We are using for meet.jit.si trunk version 747 for a long time.

sapkra:
Thanks for this information. I tried running prosody-trunk `1018` but I always got the error `token required`. Now I installed `747` and it's working fine.

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -37,9 +53,15 @@ Component "{{ .Env.XMPP_INTERNAL_MUC_DOMAIN }}" "muc"
...
     muc_room_cache_size = 1000
+    {{ if .Env.ENABLE_TOKEN_AUTH }}
+    modules_enabled = { "token_verification" }
+    {{ end }}
 
 Component "{{ .Env.XMPP_MUC_DOMAIN }}" "muc"
     storage = "none"
+    {{ if .Env.ENABLE_TOKEN_AUTH }}
+    modules_enabled = { "token_verification" }
+    {{ end }}
```
sapkra: One more question: Is it right to add this module on both `XMPP_INTERNAL_MUC_DOMAIN ` and `XMPP_MUC_DOMAIN `?

prosody/Dockerfile
saghul: @aaronkvanmeerten sure, but being in standard locations, the env variables may be not required?

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
saghul: It should only be needed in XMPP_MUC_DOMAIN

prosody/Dockerfile
```diff
@@ -1,12 +1,24 @@
...
+RUN \
+	luarocks install luaossl CRYPTO_LIBDIR=/usr/lib/x86_64-linux-gnu CRYPTO_INCDIR=/usr/include OPENSSL_LIBDIR=/usr/lib/x86_64-linux-gnu OPENSSL_INCDIR=/usr/include && \
+	luarocks install luajwt CRYPTO_LIBDIR=/usr/lib/x86_64-linux-gnu CRYPTO_INCDIR=/usr/include OPENSSL_LIBDIR=/usr/lib/x86_64-linux-gnu OPENSSL_INCDIR=/usr/include && \
+	luarocks install luajwtjitsi CRYPTO_LIBDIR=/usr/lib/x86_64-linux-gnu CRYPTO_INCDIR=/usr/include OPENSSL_LIBDIR=/usr/lib/x86_64-linux-gnu OPENSSL_INCDIR=/usr/include && \
+	luarocks install jwt-jitsi CRYPTO_LIBDIR=/usr/lib/x86_64-linux-gnu CRYPTO_INCDIR=/usr/include OPENSSL_LIBDIR=/usr/lib/x86_64-linux-gnu OPENSSL_INCDIR=/usr/include && \
+	luarocks install luasec && \
+	luarocks install luabitop
 
 COPY rootfs/ /
+COPY git-modules/jitsi-meet/resources/prosody-plugins /prosody-plugins
```
saghul: Using a git submodule just for this is quite overkill. Please don't. use ADD for the plugins we specifically want instead. (we don't need all the plugins there)

sapkra: Do you know which plugins we need? I'm sure that we need `token/util.lib.lua` and `mod_token_verification.lua`. Do we need more e.g. `mod_token_auth.lua`?

saghul: I believe we do, yes.

saghul:
> There is a patch about an event for bosh-session. Without it jwt will not work. And that change is available only in trunk.
> We are using for meet.jit.si trunk version 747 for a long time.

@damencho The code in our patch (https://github.com/jitsi/jitsi-meet/blob/master/resources/prosody-plugins/mod_bosh.lua.patch) is already part of the 0.10.2 release, which is what we are using, because we are pulling it from stretch-backports: https://packages.debian.org/stretch-backports/prosody

So, do we really need to use prosody-trunk here?

@sapkra Did you test if this war working before switching to prosody-trunk? I'm afraid the docs may need an update.

damencho:
@saghul if we are talking just for jwt maybe it is ok. There are some muc events that were only available in trunk.

saghul:
@damencho What MUC events? What do we use them for? I'd like to take the opportunity and document this properly.

damencho:
So for tokens I see just `module:hook("muc-room-pre-create"` and `module:hook("muc-occupant-pre-join"`. For other modules we use more of those.

saghul:
@damencho Thanks! Ok, it seems we need Prosody 0.11.X at least then. AFAIS Prosody does provide a Debian repo: https://prosody.im/download/package_repository so that should be the best way forward IMHO.

damencho:
I have in my todo list to test 11, cause somebody reported not working (he was testing jwt).

saghul:
Sounds like this PR is a good opportunity then! @sapkra can you please update so we can test this? Cheers!

sapkra:
I was the guy who reported that it's not working with prosody 0.11.0. :smile:
I updated the Dockerfile already locally. I will push these changes in the next hours.

saghul:
@sapkra Haha we came full-circle then :-) No rush, let's get this done The Right Way (TM).

prosody/Dockerfile
sapkra: To apply this change I need a full list of all necessary plugins we need.

sapkra:
I tested it again with 0.11.1 & I'm still getting the error I mentioned in https://github.com/jitsi/lib-jitsi-meet/issues/837

prosody/Dockerfile
saghul: mod_auth_token
mod_token_verification
token/

saghul:
@sapkra Thanks! Once this PR is complete we'll have an easy way to test / debug this.

prosody/Dockerfile
sapkra: Updated the PR - please have a look.

prosody/Dockerfile
sapkra: Updated the PR and removed these variables - I tested it without them and it's working fine.

sapkra:
I think the PR is complete now - feel free to test/debug it.

prosody/Dockerfile
```diff
@@ -1,10 +1,25 @@
 FROM jitsi/base
 
+ADD https://packages.prosody.im/debian/pool/main/p/prosody/prosody_0.11.1-1~stretch1_amd64.deb /tmp/prosody.deb
```
saghul: Please add the repo instead of pulling the file manually, this way we'll get automatic updates as they are pushed to the repo.

prosody/Dockerfile
```diff
@@ -1,10 +1,25 @@
...
+
+ADD https://raw.githubusercontent.com/jitsi/jitsi-meet/master/resources/prosody-plugins/token/util.lib.lua /prosody-plugins/token/util.lib.lua
+ADD https://raw.githubusercontent.com/jitsi/jitsi-meet/master/resources/prosody-plugins/mod_token_verification.lua /prosody-plugins/mod_token_verification.lua
+ADD https://raw.githubusercontent.com/jitsi/jitsi-meet/master/resources/prosody-plugins/mod_auth_token.lua /prosody-plugins/mod_auth_token.lua
+
 RUN \
 	apt-dpkg-wrap apt-get update && \
-	apt-dpkg-wrap apt-get install -t stretch-backports -y prosody && \
+	apt-dpkg-wrap apt-get install -y lua5.2 liblua5.2-dev lua-expat lua-socket lua-filesystem ssl-cert luarocks gcc libssl1.1 libssl1.0-dev git && \
```
saghul: Why is git necessary?

sapkra: For me it seems like git is used to download `luajwt`.

````
Installing https://luarocks.org/luajwtjitsi-1.3-7.rockspec
Cloning into 'luajwt'...
Note: checking out 'e7a6ca8cbe5941638db34e280eff7e75c07c4ec9'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b <new-branch-name>
````

Comment (saghul) [reply prosody/Dockerfile]: 👍

File: prosody/Dockerfile (hunk 6 above)
Comment (sapkra) [reply prosody/Dockerfile]: Ohh yeah you're right. I changed it.

File: prosody/Dockerfile [hunk 8]
```diff
@@ -1,10 +1,30 @@
...
+ADD https://prosody.im/files/prosody-debian-packages.key /tmp/key
+
+ADD https://raw.githubusercontent.com/jitsi/jitsi-meet/master/resources/prosody-plugins/token/util.lib.lua /prosody-plugins/token/util.lib.lua
+ADD https://raw.githubusercontent.com/jitsi/jitsi-meet/master/resources/prosody-plugins/mod_token_verification.lua /prosody-plugins/mod_token_verification.lua
+ADD https://raw.githubusercontent.com/jitsi/jitsi-meet/master/resources/prosody-plugins/mod_auth_token.lua /prosody-plugins/mod_auth_token.lua
+
 RUN \
 	apt-dpkg-wrap apt-get update && \
-	apt-dpkg-wrap apt-get install -t stretch-backports -y prosody && \
+	apt-dpkg-wrap apt-get install -y gnupg lua5.2 liblua5.2-dev lua-expat lua-socket lua-filesystem ssl-cert luarocks gcc libssl1.1 libssl1.0-dev git
```
saghul: Please remove gnupg  (and I think all -dev libraries too) to keep the container smaller.

prosody/Dockerfile
```diff
@@ -1,10 +1,30 @@
...
 	apt-cleanup && \
-	rm -rf /etc/prosody
+	rm -rf /etc/prosody /tmp/key
+
+RUN \
+	luarocks install luaossl && \
+	luarocks install luajwt && \
+	luarocks install luajwtjitsi && \
+	luarocks install jwt-jitsi && \
+	luarocks install luasec && \
```
saghul: Let's use the debian package: https://packages.debian.org/source/stretch/lua-sec

prosody/Dockerfile
```diff
@@ -1,10 +1,30 @@
...
+	luarocks install lua-cjson 2.1.0-1
```
saghul: Let's use the debian package: https://packages.debian.org/stretch/lua-cjson

prosody/Dockerfile
```diff
@@ -1,10 +1,30 @@
...
+	echo deb http://packages.prosody.im/debian stretch main | tee -a /etc/apt/sources.list && \
+	apt-key add /tmp/key && \
+	apt-dpkg-wrap apt-get update && \
+	apt-dpkg-wrap apt-get install -y prosody && \
 	apt-cleanup && \
-	rm -rf /etc/prosody
+	rm -rf /etc/prosody /tmp/key
+
+RUN \
+	luarocks install luaossl && \
```
saghul: Let's use the Debian package: https://packages.debian.org/stretch/lua-luaossl

prosody/Dockerfile
```diff
@@ -1,10 +1,30 @@
 FROM jitsi/base
 
+ADD https://prosody.im/files/prosody-debian-packages.key /tmp/key
+
+ADD https://raw.githubusercontent.com/jitsi/jitsi-meet/master/resources/prosody-plugins/token/util.lib.lua /prosody-plugins/token/util.lib.lua
```
saghul: Please pin this to a specific commit, so we can control what version of the file is pulled.

prosody/Dockerfile
sapkra: Gnupg is needed to add the key and there was another problem with the upgrade to lua 5.2 so that I had to add the dev package.
I can try it without them again.
Gnupg can be removed after adding the key.

saghul: Sorry, I meant remove them at the end, after the installation is done :-) They are probably needed at build time indeed.

prosody/Dockerfile
sapkra: But because we are using lua 5.2 now I have to lock the version to 2.1.0-1. Otherwise I get this error while building.

```
lua_cjson.c: In function 'json_append_data':
lua_cjson.c:743:19: warning: implicit declaration of function 'lua_objlen' [-Wimplicit-function-declaration]
             len = lua_objlen(l, -1);
```

And this one while running the container:
```
error loading module 'cjson' from file '/usr/local/lib/lua/5.2/cjson.so':
/usr/local/lib/lua/5.2/cjson.so: undefined symbol: lua_objlen
```

prosody/Dockerfile
sapkra: Done.

prosody/Dockerfile
sapkra: Alright. It's pushed.

prosody/Dockerfile
saghul: Do you see this error when using the Debian packages themselves?

sapkra: Yes, I think I copied these errors from exactly this build.

saghul: Looks like you still have some old module there. Debian won't install the package in /usr/local

sapkra: I just noticed that luajwt is also installing it as a dependency. Maybe the log was from this part of the build process.

saghul: Hum, that's odd, why on earth doesn't it detect it's installed? Maybe because it's not installed via luarocks?

saghul:
We have been quite busy this week, sorry. Didn’t have any time at all to look at this.

sapkra:
Are you still busy or is it possible to find some time?
It would be great to get this tested & merged soon - it will simplify a lot in my project.

docker-compose.yml
```diff
@@ -15,6 +15,11 @@ services:
             - ENABLE_LETSENCRYPT
             - ENABLE_HTTP_REDIRECT
             - DISABLE_HTTPS
+            - JWT_ENABLE_TOKEN_AUTH
+            - JWT_APP_ID
+            - JWT_APP_SECRET
+            - JWT_ACCEPTED_ISSUERS
+            - JWT_ACCEPTED_AUDIENCES
```
sapkra: I noticed that I set the variables for the wrong container. They has to be added to the prosody container.

saghul:
>     1. Prosody 0.11.x is saving the certs in the `data_path` so I have to edit the script. I would prefer to let them at this path but it would result in a more or less breaking change. What's your opinion?

I'd set that to /config/keys

As for breaking, this repo is contains "experimental" in all caps, so it's fair game :-)



>     1. I'm getting the following error. To fix this the `storage` should be changed to `internal` where we have set the `muc_room_cache_size` in `jitsi-meet.cfg.lua`. [jitsi/jitsi-meet#3454](https://github.com/jitsi/jitsi-meet/issues/3454)

The thing is we want to have no storage at all. Maybe it has a different name in 0.11 ?

Thanks for continuing your work on this!

saghul:
@sapkra Hey Paul, after all your latest changes, are you able to get it working on 0.11 ?

sapkra:
No, I just applied the changes I already had in my test environment. I also tested it again and I got the same error.

saghul:
@sapkra Thanks for confirming.

saghul:
@sapkra Thanks! Some things must have become simpler, since Prosody is now already at 0.11.

sapkra:
Ohh, I didn't noticed that prosody 0.11 is already published in stretch-backports. I will simplify the Dockerfile.

sapkra:
I refactored the Dockerfile and also removed `luarocks install jwt-jitsi` from it.
I think we don't need it because it is also not getting installed within the installation process of the `jitsi-meet-tokens` debian package (I couldn't find it).

prosody/Dockerfile
```diff
@@ -1,11 +1,24 @@
...
 	apt-dpkg-wrap apt-get update && \
+	apt-dpkg-wrap apt-get install -y lua5.2 liblua5.2-dev libssl1.0-dev luarocks gcc git && \
 	apt-dpkg-wrap apt-get install -t stretch-backports -y prosody && \
-	apt-cleanup && \
 	rm -rf /etc/prosody
 
+RUN \
+  luarocks install lua-cjson 2.1.0-1 && \
+	luarocks install luajwtjitsi && \
+	luarocks install basexx
```
saghul: This one we have on Debian: https://packages.debian.org/stretch/lua-basexx

prosody/Dockerfile
```diff
@@ -1,11 +1,24 @@
...
+
 RUN \
 	apt-dpkg-wrap apt-get update && \
+	apt-dpkg-wrap apt-get install -y lua5.2 liblua5.2-dev libssl1.0-dev luarocks gcc git && \
 	apt-dpkg-wrap apt-get install -t stretch-backports -y prosody && \
-	apt-cleanup && \
 	rm -rf /etc/prosody
 
+RUN \
+  luarocks install lua-cjson 2.1.0-1 && \
```
saghul: I hate this cjson thing so much I took a closer look. It's totally unmaintained and there is no fork which supports Lua 5.2 properly (unless you se an old version which didn't implement certain feature). So it doesn't seem very future proof.

Since the luajwtjitsi module is ours, I think we can just swap the cjson module for another better JSON library.

In addition, luacrypto is super-deprecated and it's recommended to use luaossl now.

I'm going to try to update luajwtjitsi so it uses those 2 so this can be greatly simplified. You don't have to do anything, but thanks for listening to my rant :-)

prosody/Dockerfile
sapkra: I wanted to use the same methods like `jitsi-meet-tokens`.
https://github.com/jitsi/jitsi-meet/blob/master/debian/jitsi-meet-tokens.postinst#L68

Why do we want to do it differently?

saghul: If it works as expected I'll also make the change in the jitsi-meet-tokens package.

sapkra: Done.

saghul:
Ok, this looks pretty much ready to test and dig why it's not working with the update to 0.11. I'll get on it soon.

FYI, I re-targetted your PR to the `dev` branch. I want to try a new way of doing things here: we merge the stuff to the dev branch and merge dev into master when a new release to DockerHub is made.

saghul:
@sapkra Now that we know that the problem was, can you please embed the required JWT files in this PR (in defaults/jwt for example) and we'll remove them later, once we have this in jitsi-meet.

sapkra:
I just added an `sed` command to the Dockerfile which will replace `hook` through `hook_global`. I think this is the cleanest temporary solution.

saghul:
> I just added an `sed` command to the Dockerfile which will replace `hook` through `hook_global`. I think this is the cleanest temporary solution.

Yeah, sounds good! I hope to merge this really soon. Thanks for your patience!

sapkra:
@saghul There are still two open discussions related to `lua-cjson`. Do you want to refactor this stuff after we merged the PR?

saghul:
> There are still two open discussions related to `lua-cjson`. Do you want to refactor this stuff after we merged the PR?

Yes. I have no time to address those now, but I think this PR is very valuable as is, so we can address those later.

saghul:
@sapkra I sarted testing this and ran into some issues: the prosody plugin paths are owned by root, so prosody cznnot read them. I solved by adding the following:

```
diff --git a/prosody/rootfs/etc/cont-init.d/10-config b/prosody/rootfs/etc/cont-init.d/10-config
index 45efdbb..01b4925 100644
--- a/prosody/rootfs/etc/cont-init.d/10-config
+++ b/prosody/rootfs/etc/cont-init.d/10-config
@@ -10,6 +10,14 @@ if [[ "$(stat -c %U /config)" != "prosody" ]]; then
     chown -R prosody /config
 fi
 
+if [[ "$(stat -c %U /prosody-plugins)" != "prosody" ]]; then
+    chown -R prosody /prosody-plugins
+fi
+
+if [[ "$(stat -c %U /prosody-plugins-custom)" != "prosody" ]]; then
+    chown -R prosody /prosody-plugins-custom
+fi
+
 if [[ ! -f $PROSODY_CFG ]]; then
     cp -r /defaults/* /config
     tpl /defaults/conf.d/jitsi-meet.cfg.lua > /config/conf.d/jitsi-meet.cfg.lua

```

Can you please add this changes to the PR? Thank you!

sapkra:
You are also able to push it. `Allow edits from maintainers.` is checked. 😉

saghul:
> You are also able to push it. `Allow edits from maintainers.` is checked. 😉

Oh, I forgot to check.



> OpenShift is starting the container always with a user which is in the root group - so I didn't noticed this problem.

No worries. Also, good to know!

saghul:
Thanks for all your work here @sapkra, it works beautifully! I'll hold on making a new release until I make some changes I hope to make soon: introduce some AUTH_TYPE wich can be "internal" (the default) "jwt" and soon "ldap".

alexanderankin:
if its not too much to ask, what is the use case of this? i know how you can have ldap server, and user accounts on it, but jwt is like crypto signiature?

saghul:
This is the typical auth scenario when you are embedding Jitsi into an existing app which has its own authentication. You’d create a token in your backend and configure Jitsi to use JWT for authentication. The token can limit the room that a user can join, establish their saben and avatar on the Jitsi side of things in addition.

alexanderankin:
so i have to link to the page with "myinstance.mydomain.com/theroom?token=eyJhbGci..."? where is token expected? in "Authorization: Bearer eyJhbGci..."? I am used to seeing this in API call not in page. Or am I supposed to integrate much closer, like consume video feed stream and show to user in my own app with html5 technologies?

saghul:
You can use the URL parameter, yes, or our iframe API, in you are embedding Jitsi into an existing site / application.

valsaraj:
Following error when try to install JWT token with latest jitsi-meet:

> apt-get install jitsi-meet-tokens
> Reading package lists... Done
> Building dependency tree       
> Reading state information... Done
> Some packages could not be installed. This may mean that you have
> requested an impossible situation or if you are using the unstable
> distribution that some required packages have not yet been created
> or been moved out of Incoming.
> The following information may help to resolve the situation:
> 
> The following packages have unmet dependencies:
>  jitsi-meet-tokens : Depends: prosody-trunk (>= 1nightly747) but it is not installable or
>                               prosody-0.11 but it is not installable or
>                               prosody (>= 0.11.2) but 0.10.0-1build1 is to be installed

saghul:
You don’t need to install that if you are using the Docker setup.

valsaraj:
Would you share the link of Docker referred?
Why it's different from the one comes with quick install?

saghul:
You are commenting in an issue on the Docker setup repo. Hence we will assume you are using the Docker setup. Please open a discussion topic in our community otherwise.

PR #31, opened 2018-11-20
Add TCP Harvester configuration options

Fishbowler:
Add TCP fallback comms to the JVB, leaving it off by default.
Was useful in my specific context, where OSX Firewall blocked UDP.

### DIFF
--- a/jvb/rootfs/defaults/sip-communicator.properties
+++ b/jvb/rootfs/defaults/sip-communicator.properties
@@ -1,6 +1,8 @@
 org.jitsi.videobridge.AUTHORIZED_SOURCE_REGEXP={{ .Env.JICOFO_AUTH_USER }}@{{ .Env.XMPP_AUTH_DOMAIN }}/.*
 org.jitsi.videobridge.SINGLE_PORT_HARVESTER_PORT={{ .Env.JVB_PORT }}
-org.jitsi.videobridge.DISABLE_TCP_HARVESTER=true
+org.jitsi.videobridge.DISABLE_TCP_HARVESTER={{ .Env.JVB_TCP_HARVESTER_DISABLED }}
+org.jitsi.videobridge.TCP_HARVESTER_PORT={{ .Env.JVB_TCP_PORT }}
+org.jitsi.videobridge.TCP_HARVESTER_MAPPED_PORT={{ .Env.JVB_TCP_PORT }}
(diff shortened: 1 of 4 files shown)


saghul:
Hey Dan, thanks for the PR! I think I'd like to explore other possibilities before, like sending media on port 443 as the fallback, but that requires some more work on the nginx config side. What scenario do you have more precisely?

Fishbowler:
With strict mode enabled on the macOS firewall, UDP doesn't make it from a browser on the host through to the JVB.

Also happy to look at other methods, but would say that offering more configurations options (with good defaults - this PR has the TCP Harvester switched off) rather than different configuration options feels like a worthy goal.

Realised I've missed updating the readme for these config options 😬

jvb/rootfs/defaults/sip-communicator.properties
```diff
@@ -1,6 +1,8 @@
 org.jitsi.videobridge.AUTHORIZED_SOURCE_REGEXP={{ .Env.JICOFO_AUTH_USER }}@{{ .Env.XMPP_AUTH_DOMAIN }}/.*
 org.jitsi.videobridge.SINGLE_PORT_HARVESTER_PORT={{ .Env.JVB_PORT }}
-org.jitsi.videobridge.DISABLE_TCP_HARVESTER=true
+org.jitsi.videobridge.DISABLE_TCP_HARVESTER={{ .Env.JVB_TCP_HARVESTER_DISABLED }}
+org.jitsi.videobridge.TCP_HARVESTER_PORT={{ .Env.JVB_TCP_PORT }}
+org.jitsi.videobridge.TCP_HARVESTER_MAPPED_PORT={{ .Env.JVB_TCP_PORT }}
```
saghul: I don't think this last one is needed. It only applies if different from the port.

saghul:
@Fishbowler I see. Ok, let's go for it then. Can you please update the readme? I also left a minor comment.

jvb/rootfs/defaults/sip-communicator.properties
Fishbowler: Have removed it for now.

Fishbowler:
Readme updated. Extra config removed.

saghul:
Nice work @Fishbowler ! ✨

PR #36, opened 2018-12-04
add JIGASI_SIP_PORT and JIGASI_SIP_TRANSPORT variables

netaskd:
Add JIGASI_SIP_PORT and JIGASI_SIP_TRANSPORT variables for a native configure sip transport and sip port.


saghul:
Cheers!

PR #47, opened 2019-01-15
Custom prosody modules

sapkra:
I need a possibility to add custom prosody modules. I think this is a flexible solution to do that.

FYI: It will have some small conflicts with #29. I have to resolve them after one of the PRs is merged.

### DIFF
--- a/prosody/Dockerfile
+++ b/prosody/Dockerfile
@@ -10,5 +10,4 @@ COPY rootfs/ /
 
 EXPOSE 5222 5269 5347 5280
 
-VOLUME /config
-
+VOLUME ["/config", "/prosody-plugins-custom"]
--- a/README.md
+++ b/README.md
@@ -152,6 +152,7 @@ Variable | Description | Default value
 `XMPP_AUTH_DOMAIN` | Internal XMPP domain for authenticated services | auth.meet.jitsi
 `XMPP_MUC_DOMAIN` | XMPP domain for the MUC | muc.meet.jitsi
 `XMPP_INTERNAL_MUC_DOMAIN` | XMPP domain for the internal MUC | internal-muc.meet.jitsi
+`XMPP_GUEST_DOMAIN` | XMPP domain for unauthenticated users | guest.meet.jitsi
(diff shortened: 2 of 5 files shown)


prosody/Dockerfile
```diff
@@ -10,5 +12,4 @@ COPY rootfs/ /
 
 EXPOSE 5222 5269 5347 5280
 
-VOLUME /config
-
+VOLUME ["/config", "$CUSTOM_MODULES"]
```
saghul: this should be a fixed volume: `/prosody-plugins`

prosody/Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM jitsi/base
 
+ENV CUSTOM_MODULES=/prosody-plugins-custom
```
saghul: this is not needed

prosody/Dockerfile
saghul: You can add a variable in the yaml file to device which directory is mounted there

sapkra: But when we add a fixed volume to `/prosody-plugins` the volume will overwrite the prosody plugins we have added in #29 
So I can change it to be a hardcoded path but it shouldn't be `/prosody-plugins`.
I separated the default and custom modules in different branches so at the end we will have one `/prosody-plugins` and `/prosody-plugins-custom` directory.

saghul: Sounds good, let's make it `/prosody-plugins-custom` then.

sapkra: Done

saghul:
Nice work! I'll give it a test and merge.

saghul:
Landed in fb1fdb1d53d805804245be93cc009337b4681727 and 95a55915b7f43312934f391aeda87782366b1689 cheers!

May I please ask you to follow the commit message "convention" we have? Thanks a lot for all your contributions thus far!

sapkra:
Is the convention documented somewhere or should I just learn it by reading the commit messages and understand how they are structured?

saghul:
No it's not documented, but if you look at git log you'll see there is not much to it :-)

PR #49, opened 2019-01-18
Make jigasi completely optional

### LINKED ISSUE #48: Make jigasi completely optional
I noticed that when I want to setup a cluster without jigasi some jigasi specific code will be executed.

For example:
```bash
prosodyctl --config $PROSODY_CFG register $JIGASI_XMPP_USER $XMPP_AUTH_DOMAIN $JIGASI_XMPP_PASSWORD
```

All `JIGASI_*` variables should be optional.
To achieve this we should implement an environment variable like `JIGASI_ENABLE`.

**Edit:**
Or maybe the following is also good enough.

```
if [[ ! -z $JIGASI_XMPP_USER ]] && [[ ! -z $JIGASI_XMPP_PASSWORD ]]; then
    prosodyctl --config $PROSODY_CFG register $JIGASI_XMPP_USER $XMPP_AUTH_DOMAIN $JIGASI_XMPP_PASSWORD
fi
```

sapkra:
Closes #48

I haven't tested it yet - but I think it should work.


saghul:
Landed in 0f0a9420a05c363746e881d835f13fec373fb002 with some slight changes. Cheers!

PR #52, opened 2019-01-24
Fix for the issue with default directory for generated certs

alexcustos:
I'm not sure about the source of this issue, but prosodyctl kept insisting on creating certs in the data directory.

### DIFF
--- a/prosody/rootfs/etc/cont-init.d/10-config
+++ b/prosody/rootfs/etc/cont-init.d/10-config
@@ -4,6 +4,7 @@ PROSODY_CFG="/config/prosody.cfg.lua"
 
 if [[ ! -d /config/data ]]; then
     mkdir -p /config/data
+    chown prosody:prosody /config/data
     chmod 777 /config/data
 fi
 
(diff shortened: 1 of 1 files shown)


saghul:
Does this happen with the default setup or are you making any other changes yourself? This should not be a problem.

prosody/rootfs/etc/cont-init.d/10-config
```diff
@@ -4,6 +4,7 @@ PROSODY_CFG="/config/prosody.cfg.lua"
 
 if [[ ! -d /config/data ]]; then
     mkdir -p /config/data
+    chown prosody:prosody /config/data
```
saghul: This is unnecessary since we are chaning the permissions to 777 below.

saghul:
Hum, now that I see this makes sense since we do specify `data_dir`... maybe an older version of prosody had a bug. I'll give it a test.

prosody/rootfs/etc/cont-init.d/10-config
saghul: Thank you both!

saghul:
Landed in 0760253ebd78b734a2263b3cbd8e8628a5b3485d, thanks!

PR #54, opened 2019-01-29
jibri: initial commit

netaskd:
add jibri container


saghul:
Impressive work! It needs a rebase, but I'll start looking into it as soon as I can. Thank you!

netaskd:
> Impressive work! It needs a rebase, but I'll start looking into it as soon as I can. Thank you!

I' ve rebased. Please check.

docker-compose.yml
```diff
@@ -7,6 +7,9 @@ services:
         ports:
             - '${HTTP_PORT}:80'
             - '${HTTPS_PORT}:443'
+        expose:
```
saghul: Why is this needed?

docker-compose.yml
```diff
@@ -23,13 +27,18 @@ services:
             - XMPP_BOSH_URL_BASE=http://xmpp.meet.jitsi:5280
             - XMPP_GUEST_DOMAIN
             - XMPP_MUC_DOMAIN
+            - XMPP_RECORDER_DOMAIN
             - TZ
         networks:
             meet.jitsi:
+                aliases:
```
saghul: Why is this needed?

docker-compose.yml
```diff
@@ -23,13 +27,18 @@ services:
...
+                    - meet.jitsi
 
     # XMPP server
     prosody:
         image: jitsi/prosody
+        ports:
```
saghul: why is this needed?

jibri/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM jitsi/base-java
+
+ARG JITSI_RELEASE=unstable
```
saghul: Let's drop this and just use unstable for this directly, since it's "temporary" until we move jibri to stable.

jibri/rootfs/etc/cont-init.d/10-config
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/with-contenv bash
+export JIBRI_INSTANCE_ID=$(date +%N)
+
+# set external IP as xmpp domain
+echo "${DOCKER_HOST_ADDRESS}	${XMPP_DOMAIN}" >> /etc/hosts
```
saghul: why is this needed?

jibri/rootfs/etc/cont-init.d/10-config
```diff
@@ -0,0 +1,19 @@
...
+echo "${DOCKER_HOST_ADDRESS}	${XMPP_SERVER}" >> /etc/hosts
+
+# check loaded snd_aloop module
+[[ $(lsmod |grep -om1 snd_aloop) ]] \
+|| ( echo -e "\e[31mERROR: Please load snd-aloop module on the docker host.\e[39m"; kill 1; exit 1 )
+
+# always recreate configs
+frep --overwrite /defaults/config.json:/etc/jitsi/jibri/config.json
```
saghul: can you use the tpl helper please?

bbaldino:
very exciting, so everything with jibri is working in this container?

docker-compose.yml
netaskd: I tested it in two cases:
1. the container on the same host with prosody and web (I guess its not good idea)
2. the container on a different host
So, in the first case I can use it through internal docker bridge (like meet.jitsi) and I exposed the ports in the web container. For the second case it's not needed.

docker-compose.yml
netaskd: for the the same reason as before. it maps domain name of web container "meet.jitsi" to the internal "dns"

docker-compose.yml
netaskd: it needs for connect to prosody from external jibri instance (case 2)

jibri/Dockerfile
netaskd: I guess need a comment in README about it, because it's it is not clear.

jibri/rootfs/etc/cont-init.d/10-config
netaskd: `JIBRI_INSTANCE_ID` set "uniq" id for the control_muc -> nickname.
/etc/hosts - I use it for the second case, but maybe will better use external `LETSENCRYPT_DOMAIN` in jibri/rootfs/defaults/config.json->xmpp_environments->xmpp_domain instead. But again, its not clear

jibri/rootfs/etc/cont-init.d/10-config
netaskd: sure )

netaskd:
@bbaldino, for now I tested only recording and only one instance on the same host - it works well. I see it needs more settings for get multi instance on the same host.
The main and long troubleshooting was with crhome-out sound, because s6 by default do not set env `HOME` for jibri when start it. But settings for alsa (.asoundsrc) there are. So I added set env HOME before start jibri and now it works perfect.

docker-compose.yml
saghul: So, as a starting point, we are only interested in case 1. In the future, this should be solved by something like Docker Swarm, that is, by having multi-host transparent uder defined networking (the meet.jitsi network).

In this case, I don't thing this is necessary.

docker-compose.yml
saghul: But it isn't, if we focus on case 1, right? Let's start there.

docker-compose.yml
saghul: Then drop it, let's start with case 1.

jibri/Dockerfile
saghul: I don't think it's worth mentioning. We should mention that recording is "beta" though.

jibri/rootfs/defaults/config.json
```diff
@@ -0,0 +1,58 @@
+{
+    "recording_directory":"/tmp/recordings",
```
saghul: Do we need to specify these?

jibri/rootfs/defaults/logging.properties
```diff
@@ -0,0 +1,33 @@
+handlers = java.util.logging.FileHandler, java.util.logging.ConsoleHandler
+
+java.util.logging.FileHandler.level = FINE
+java.util.logging.FileHandler.pattern   = /var/log/jitsi/jibri/log.%g.txt
```
saghul: Put the logs in /config/logs

jibri/rootfs/etc/cont-init.d/10-config
saghul: I don't understand what you mean. But since you mentioned the second case, let's drop it please.

jibri/rootfs/etc/cont-init.d/10-config
```diff
@@ -0,0 +1,19 @@
...
+frep --overwrite /defaults/logging.properties:/etc/jitsi/jibri/logging.properties
+
+# make recording dir
+mkdir -p /tmp/recordings
```
saghul: Put this in /config so they are accessible from the outside.

docker-compose.yml
netaskd: It's needed for case 1

jibri/rootfs/etc/cont-init.d/10-config
sapkra: I think it would be a better idea to put this in another directory like /recordings for multiple reasons.
1. It's not a configuration
2. You can separate the volumes to use ConfigMaps and persitent storage in kubernetes environments

netaskd: @sapkra its already changed. Please see 48286a3

netaskd: use `$JIBRI_RECORDING_DIR` and `$JIBRI_LOGS_DIR` for adjust paths.

docker-compose.yml
saghul: This alias get's confusing, can you please change it to "web.meet.jitsi"?

jibri/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM jitsi/base-java
+
+RUN \
+	apt-dpkg-wrap apt-get update \
+	&& apt-dpkg-wrap apt-get install -y curl gnupg \
+	&& curl -4s https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
+	&& echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" > /etc/apt/sources.list.d/google-chrome.list \
+	&& apt-dpkg-wrap apt-get update \
+	&& apt-dpkg-wrap apt-get install -y jibri google-chrome-stable vim procps \
```
saghul: You'll need to change the jitsi repo to unstable or this won't work. For the moment, all containers will use the stable repo except this one, because it's not yet available on stable.

netaskd: done

docker-compose.yml
netaskd: Yes, but changing the name will affects:
```
"xmpp_environments": [
{
...
 "xmpp_domain": "{{ .Env.XMPP_DOMAIN }}" 
...
}
``` 
in jibri/rootfs/defaults/config.json
So, I guess, by logic it must be xmpp_domain and we already have it as "meet.jitsi" isn't it?

saghul: Maybe @bbaldino can enlighten us here. What should that point to?

saghul: Given that we set `xmpp_server_hosts` to the actual XMPP server, I wonder what's going on. @netaskd what happens if you remove the alias?

netaskd: jibri container won't find https://meet.jitsi/

saghul: If that is the case, then that config option is really ill named. Anyway, plase alias the container to web.meet.jitsi and let's pass an extra env variable: XMPP_WEB_SERVER.

netaskd: When I change it I get the error
```
jibri_1_1838a0eed8da | 2019-01-31 18:09:10.093 SEVERE: [40] org.jitsi.jibri.api.xmpp.XmppApi.run() Error starting Jibri service : java.lang.StringIndexOutOfBoundsException: String index out of range: -1 with stack: 
jibri_1_1838a0eed8da | java.lang.String.substring(String.java:1967)
jibri_1_1838a0eed8da | java.lang.String.subSequence(String.java:2003)
jibri_1_1838a0eed8da | org.jitsi.jibri.util.XmppUtilsKt.getCallUrlInfoFromJid(XmppUtils.kt:38)
jibri_1_1838a0eed8da | org.jitsi.jibri.api.xmpp.XmppApi.handleStartService(XmppApi.kt:276)
jibri_1_1838a0eed8da | org.jitsi.jibri.api.xmpp.XmppApi.access$handleStartService(XmppApi.kt:71)
jibri_1_1838a0eed8da | org.jitsi.jibri.api.xmpp.XmppApi$handleStartJibriIq$1.run(XmppApi.kt:200)
jibri_1_1838a0eed8da | java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
jibri_1_1838a0eed8da | java.util.concurrent.FutureTask.run(FutureTask.java:266)
jibri_1_1838a0eed8da | java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
jibri_1_1838a0eed8da | java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
jibri_1_1838a0eed8da | java.lang.Thread.run(Thread.java:748)
```
I guess it because some function try to handle config option: 
```
"room_jid_domain_string_to_strip_from_start"
```

saghul: Hum. Let's see if @bbaldino can enlighten us. I see xmppDomain is used here: https://github.com/jitsi/jibri/blob/0e96134dcc5370d4668533e160d573f13436f591/src/main/kotlin/org/jitsi/jibri/util/XmppUtils.kt#L42 But that seems a bit off to me, since the web URL may be completely diffrent to the XMPP domain, as we are seeing here.

bbaldino: web url is currently defined by the xmpp domain, because we derive it from looking at the domain of the jid in the jibri request.  i don't think jicofo actually knows the web url of the conference.  @aaronkvanmeerten probably has some good info here.

netaskd: the error from here:
https://github.com/jitsi/jibri/blob/0e96134dcc5370d4668533e160d573f13436f591/src/main/kotlin/org/jitsi/jibri/util/XmppUtils.kt#L38
Because it don't expected domain like web.meet.jitsi only like meet.jitsi
muc.meet.jitsi (from jid) can't be parsed to subdomain from web.meet.jitsi, only like muc.web.meet.jitsi can be.

saghul: That's unfortunate, because they are not really related :-/

@netaskd please leave the alias there then. Thanks for trying things out!

netaskd:
I found one else dependence for start containers together. jibri must be run on the last stage. depends_on is not good here, because it bind jibri to another container. I think we need something like script "wait-for-it"

MattGook:
Jibri is really helpful for a small group to review their meeting record. We are looking forward to the Jibri join the Docker Compose set. Thanks so much for the efforts from @netaskd @saghul.

nileshsargar-idx9:
This is very nice work, we were waiting for Jibri container from long time. How soon are expecting this to be merged?

netaskd:
Multi-instance can be used with JiBRI commit https://github.com/jitsi/jibri/pull/204/commits/9ebe262. PR https://github.com/jitsi/jibri/pull/204
For select alsa sound card, just set new loopback interface inside `/home/jibri/.asoundrc` (https://github.com/jitsi/jibri/blob/78258b63aeb24be9abf8bda518dcf328f1e98854/resources/debian-package/etc/jitsi/jibri/asoundrc#L5) for each plug for every docker instance.
For example:
Default the first instance has:
```
...
slave.pcm "hw:Loopback,0,0"
...
slave.pcm "hw:Loopback,0,1"
...
slave.pcm "hw:Loopback,1,1"
...
slave.pcm "hw:Loopback,1,0"
...
```
for the second instance it can be changed to:
```
...
slave.pcm "hw:Loopback_1,0,0"
...
slave.pcm "hw:Loopback_1,0,1"
...
slave.pcm "hw:Loopback_1,1,1"
...
slave.pcm "hw:Loopback_1,1,0"
...
```
also you can use numbering id. The third instance will use the third loopback interface:
```
...
slave.pcm "hw:2,0,0"
...
slave.pcm "hw:2,0,1"
...
slave.pcm "hw:2,1,1"
...
slave.pcm "hw:2,1,0"
...
```
and so on.

jinfeei:
Hi, am unsure how to use this jibri container. What are the settings I need to make to use this? Some very brief notes would be very helpful... Can please help? Thanks.

saghul:
@jinfeei Have you checked the README? https://github.com/jitsi/docker-jitsi-meet/pull/54/files#diff-04c6e90faac2675aa89e2176d2eec7d8R122

jinfeei:
@saghul Yes, I'm very ashamed to say I'm confused.  For example, in the section "JItsi BRoadcasting Infrastructure configuration", I'm not sure where should I make these changes - isn't it everything already inside a docker container?  Also, the jibri.yml file, how do I run this? Do I clone the whole directory over and re-make everything on my server?

Without Jibri, I can already run everything using docker compose and it is pretty straightforward. Just that I'm confuse how/where to plug in the Jibri container... Very sorry for my ignorance and THANK YOU!

lnovara:
@netaskd , @saghul, @sapkra what's the state of this PR?

We are in the process of deploying jibri on Kubernetes, can we help in getting this merged?

netaskd:
@saghul could you re-open this PR? It was auto-closed after my the latest push and I can't re-open it back

saghul:
Weird, the button is gayer out for me too. Just open a new one then.

netaskd:
PR moved to #129

PR #56, opened 2019-02-01
prosody: fix owner for /config dir. it affects many cases.

netaskd:
Due the /config directory prosody uses under user "prosody" we need to set this directory owned this user.

### DIFF
--- a/prosody/rootfs/etc/cont-init.d/10-config
+++ b/prosody/rootfs/etc/cont-init.d/10-config
@@ -3,8 +3,11 @@
 PROSODY_CFG="/config/prosody.cfg.lua"
 
 if [[ ! -d /config/data ]]; then
-    mkdir -p /config/data
-    chmod 777 /config/data
+    mkdir -pm 750 /config/data
+fi
+
+if [[ "$(stat -c %U /config)" != "prosody" ]]; then
+    chown -R prosody /config
 fi
 
 if [[ ! -f $PROSODY_CFG ]]; then


saghul:
What problems did you run into? IIRC this was fixed in Prosody upstream and the version we now use (0.11) should be fixed.

netaskd:
@saghul it's not dependend from prosody server (i guess). the first thing it affect creating cer files for prosody:
try to do inside the prosody container:
1. `rm -rf /config/*`
2. `bash -x /var/run/s6/etc/cont-init.d/10-config`
you'll see:
```
...
+ prosodyctl --config /config/prosody.cfg.lua cert generate meet.jitsi
The directory /config/data is not owned by the current user, won't be able to write files to it
+ [[ ! -f /config/certs/auth.meet.jitsi.crt ]]
+ echo
+ prosodyctl --config /config/prosody.cfg.lua cert generate auth.meet.jitsi
The directory /config/data is not owned by the current user, won't be able to write files to it
...
```
then:
3. `ls -al /config/certs/`
shows `total 0`
So, certs is not created... 

I was catching another glitches but can't say it now

saghul:
That seems like https://issues.prosody.im/1075 which @sapkra also ran into, and AFAIK it's fixed. I mean, we should get errors when building the current containers otherwise, shouldn't we?

netaskd:
@saghul  
The fix 1075 is just do error more clarify in case when "cert" dir has bit 777. But I say about set owner as prosody wants. I get the warning (in fact it is an error because they do "return 1") not when building, but when starting it.

saghul:
@netaskd Sorry, I had missed part of your message.

> try to do inside the prosody container:
> 
>     1. `rm -rf /config/*`
> 
>     2. `bash -x /var/run/s6/etc/cont-init.d/10-config`
>        you'll see:

But why would anyone want to do that? I just tried a fresh install and get no problems. What is it exactly you are trying to do here?

saghul:
Ah, I see now. This is triggered when restarting the container, doesn't matter why, but it's important. Thanks @netaskd !

karser:
Hey guys, I spent half a day fighting with permission issue. I was using master branch and latest images. @saghul  It is to do not only with prosody, so likely your fix has to be applied to other containers as well.

Here is what I came up with:

First launch
```
# Assuming /mnt/data/jitsi-meet/ is empty
# Fix permissions before the next line
docker-compose up -d
docker-compose down
# Fix permissions once again
# And launch everything to trigger init scripts
docker-compose up -d
# See the logs to make sure there are no errors
docker-compose logs -f
```

Fix permissions:
```
mkdir -p /mnt/data/jitsi-meet/; \
chmod 777 /mnt/data/jitsi-meet/; \
mkdir -p /mnt/data/jitsi-meet/prosody/data/; \
chown -R 101:101 /mnt/data/jitsi-meet/prosody/; \
mkdir -p /mnt/data/jitsi-meet/jicofo/; \
chown -R 999:999 /mnt/data/jitsi-meet/jicofo/; \
mkdir -p /mnt/data/jitsi-meet/jvb/; \
chown -R 999:999 /mnt/data/jitsi-meet/jvb/
```

saghul:
The master branch doesn't include this fix, it's on the dev branch. If you think more fixes are necessary (I haven't run into issues myself) please open a PR.

karser:
Ok, let's see if the errors are gone when this PR is released.
I guess, these errors are not fatal, so you could miss them if you don't look at logs of jicofo/jvb containers.

saghul:
@karser You can check this branch to test it: https://github.com/jitsi/docker-jitsi-meet/tree/dev

PR #58, opened 2019-02-07
web: add option for enable chrome screen sharing

netaskd:
The PR enable screensharing for Chrome (has been working in Chrome since version 72)


saghul:
This shouldn't be needed. It's enabled by default on M72 or greater. If it isn't, then we need to change that in Jitsi Meet itself.

netaskd:
In config.js the option sharing in chrome disabled. Why it should work withouit enabling in Chrome 72?
https://github.com/jitsi/docker-jitsi-meet/blob/2da615c1f43a83e91221b04452855909e16aa59a/web/rootfs/defaults/config.js#L149

saghul:
That's a dumb default. I'll get it changed in Jitsi Meet and we'll sync the configs. It should also be commented out. Sigh.

saghul:
@netaskd https://github.com/jitsi/jitsi-meet/pull/3883

netaskd:
@saghul looks good, I'm waiting sync config.js

PR #67, opened 2019-02-16
jibri patched into compose

AliRezaTaleghani:
trying to adding broadcast  (Jibri) service based on Docker build recipes.


saghul:
Thanks for the effort, but we are already working on it at #54 Please take a look and feel free to point out potential problems.

PR #76, opened 2019-03-14
misc: now all variables moved to .env file

netaskd:
variables moved to .env file for fully customization of jitsi-meet service domains


saghul:
Nice, thanks!

PR #80, opened 2019-03-15
jigasi: add SDES srtp enabling possibility and healthcheck

netaskd:
this PR adds possibility to set in .env file: 
- JIGASI_ENABLE_SDES_SRTP for enabling SDES srtp in jigasi
- JIGASI_HEALTH_CHECK_SIP_URI set healthcheck extansion for call to
- JIGASI_HEALTH_CHECK_INTERVAL set time interval (in ms) for calling healthcheck extension
- JIGASI_SIP_KEEP_ALIVE_METHOD set change default sip ping method, by default is OPTION

### DIFF
--- a/jigasi/rootfs/defaults/sip-communicator.properties
+++ b/jigasi/rootfs/defaults/sip-communicator.properties
@@ -11,8 +11,20 @@ net.java.sip.communicator.impl.protocol.sip.acc1=acc1
 {{ if and .Env.JIGASI_SIP_PORT .Env.JIGASI_SIP_TRANSPORT }}
 net.java.sip.communicator.impl.protocol.sip.acc1.PROXY_ADDRESS={{ .Env.JIGASI_SIP_SERVER }}
 net.java.sip.communicator.impl.protocol.sip.acc1.PROXY_AUTO_CONFIG=false
-net.java.sip.communicator.impl.protocol.sip.acc1.PROXY_PORT={{ .Env.JIGASI_SIP_PORT }}
-net.java.sip.communicator.impl.protocol.sip.acc1.PREFERRED_TRANSPORT={{ .Env.JIGASI_SIP_TRANSPORT }}
+net.java.sip.communicator.impl.protocol.sip.acc1.PROXY_PORT={{ .Env.JIGASI_SIP_PORT | default "5060" }}
+net.java.sip.communicator.impl.protocol.sip.acc1.PREFERRED_TRANSPORT={{ .Env.JIGASI_SIP_TRANSPORT | default "UDP" }}
+{{ end }}
+{{ if .Env.JIGASI_ENABLE_SDES_SRTP | default "0" | toBool }}
+net.java.sip.communicator.impl.protocol.sip.acc1.SAVP_OPTION=1
+net.java.sip.communicator.impl.protocol.sip.acc1.DEFAULT_ENCRYPTION=true
+net.java.sip.communicator.impl.protocol.sip.acc1.DEFAULT_SIPZRTP_ATTRIBUTE=false
+net.java.sip.communicator.impl.protocol.sip.acc1.ENCRYPTION_PROTOCOL.ZRTP=0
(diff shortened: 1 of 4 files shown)


jigasi/rootfs/defaults/sip-communicator.properties
```diff
@@ -11,16 +11,28 @@ net.java.sip.communicator.impl.protocol.sip.acc1=acc1
 {{ if and .Env.JIGASI_SIP_PORT .Env.JIGASI_SIP_TRANSPORT }}
 net.java.sip.communicator.impl.protocol.sip.acc1.PROXY_ADDRESS={{ .Env.JIGASI_SIP_SERVER }}
 net.java.sip.communicator.impl.protocol.sip.acc1.PROXY_AUTO_CONFIG=false
-net.java.sip.communicator.impl.protocol.sip.acc1.PROXY_PORT={{ .Env.JIGASI_SIP_PORT }}
-net.java.sip.communicator.impl.protocol.sip.acc1.PREFERRED_TRANSPORT={{ .Env.JIGASI_SIP_TRANSPORT }}
+net.java.sip.communicator.impl.protocol.sip.acc1.PROXY_PORT={{ .Env.JIGASI_SIP_PORT | default "5060" }}
+net.java.sip.communicator.impl.protocol.sip.acc1.PREFERRED_TRANSPORT={{ .Env.JIGASI_SIP_TRANSPORT | default "UDP" }}
+{{ end }}
+{{ if and .Env.JIGASI_ENABLE_SDES_SRTP | default "0" | toBool }}
```
saghul: why do you need the `and` here?

jigasi/rootfs/defaults/sip-communicator.properties
```diff
@@ -11,16 +11,28 @@ net.java.sip.communicator.impl.protocol.sip.acc1=acc1
...
+net.java.sip.communicator.impl.protocol.sip.acc1.SAVP_OPTION=1
+net.java.sip.communicator.impl.protocol.sip.acc1.DEFAULT_ENCRYPTION=true
+net.java.sip.communicator.impl.protocol.sip.acc1.DEFAULT_SIPZRTP_ATTRIBUTE=false
+net.java.sip.communicator.impl.protocol.sip.acc1.ENCRYPTION_PROTOCOL.ZRTP=0
+net.java.sip.communicator.impl.protocol.sip.acc1.ENCRYPTION_PROTOCOL.SDES=1
+net.java.sip.communicator.impl.protocol.sip.acc1.ENCRYPTION_PROTOCOL.DTLS-SRTP=2
```
saghul: since we are here we should also make the use of DTLS configurable perhaps?

netaskd: I have no environment for DTLS, for now only SDES

jigasi/rootfs/defaults/sip-communicator.properties
netaskd: hmm, it doesnt need here

jigasi/rootfs/defaults/sip-communicator.properties
saghul: then lets set it to 0 then and disable AVPF

netaskd: do you mean  DTLS-SRTP=0 and SAVP_OPTION=false?

netaskd: changed DTLS options
- ENCRYPTION_PROTOCOL.DTLS-SRTP=0
- ENCRYPTION_PROTOCOL_STATUS.DTLS-SRTP=fasle

saghul: exactly

jigasi/rootfs/defaults/sip-communicator.properties
saghul: I still see the `and` here.

netaskd: magic :)  in a777d0e removed

PR #81, opened 2019-03-15
web: add option for join users with muted video

netaskd:
the option START_WITH_VIDEO_MUTED makes video muted when user join.


saghul:
I don't want to add an env variable for each option in config.js. Users can customize it and then restasrt the webserver.

Ideally I'd like to split config.js in the part that we need to template and the rest, so the user wouldn't need to touch the same config file, but I haven't found a good solution yet to merge both config files afterwards.

netaskd:
ok, I'll close it

saghul:
Thanks for understanding.

sapkra:
@saghul Maybe we can store the configs in two json files, then merge them (there some tools for it) and store the result behind the “var config =" as the config.js

saghul:
That could work!

PR #89, opened 2019-04-10
web: add config merger for config.js

sapkra:
In this PR I'm adding the possibility to change configurations in the `config.js` by adding a `config.json` to the `/config` volume.

We already discussed it in [this ticket](https://github.com/jitsi/docker-jitsi-meet/pull/81#issuecomment-473307587).


sapkra:
Alternatively we can also add the possibility to provide the custom `config.json` as an environment variable - I think this would be a bit more cloud native. 

The variable name could be `WEB_CONFIG` or `JM_CONFIG`.

@saghul What do you think?

But the best way to make the `config.js` cloud native would be to add an environment variable for each option.

saghul:
Sorry @sapkra I'm a bit behind on Docker PRs, as you might have seen. I hope to get to them soon.

saghul:
I've been thinking about this in the course of multiple showers, my favorite thinking moment.

The one thing I dislike about this approach is that JSON is a terrible format for editing configuration. The fact that we need to add the entire Python runtime just to strip some comments rubs the wrong way.

However, here is an idea! What if we stored the config in a sane format like YAML? Then we could combine the connection config and the user provided config and create a JSON out of it with the toJSON thing. I think it would work.

Now, this is not a decission to be taken lightly, so let's think about it a bit. WDYT @damencho ?

README.md
```diff
@@ -390,6 +390,11 @@ Variable | Description | Default value
 `ENABLE_HTTP_REDIRECT` | Redirects HTTP traffic to HTTPS | 1
 `LOG_LEVEL` | Controls which logs are output from prosody and associated modules | info
 
+#### Jitsi Meet `config.js`
```
luixxiul, line 393: This needs to be moved to https://github.com/jitsi/handbook/blob/master/docs/devops-guide/docker.md

sapkra: I will do this after it's clear how the final solution will look like.

sapkra:
superseded by #619

PR #104, opened 2019-06-17
web: add option CALENDAR_MS_APP_ID

netaskd:
The option uses for integrate Azure calendar to start web page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,21 @@ This can be tested using the [jwt.io] debugger. Use the following samople payloa
   "room": "*"
 }
 ```
+### Calendar integration
+
+#### Microsoft Azure calendar integration
+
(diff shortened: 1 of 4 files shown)


saghul:
I think I mentioned this in some other PR: I don't want to add an env var for every setting in config.js. The plan is to do something like https://github.com/jitsi/docker-jitsi-meet/pull/89 basically this setup will provide the most minimal config.js and merge it with the user provided one, so you could set any setting there.

We are pretty close to being able to do this. The plan would be to just fill in the `hosts` and `bosh` options. In order to do that all the options that are here https://github.com/jitsi/jitsi-meet/blob/master/config.js and are not commented out need to be commented out and the default value handled internally, then we can start.

PR #105, opened 2019-06-17
web,etherpad: add etherpad addon for sharing document

netaskd:


etherpad.yml
```diff
@@ -0,0 +1,8 @@
+version: '3'
+
+services:
+    # Etherpad: real-time collaborative document editing
+    etherpad:
+        image: jitsi/etherpad
+        networks:
+            meet.jitsi:
```
saghul: add an alias here so we can make it work automatically

web/rootfs/defaults/meet.conf
```diff
@@ -32,3 +32,13 @@ location /http-bind {
     proxy_set_header X-Forwarded-For $remote_addr;
     proxy_set_header Host {{ .Env.XMPP_DOMAIN }};
 }
+
+{{ if .Env.ETHERPAD_URL_BASE }}
+# Etherpad-lite
+location /etherpad/ {
```
saghul: I think we need some config here to allow for cross-origin requests. This is what I get otherwise:

Screen Shot 2019-06-20 at 20 50 46

web/rootfs/etc/cont-init.d/10-config
```diff
@@ -72,6 +72,12 @@ if [[ ! -f /config/config.js ]]; then
             -e "s#// authdomain:.*#authdomain: '${XMPP_DOMAIN}',#" \
             /config/config.js
     fi
+
+    if [[ ! -z "${ETHERPAD_URL_BASE}" && -z "$(grep -om1 'etherpad_base:' /config/config.js)" ]]; then
+        sed -i \
+            -e "/enableWelcomePage/a\    etherpad_base: '//${LETSENCRYPT_DOMAIN}/etherpad/p/'," \
```
saghul: I think we'd need the PUBLIC_URL thing here too

etherpad.yml
netaskd: @saghul, sorry I don't understand what do you mean. It's already works automatically because we set ETHERPAD_URL_BASE=http://etherpad:9001 and the service `etherpad` visible in network `meet.jitsi`

web/rootfs/defaults/meet.conf
netaskd: I could not find what raising it.
Upgrade connection for websocket added in 5f06ecc

saghul: Ack

etherpad/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM etherpad/etherpad
+
+ADD ./rootfs/defaults/settings.json /opt/etherpad-lite/settings.json
+
+EXPOSE 9001
```
saghul: Can we make this 80?

etherpad.yml
saghul: Yeah, you;re right, sorry.

env.example
```diff
@@ -222,3 +225,7 @@ JIGASI_PORT_MAX=20050
 
 # Redirects HTTP traffic to HTTPS. Only works with the standard HTTPS port (443).
 #ENABLE_HTTP_REDIRECT=1
+
+# Set etherpad-lite URL
+#ETHERPAD_URL_BASE=http://etherpad.meet.jitsi:9001
```
saghul: I'd like to avoid the port here

web/rootfs/etc/cont-init.d/10-config
```diff
@@ -72,6 +72,12 @@ if [[ ! -f /config/config.js ]]; then
             -e "s#// authdomain:.*#authdomain: '${XMPP_DOMAIN}',#" \
             /config/config.js
     fi
+
+    if [[ ! -z "${ETHERPAD_URL_BASE}" && -z "$(grep -om1 'etherpad_base:' /config/config.js)" ]]; then
+        sed -i \
+            -e "/enableWelcomePage/a\    etherpad_base: '${PUBLIC_URL}/etherpad/p/'," \
```
saghul: have you tested if setting the etherpad_base option to just `/etherpad/p/` works ok? I think we shouldn't need the full URL there, and this way we can avoid having the PUBLIC_URL thing altogether.

etherpad/Dockerfile
sapkra: That would be a bad idea.

From [OpenShift docs](https://docs.openshift.com/container-platform/3.11/creating_images/guidelines.html):

> Because the container user is always a member of the root group, the container user can read and write these files. The root group does not have any special permissions (unlike the root user) so there are no security concerns with this arrangement. In addition, the processes running in the container must not listen on privileged ports (ports below 1024), since they are not running as a privileged user.

saghul: Ah, thanks for the insight! Let's keep the default one then.

env.example
saghul: Ignore this.

web/rootfs/etc/cont-init.d/10-config
netaskd: You are right. It works as absolute path instead url. Was changed in 253e638

saghul: Fantastic! The we can also remove the PUBLIC_URL variable from this PR, since it's no longer needed here, right? (we need it for the transcriber settings, right?)

netaskd: Yes, you are right. Commit b003fc9 will remove it.

saghul:
This is awesome @netaskd! Landed in 62f2d11 squashed and with some mino tweaks to the docs, Makefile.

I had to add an alias for the etherpad container or nginx would complain:

```
web_1       | nginx: [emerg] host not found in upstream "etherpad.meet.jitsi" in /config/nginx/meet.conf:45
web_1       | nginx: [emerg] host not found in upstream "etherpad.meet.jitsi" in /config/nginx/meet.conf:45
```

Thanks a ton for your continuos help with this project! ❤️

rplancarte:
I'm getting the `nginx: [emerg] host not found in upstream "etherpad.meet.jitsi" in /config/nginx/meet.conf:57` error when using etherpad. I can't find what you did to add an alias for it on the README. Is it just creating a network alias for the container? Do I have to point thing to that alias somewhere? 

Thanks a lot, all this work is awesome!

saghul:
Please open a new issue.

netaskd:
@rplancarte, do you run th setup like: `docker-compose -f docker-compose.yml -f etherpad.yml up` ?
alias etherpad.meet.jitsi exists in etherpad.yml

rplancarte:
Thanks for the help, I'll look further into this today and will create a new issue if I keep getting the error.

PR #106, opened 2019-06-17
misc: refactoring Makefile

netaskd:
PR makes Makefile more flexible and simple.
* Variable `JITSI_REPO` - can be set to you private registry path (default jitsi dockerhub)
* Variable `JITSI_SERVICES` - contains all services that need to build/tag/push
* Target `all` will build all services then tag it from default name to custom name and push to registry
* Target pattern `%-all` will `build-all`/`tag-all`/`push-all` services described on `JITSI_SERVICES`
* Command `make JITSI_SERVICE=base push` will push only selected service.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,49 +1,31 @@
 FORCE_REBUILD ?= 0
-JITSI_RELEASE ?= "stable"
-JITSI_BUILD ?= "latest"
+JITSI_RELEASE ?= stable
+JITSI_BUILD ?= latest
+JITSI_REPO ?= jitsi
+JITSI_SERVICES ?= base base-java web prosody jicofo jvb jigasi
 
 ifeq ($(FORCE_REBUILD), 1)
   BUILD_ARGS = "--no-cache"
 endif
 
-build-all:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,49 +1,31 @@
...
-	docker push jitsi/jigasi:latest
-	docker push jitsi/base:$(JITSI_BUILD)
-	docker push jitsi/base-java:$(JITSI_BUILD)
-	docker push jitsi/web:$(JITSI_BUILD)
-	docker push jitsi/prosody:$(JITSI_BUILD)
-	docker push jitsi/jicofo:$(JITSI_BUILD)
-	docker push jitsi/jvb:$(JITSI_BUILD)
-	docker push jitsi/jigasi:$(JITSI_BUILD)
 
+all:	build-all tag-all push-all
```
saghul: Let's make the default target not tag and push, since it's what people use for testing. Ldt's have `all` just be `build-all` and a new `release` target which does all of it.

netaskd: Good idea! New target added.

saghul:
Thanks a lot for this!

PR #111, opened 2019-06-20
Fixes prosody create user command in readme

### LINKED ISSUE #109: Failed to create prosody users
I'd like to use auth. However, I fail to create prosody users:

```
$ prosodyctl --config /config/prosody.cfg.lua register user myuser meet.jitsi mypassword
The given hostname does not exist in the config
```

Any ideas?

PS: the docs are wrong, I needed `docker exec -it jitsi_prosody_1 /bin/bash` to get into the container

arnuschky:
Fixes #109

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,9 @@ In order to do that, first execute a shell in the corresponding container:
 
 Once in the container, run the following command to create a user:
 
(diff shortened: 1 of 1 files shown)


saghul:
Merged in dev, thank you!

PR #115, opened 2019-07-08
fix prosody certs directory ownership

kegato:
Hi,
in some environments I'm getting permission errors when creating the prosody container:
```
prosody_1  | The directory /config/data is not owned by the current user, won't be able to write files to it
prosody_1  | mv: cannot stat '/config/data/*.crt': No such file or directory
prosody_1  | mv: cannot stat '/config/data/*.key': No such file or directory
prosody_1  | The directory /config/data is not owned by the current user, won't be able to write files to it
...
prosody_1  | certmanager         error	SSL/TLS: Failed to load '/config/certs/meet.jitsi.key': Check that the path is correct, and the file exists. (for meet.jitsi)
prosody_1  | meet.jitsi:tls      error	Error creating context for c2s: error loading private key (No such file or directory)
```
This is happening because an ownership check in prosody fails. To solve this I replaced the chmod 777 in the prosody init script with a chown and added a chown for the certs directory.

### DIFF
--- a/prosody/rootfs/etc/cont-init.d/10-config
+++ b/prosody/rootfs/etc/cont-init.d/10-config
@@ -4,7 +4,7 @@ PROSODY_CFG="/config/prosody.cfg.lua"
 
 if [[ ! -d /config/data ]]; then
     mkdir -p /config/data
-    chmod 777 /config/data
+    chown -R prosody:prosody /config/data
 fi
 
 if [[ ! -f $PROSODY_CFG ]]; then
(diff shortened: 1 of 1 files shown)


saghul:
This should be fixed in `dev` already, can you give it a go?

kegato:
> This should be fixed in `dev` already, can you give it a go?
Yes, I've just tested it and it works with the init script from dev. Sorry, I didn't know there was a dev branch...

saghul:
No worries. I'll update the README so the dev branch is more prominent.

PR #116, opened 2019-07-08
add letsencrypt volume

kegato:
Hi,
I think it would make sense to declare `/etc/letsencrypt` as a volume so the certificate is preserved on container upgrades.


saghul:
Landed in the `dev` branch, cheers!

PR #125, opened 2019-07-26
changes the test for "unstable" to proper /bin/sh syntax

r900:
i noticed that in the build of the base image the check for $JITSI_RELEASE as always failing.
this happens because with `/bin/sh` a single `=` has to be used, otherwise it will always fail with "unexpected operator".

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -23,7 +23,7 @@ RUN \
 	chmod +x /usr/bin/frep
 
 RUN \
-	[[ "$JITSI_RELEASE" == "unstable" ]] && \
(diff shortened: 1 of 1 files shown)


sapkra:
I also noticed it and moved the script to an `install-devtools.sh` file. It's part of my OpenShift PR #126 .

saghul:
Why close it though?

r900:
@saghul it was early in the morning and i realized i hadn't tested if it still works with bash - which, of course, it does. reopening.

saghul:
Merged, cheers!

PR #130, opened 2019-08-08
doc,env,jicofo,jvb,web: add configuration for octo cascaded bridges

netaskd:
This PR adds configuration for using Octo (cascaded bridges) in docker infrastructure.
Based on doc: https://github.com/jitsi/jitsi-videobridge/blob/master/doc/octo.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,6 +268,36 @@ Variable | Description | Default value
 `DISABLE_HTTPS` | Disable HTTPS, this can be useful if TLS connections are going to be handled outside of this setup | 1
 `ENABLE_HTTP_REDIRECT` | Redirects HTTP traffic to HTTPS | 1
 
+#### Setting up Octo (cascaded bridges)
+NOTE: For get working octo properly you have to set header "X-User-Region" before it passing to nginx. It can be realized via geoip or another logic and it's not described here.
+
+The header "X-User-Region" will be passed through nginx and dynamically set variable for `userRegion` in config.js file via ssi nginx module.
+If `userRegion` and `JVB_OCTO_REGION` the same region, user will be connected to the instanse JVB that has this region.
+
+This behavior already pre-configured and receiving header "X-User-Region" in config.js looks like this:
+
+```
+deploymentInfo: {
+    // shard: "shard1",
+    // region: europe" -->',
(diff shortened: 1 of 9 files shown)


jvb/rootfs/etc/services.d/jvb/run
```diff
@@ -2,9 +2,14 @@
 
 JAVA_SYS_PROPS="-Dnet.java.sip.communicator.SC_HOME_DIR_LOCATION=/ -Dnet.java.sip.communicator.SC_HOME_DIR_NAME=config -Djava.util.logging.config.file=/config/logging.properties"
 
-if [[ ! -z "$DOCKER_HOST_ADDRESS" ]]; then
-    LOCAL_ADDRESS=$(hostname -I | cut -d " " -f1)
-    JAVA_SYS_PROPS="$JAVA_SYS_PROPS -Dorg.ice4j.ice.harvest.NAT_HARVESTER_LOCAL_ADDRESS=$LOCAL_ADDRESS -Dorg.ice4j.ice.harvest.NAT_HARVESTER_PUBLIC_ADDRESS=$DOCKER_HOST_ADDRESS"
+PUBLIC_ADDRESS="$(curl -4qs --connect-timeout 15 icanhazip.com)"
```
sapkra: Because you've removed the `DOCKER_HOST_ADDRESS` it will not be possible to run the container without an internet connection in a LAN environment anymore I think. Additionally the docs for this env var are still there.

netaskd:
Be careful, jvb2 does not support some configure options from the run string.
We need to replace logic from jvb/rootfs/etc/cont-init.d/10-config
to
jvb/rootfs/defaults/sip-communicator.properties

README.md
```diff
@@ -268,6 +268,36 @@ Variable | Description | Default value
 `DISABLE_HTTPS` | Disable HTTPS, this can be useful if TLS connections are going to be handled outside of this setup | 1
 `ENABLE_HTTP_REDIRECT` | Redirects HTTP traffic to HTTPS | 1
 
+#### Setting up Octo (cascaded bridges)
```
luixxiul, line 271: This needs to be moved to https://github.com/jitsi/handbook/blob/master/docs/devops-guide/docker.md

saghul:
I believe all the necessary options are configurable now.

PR #131, opened 2019-08-08
prosody: refactor getting prosody-plugins.

netaskd:
Now we get all files from the deb package jitsi-meet-tokens into /prosody-plugins

### DIFF
--- a/prosody/Dockerfile
+++ b/prosody/Dockerfile
@@ -1,11 +1,5 @@
 FROM jitsi/base
 
-ADD https://raw.githubusercontent.com/jitsi/jitsi-meet/fc129d9849ca5e26245d54df6451931b6c179987/resources/prosody-plugins/token/util.lib.lua /prosody-plugins/token/util.lib.lua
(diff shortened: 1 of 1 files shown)


prosody/Dockerfile
```diff
@@ -31,7 +25,18 @@ RUN \
       libssl1.0-dev \
       liblua5.2-dev \
     && apt-cleanup \
-    && rm -rf /etc/prosody
+    && rm -rf /etc/prosody /var/cache/apt
```
saghul: any reason to duplicate the cleanup steps? we could run them once, at the end

netaskd:
Because every layer is persistent. So for get less image we have to clean unused files in which layer.

prosody/Dockerfile
netaskd: Because every layer is persistent. So for get less image we have to clean unused files in which layer.

saghul: Ah, true, I always forget.

PR #132, opened 2019-08-08
prosody: add speakerstats module handler

netaskd:
If the `speakerstats` module is enabled in variable `XMPP_MUC_MODULES`, the properly handling it will already preconfigured. This PR depends on #131


saghul:
Ops, forgot to close this, speaker stats is now properly integrated and enabled by default.

PR #133, opened 2019-08-08
prosody: apply muc_owner_allow_kick.patch if muc_allowners is enabled

netaskd:
Check `XMPP_MUC_MODULES` variable and if module `muc_allowners` is enabled, then apply muc_owner_allow_kick.patch from /prosody-plugins directory. This PR depends on #131

### DIFF
--- a/prosody/Dockerfile
+++ b/prosody/Dockerfile
@@ -14,6 +14,7 @@ RUN \
       luarocks \
       git \
       gcc \
+      patch \
     && luarocks install cyrussasl 1.1.0-1 \
     && luarocks install lua-cjson 2.1.0-1 \
     && luarocks install luajwtjitsi 1.3-7 \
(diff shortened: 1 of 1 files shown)


prosody/rootfs/etc/cont-init.d/10-config
```diff
@@ -29,6 +29,9 @@ if [[ "$(stat -c %U /prosody-plugins-custom)" != "prosody" ]]; then
     chown -R prosody /prosody-plugins-custom
 fi
 
+grep -q muc_allowners <<< ${XMPP_MUC_MODULES} \
+&& patch -d /usr/lib/prosody/modules/muc -p0 < /prosody-plugins/muc_owner_allow_kick.patch
```
saghul: why don't we unconditionally patch it? I know it only has an effect if that plugin is enabled, but it would simplify things, and we can do it at container build time.

netaskd: I agree with you. Moved to Docker file for unconditional patching in 65e3c89

saghul:
Cheers!

PR #126, opened 2019-07-27
WIP: OpenShift Support

sapkra:
@lsytnik @saghul Fixes #120 

I'm using Jitsi with OpenShift for some time and I'm ready to contribute my change now.

The problem with these Dockerfiles are that they are running as root which is not allowed in OpenShift for security reasons. One big problem while fixing the permissions are always the supervisor tools they are always running as root and are really hard to fix - so I had to add some workarounds. I hope we can get rid of s6 in the future but this is just my opinion.

I would appreciate if someone is able to test the changes with docker-compose and OpenShift and can provide some feedback.

My environment:
OpenShift/OKD 3.11

TODO
- [ ] Make it possible to use Let'sEncrypt with cronjobs
- [ ] Check if etherpad is working

PR #134, opened 2019-08-09
jigasi: generate google cloud credentials from env vars

sapkra:
When using the transcription feature in an cloud environment like kubernetes the credentials should be saved in a secret. This change makes it possible to do exactly this.

A cloud native container should be configurable via env variables instead of manually adding config files.

need this for #126

### DIFF
--- a/jigasi/Dockerfile
+++ b/jigasi/Dockerfile
@@ -1,9 +1,11 @@
 ARG JITSI_REPO=jitsi
 FROM ${JITSI_REPO}/base-java
 
+ENV GOOGLE_APPLICATION_CREDENTIALS /config/key.json
+
 RUN \
 	apt-dpkg-wrap apt-get update && \
-	apt-dpkg-wrap apt-get install -y jigasi && \
+	apt-dpkg-wrap apt-get install -y jigasi jq && \
 	apt-cleanup
 
 COPY rootfs/ /
--- a/jigasi/rootfs/etc/cont-init.d/10-config
+++ b/jigasi/rootfs/etc/cont-init.d/10-config
@@ -10,3 +10,32 @@ fi
 
 mkdir -pm777 /tmp/transcripts
 chown jigasi:jitsi /tmp/transcripts
+
+# Create Google Cloud Credentials
+if [[ $ENABLE_TRANSCRIPTIONS -eq 1 || $ENABLE_TRANSCRIPTIONS == "true" ]] && [[ ! -f /config/key.json ]]; then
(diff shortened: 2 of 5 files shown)


sapkra:
@netaskd You mean that I've added the paths to the `VOLUME` in the dockerfiles?
I just did it to have it documented. It's already used as a persistent volume so because this is not really related to this PR you should rather open a new issue for it.

sapkra:
@saghul Can you review this one please. It would simplify the rebase of my OpenShift PR to split it afterwards.

jigasi/Dockerfile
```diff
@@ -1,11 +1,13 @@
...
 RUN \
 	apt-dpkg-wrap apt-get update && \
-	apt-dpkg-wrap apt-get install -y jigasi && \
+	apt-dpkg-wrap apt-get install -y jigasi jq && \
 	apt-cleanup
 
 COPY rootfs/ /
 
-VOLUME /config
+VOLUME ["/config", "/tmp/transcripts"]
```
saghul: side question, shouldn't we have this in / or /config? /tmp doesn't seem like the right place for it...

jigasi/rootfs/etc/cont-init.d/10-config
```diff
@@ -10,3 +10,27 @@ fi
 
 mkdir -pm777 /tmp/transcripts
 chown jigasi:jitsi /tmp/transcripts
+
+# Create Google Cloud Credentials
+if [[ $ENABLE_TRANSCRIPTIONS -eq 1 || $ENABLE_TRANSCRIPTIONS == "true" ]] && [[ ! -f /config/key.json ]]; then
+    jq -n \
+        --arg GC_PROJECT_ID "$GC_PROJECT_ID" \
```
saghul: souldn't we check for the existence of all these env vars before building the json file and error if any is missing?

web/Dockerfile
```diff
@@ -13,4 +13,4 @@ COPY rootfs/ /
 
 EXPOSE 80 443
 
-VOLUME ["/config", "/etc/letsencrypt"]
+VOLUME ["/config", "/etc/letsencrypt", "/usr/share/jitsi-meet/transcripts"]
```
saghul: why do we need this new volume?

jigasi/Dockerfile
sapkra: As I applied the changes for this PR I just noticed that he transcripts are saved inside the container without having documented the directories as volumes. So I just added them but yeah they are not at the right place in this PR. I will remove them.

web/Dockerfile
sapkra: This volume is used inside the `docker-compose.yml` but not documented in the `Dockerfile`. I will create a new small PR for this.

jigasi/rootfs/etc/cont-init.d/10-config
sapkra: I've added a check for them now.

PR #135, opened 2019-08-14
Install missing lua module `net-url`

poying:


saghul:
Where is this used?

poying:
@saghul  https://github.com/jitsi/jitsi-meet/blob/d12afc5c071184a61c8e42578be21e84ef9b6324/resources/prosody-plugins/mod_muc_size.lua#L13

saghul:
Thank you!

PR #140, opened 2019-09-03
misc: add variable JITSI_REPO instead hardcoded jitsi

netaskd:
It fixes broken tagging for different from "jitsi" JITSI_REPO


netaskd:
@saghul very fast merge =)
For work properly it needs also #141

PR #141, opened 2019-09-03
misc: add ARG JITSI_REPO for dependented dockerfiles

netaskd:
in addition to 2b30ab9 It finally covering logic for different from `jitsi` docker repo


saghul:
Cheers!

PR #144, opened 2019-09-19
WIP: Initial Spot support

saghul:


spot/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+	apt-cleanup && \
+	rm -f /etc/nginx/conf.d/default.conf
+
+COPY rootfs/ /
+COPY --from=builder /build/spot-client/dist /usr/share/spot-client/dist
+COPY --from=builder /build/spot-client/index.html /usr/share/spot-client/
+
+EXPOSE 9002
+
+VOLUME ["/config"]
```
sapkra: Can you please add `USER 1001` to prevent running this container as root per default otherwise I have to fix it in my PR. Would be great if new images are already using the security standard.

spot/rootfs/defaults/nginx.conf
```diff
@@ -0,0 +1,61 @@
+user www-data;
```
sapkra: I think this will cause problems when running without root.

README.md
```diff
@@ -76,6 +76,7 @@ several container images are provided.
 * **jigasi**: [Jigasi], the SIP (audio only) gateway.
 * **etherpad**: [Etherpad], shared document editing addon.
 * **jibri**: [Jibri], the brooadcasting infrastructure.
+* **spot**: [Spot], physical meeting rooms solution.
```
luixxiul, line 79: This change needs to be applied to https://github.com/jitsi/handbook/blob/master/docs/devops-guide/docker.md (cf: #601)

spot/rootfs/defaults/config.js
```diff
@@ -0,0 +1,36 @@
...
+    },
+    MEETING_DOMAINS_WHITELIST: [
+        '{{ $SPOT_DOMAIN }}',
+        'meet.jit.si'
+    ],
+    XMPP_CONFIG: {
+        bosh: '{{ .Env.PUBLIC_URL }}/http-bind',
+        hosts: {
+            domain: '{{ .Env.XMPP_DOMAIN }}',
+            muc: '{{ .Env.XMPP_MUC_DOMAIN }}'
```
saghul, line 33: Note to self: use a dedicated MUC for Spot.

PR #148, opened 2019-09-26
Allow disable P2P

madmath03:
:sparkles: Allow disable of P2P mode.

This can be useful if connections are often too unstable for P2P


saghul:
Hi there, thanks for your contribution!

I don't want to add an env variable for every single option in config.js, it would get messy very quickly.

This is a rough idea of the direction I'd like to take: https://github.com/jitsi/docker-jitsi-meet/pull/89

It hasn't materialized yet, but I think it will scale better with time than having to add an env variable for every option we have right now or we could have in the future.

PR #152, opened 2019-10-26
fix typo in example

sunsingerus:


saghul:
That's how a sub-section header should look like.

sapkra:
@saghul  But there is still a typo in the word `auuthentication`. It has to be `authentication`.

saghul:
🤦‍♂

PR #153, opened 2019-11-03
Allow dashes in roomnames

hnicke:
Currently it's not possible to have dashes in roomnames.
E.g., `https://meet.jit.si/test-room` works, but when using the web docker image, it doesn't work.

This patch allows dashes in roomnames.

### DIFF
--- a/web/rootfs/defaults/meet.conf
+++ b/web/rootfs/defaults/meet.conf
@@ -6,7 +6,7 @@ root /usr/share/jitsi-meet;
 index index.html
 error_page 404 /static/404.html;
 
-location ~ ^/([a-zA-Z0-9=\?]+)$ {
+location ~ ^/([a-zA-Z0-9=\?\-]+)$ {
     rewrite ^/(.*)$ / break;
 }
 
@@ -27,7 +27,7 @@ location / {
 }
 
 # BOSH
-location /http-bind {
+location = /http-bind {
     proxy_pass {{ .Env.XMPP_BOSH_URL_BASE }}/http-bind;
     proxy_set_header X-Forwarded-For $remote_addr;
     proxy_set_header Host {{ .Env.XMPP_DOMAIN }};


web/rootfs/defaults/meet.conf
```diff
@@ -27,7 +27,7 @@ location / {
 }
 
 # BOSH
-location /http-bind {
+location = /http-bind {
```
hnicke: Need an exact match here, otherwise the regex from above would have precedence.

damencho:
We recently updated those for default deb installation from packages, maybe we need to sync it with it https://github.com/jitsi/jitsi-meet/blob/master/doc/debian/jitsi-meet/jitsi-meet.example

saghul:
Good point @damencho ! @hnicke can you please sync it up?

D3473R:
Any news on this? This is blocking for me, as I want to use uuids as room names 😢

saghul:
Fixed in https://github.com/jitsi/docker-jitsi-meet/pull/245

D3473R:
Sadly still not working for me. Getting a 404 as before.
Testing with Room Name `dash-test`

The newest version is pulled and running:
```
root@jitsi:~/docker-jitsi-meet# docker-compose exec web cat /defaults/meet.conf
server_name _;

client_max_body_size 0;

root /usr/share/jitsi-meet;

# ssi on with javascript for multidomain variables in config.js
ssi on;
ssi_types application/x-javascript application/javascript;
...
```

Container logs:
```
web_1      | XXX.XXX.XXX.XXX - - [24/Mar/2020:14:50:45 +0100] "GET /dash-test HTTP/1.1" 404 200 "https://meet.example.com/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.149 Safari/537.36"
web_1      | 2020/03/24 14:50:45 [error] 219#219: *28 open() "/usr/share/jitsi-meet/dash-test" failed (2: No such file or directory), client: XXX.XXX.XXX.XXX, server: _, request: "GET /dash-test HTTP/1.1", host: "meet.example.com", referrer: "https://meet.example.com/"
```

saghul:
I just tested it locally and it works as expected. Have you pulled the latest images and removed the CONFIG volume?

D3473R:
I've pulled the latest image but had not removed the CONFIG volume 🤦‍♂ 
It is working fine now. Thanks 👍

saghul:
Great news!

Issue #22, opened 2018-10-23
Make usable behind reverse proxy (like https://meet.jit.si/)


nicolaspernoud:
https://meet.jit.si/ is usable behind a corporate reverse proxy (only TCP ports 80 and 443 allowed).
It would be amazing if the self hosted dockerised version would be as well...

Thanks.

Best regards and kudos for the fantastic work...

saghul:
@damencho We could also use that nginx module to proxy to a different upstream based on ALPN, right? That would be to do TURN on 443.

damencho:
Maybe, we need to test it, whether it works :)

damencho:
Nope, there is nothing we had tried, here is some stuff that I was reading:
https://superuser.com/questions/1135208/can-nginx-serve-ssh-and-https-at-the-same-time-on-the-same-port
http://nginx.org/en/docs/stream/ngx_stream_ssl_preread_module.html#ssl_preread
The idea is nginx to detect https web traffic and serve the web (or bosh), but the rest to be considered as media traffic and forwarded to coturn tcp port, or the other way around, detect coturn traffic and forward it and the rest consider to be served by nginx. This way we can use port 443 for both serving the web and media.

alexcustos:
It seems I ran into the same issue. It's described well here https://github.com/jitsi/jitsi-meet/issues/2805. I tried to use Turncredentials module to point my Coturn server, but Prosody crashes immediately after login, even with this module disabled. It's enough to have it in modules path.

So I would like to know if there is any way to enable TURN server for this Docker Compose setup?

<details>
<summary>Here's the relevant part of the log file</summary>

```
jitsi_prosody | mod_bosh                                  info  New BOSH session, assigned it sid 'c9307ab7-ad92-4493-992c-32cc6443b207'
jitsi_prosody | boshc9307ab7-ad92-4493-992c-32cc6443b207  info  Authenticated as hcisowftdjhb22or@guest.meet.jitsi
jitsi_prosody | mod_bosh                                  error Traceback[bosh]: /usr/lib/prosody/util/stanza.lua:62: invalid attribute value: expected string, got number
jitsi_prosody | stack traceback:
jitsi_prosody |         [C]: in function 'error'
jitsi_prosody |         /usr/lib/prosody/util/stanza.lua:62: in function 'check_text'
jitsi_prosody |         /usr/lib/prosody/util/stanza.lua:75: in function 'check_attr'
[...]

PR #163, opened 2019-12-11
turn: add TURN server

netaskd:
adds TURN server to the infrastructure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,27 @@ The diagram shows a typical deployment in a host running Docker. This project
 separates each of the components above into interlinked containers. To this end,
 several container images are provided.
 
+### External Ports
+
+The following external ports must be opened on a firweall:
+
+* 80/tcp for Web UI HTTP (really just to redirect, after uncommenting ENABLE_HTTP_REDIRECT=1 in .env)
+* 443/tcp for Web UI HTTPS
+* 4443/tcp for RTP media over TCP
+* 10000/udp for RTP media over UDP
+
+Also 20000-20050/udp for jigasi, in case you choose to deploy that to facilitate SIP acces.
+
+E.g. on a CentOS/Fedora server this would be done like this (without SIP access):
+
+```shell
+    $ sudo firewall-cmd --permanent --add-port=80/tcp
+    $ sudo firewall-cmd --permanent --add-port=443/tcp
+    $ sudo firewall-cmd --permanent --add-port=4443/tcp
+    $ sudo firewall-cmd --permanent --add-port=10000/udp
+    $ sudo firewall-cmd --reload
+```
+
 ### Images
 
 * **base**: Debian stable base image with the [S6 Overlay] for process control and the
--- a/README.md
+++ b/README.md
@@ -349,14 +349,13 @@ Variable | Description | Example
 For setting `GOOGLE_APPLICATION_CREDENTIALS` please read https://cloud.google.com/text-to-speech/docs/quickstart-protocol section "Before you begin" from 1 to 5 paragraph.
 
 ### TURN(S) server
-For enable turn server for P2P and JVB connections, please add to the variable `GLOBAL_MODULES` string `turncredentials` and set variables below
+For enable turn server for P2P and JVB connections, please set variables below
 
 Variable | Description | Default value
 --- | --- | ---
 `TURN_ENABLE` | Use TURN for P2P and JVB (bridge mode) connections | 0
 `TURN_REALM` | Realm to be used for the users with long-term credentials mechanism or with TURN REST API | realm
 `TURN_SECRET` | Secret for connect to TURN server | keepthissecret
-`TURN_TYPE` | Type of TURN(s) (turn/turns) | turns
 `TURN_HOST` | Annonce FQDN/IP address of the turn server via XMPP (XEP-0215) | 192.168.1.1
 `TURN_PUBLIC_IP` | Public IP address for an instance of turn server | set dynamically
 `TURN_PORT` | TLS/TCP/UDP turn port for connection | 5349
--- a/README.md
+++ b/README.md
@@ -47,6 +47,9 @@ If you want to use jibri too, first configure host as described in JItsi BRoadca
 and then run Docker Compose as follows: ``docker-compose -f docker-compose.yml -f jibri.yml up -d``
 or to use jigasi too: ``docker-compose -f docker-compose.yml -f jigasi.yml -f jibri.yml up -d``
 
+If you want to enable TURN server, configure it and run Docker Compose as
+follows: ``docker-compose -f docker-compose.yml -f turn.yml up``
+
 ## Architecture
 
 A Jitsi Meet installation can be broken down into the following components:
@@ -76,6 +79,7 @@ several container images are provided.
 * **jigasi**: [Jigasi], the SIP (audio only) gateway.
 * **etherpad**: [Etherpad], shared document editing addon.
 * **jibri**: [Jibri], the brooadcasting infrastructure.
+* **turn**: [Coturn], the TURN server.
 
 ### Design considerations
 
@@ -344,6 +348,27 @@ Variable | Description | Example
 
 For setting `GOOGLE_APPLICATION_CREDENTIALS` please read https://cloud.google.com/text-to-speech/docs/quickstart-protocol section "Before you begin" from 1 to 5 paragraph.
 
+### TURN(S) server
+For enable turn server for P2P and/or JVB connections, please set variables below
+
+Variable | Description | Default value
+--- | --- | ---
+`TURN_ENABLE_P2P` | Use TURN for P2P connections | 0
+`TURN_ENABLE_JVB` | Use TURN for JVB (bridge mode) connections | 0
+`TURN_REALM` | Realm to be used for the users with long-term credentials mechanism or with TURN REST API | realm
+`TURN_SECRET` | Secret for connect to TURN server | keepthissecret
+`TURN_ADMIN_USER` | Username for admin panel | admin
+`TURN_ADMIN_SECRET` | Password for admin panel | changeme
+`TURN_ADMIN_PORT` | HTTP(s) port for acess to admin panel | 8443
+`TURN_TYPE` | Type of TURN(s) (turn/turns) | turns
+`TURN_HOST` | Annonce FQDN or IP address of turn server | 8.8.8.8
+`TURN_PORT` | TLS/TCP/UDP turn port for connection | 5349
+`TURN_TRANSPORT` | transport for turn connection (tcp/udp) | tcp
+`TURN_RTP_MIN` | RTP start port for turn/turns connections | 10000
+`TURN_RTP_MAX` | RTP end port for turn/turns connections | 11000
+
+And add to the variable `GLOBAL_MODULES` string `turncredentials`.
+
 ### Advanced configuration
 
 These configuration options are already set and generally don't need to be changed.
@@ -403,7 +428,6 @@ option.
 ## TODO
 
 * Support container replicas (where applicable).
-* TURN server.
 
 [Jitsi]: https://jitsi.org/
 [Jitsi Meet]: https://jitsi.org/jitsi-meet/
@@ -421,3 +445,4 @@ option.
 [jwt.io]: https://jwt.io/#debugger-io
 [Etherpad]: https://github.com/ether/etherpad-lite
 [Jibri]: https://github.com/jitsi/jibri
+[Coturn]: https://github.com/coturn/coturn
--- a/README.md
+++ b/README.md
@@ -349,25 +349,30 @@ Variable | Description | Example
 For setting `GOOGLE_APPLICATION_CREDENTIALS` please read https://cloud.google.com/text-to-speech/docs/quickstart-protocol section "Before you begin" from 1 to 5 paragraph.
 
 ### TURN(S) server
-For enable turn server for P2P and/or JVB connections, please set variables below
+For enable turn server for P2P and JVB connections, please add to the variable `GLOBAL_MODULES` string `turncredentials` and set variables below
 
 Variable | Description | Default value
 --- | --- | ---
-`TURN_ENABLE_P2P` | Use TURN for P2P connections | 0
-`TURN_ENABLE_JVB` | Use TURN for JVB (bridge mode) connections | 0
+`TURN_ENABLE` | Use TURN for P2P and JVB (bridge mode) connections | 0
 `TURN_REALM` | Realm to be used for the users with long-term credentials mechanism or with TURN REST API | realm
 `TURN_SECRET` | Secret for connect to TURN server | keepthissecret
-`TURN_ADMIN_USER` | Username for admin panel | admin
-`TURN_ADMIN_SECRET` | Password for admin panel | changeme
-`TURN_ADMIN_PORT` | HTTP(s) port for acess to admin panel | 8443
 `TURN_TYPE` | Type of TURN(s) (turn/turns) | turns
-`TURN_HOST` | Annonce FQDN or IP address of turn server | 8.8.8.8
+`TURN_HOST` | Annonce FQDN/IP address of the turn server via XMPP (XEP-0215) | 192.168.1.1
+`TURN_PUBLIC_IP` | Public IP address for an instance of turn server | set dynamically
 `TURN_PORT` | TLS/TCP/UDP turn port for connection | 5349
 `TURN_TRANSPORT` | transport for turn connection (tcp/udp) | tcp
 `TURN_RTP_MIN` | RTP start port for turn/turns connections | 10000
 `TURN_RTP_MAX` | RTP end port for turn/turns connections | 11000
 
-And add to the variable `GLOBAL_MODULES` string `turncredentials`.
+
+For enable web-admin panel for turn, please set variables below
+
+Variable | Description | Default value
+--- | --- | ---
+`TURN_ADMIN_ENABLE` | Enable web-admin panel | 0
+`TURN_ADMIN_USER` | Username for admin panel | admin
+`TURN_ADMIN_SECRET` | Password for admin panel | changeme
+`TURN_ADMIN_PORT` | HTTP(s) port for acess to admin panel | 8443
 
 ### Advanced configuration
 
--- a/README.md
+++ b/README.md
@@ -360,8 +360,8 @@ Variable | Description | Default value
 `TURN_PUBLIC_IP` | Public IP address for an instance of turn server | set dynamically
 `TURN_PORT` | TLS/TCP/UDP turn port for connection | 5349
 `TURN_TRANSPORT` | transport for turn connection (tcp/udp) | tcp
-`TURN_RTP_MIN` | RTP start port for turn/turns connections | 10000
-`TURN_RTP_MAX` | RTP end port for turn/turns connections | 11000
+`TURN_RTP_MIN` | RTP start port for turn/turns connections | 16000
+`TURN_RTP_MAX` | RTP end port for turn/turns connections | 17000
 
 
 For enable web-admin panel for turn, please set variables below
--- a/README.md
+++ b/README.md
@@ -75,6 +75,9 @@ The following external ports must be opened on a firweall:
 * 443/tcp for Web UI HTTPS
 * 4443/tcp for RTP media over TCP
 * 10000/udp for RTP media over UDP
+* 5349/tcp for TURN data over TCP
+* 5349/udp for TURN data over UDP
+* 16000-17000/udp for TURN RTP media over UDP
 
 Also 20000-20050/udp for jigasi, in case you choose to deploy that to facilitate SIP acces.
 
@@ -85,6 +88,9 @@ E.g. on a CentOS/Fedora server this would be done like this (without SIP access)
     $ sudo firewall-cmd --permanent --add-port=443/tcp
     $ sudo firewall-cmd --permanent --add-port=4443/tcp
     $ sudo firewall-cmd --permanent --add-port=10000/udp
+    $ sudo firewall-cmd --permanent --add-port=5349/tcp
+    $ sudo firewall-cmd --permanent --add-port=5349/udp
+    $ sudo firewall-cmd --permanent --add-port=16000-17000/udp
     $ sudo firewall-cmd --reload
 ```
 
--- a/env.example
+++ b/env.example
@@ -294,3 +294,43 @@ JIBRI_LOGS_DIR=/config/logs
 
 # Redirects HTTP traffic to HTTPS. Only works with the standard HTTPS port (443).
 #ENABLE_HTTP_REDIRECT=1
+#
+## Use TURN for P2P connections
+##TURN_ENABLE_P2P=0
+#
+## Use TURN for JVB (bridge mode) connections
+##TURN_ENABLE_JVB=0
+#
+## Realm to be used for the users with long-term credentials mechanism or with TURN REST API
+##TURN_REALM=realm
+#
+## Secret for connect to TURN server
+##TURN_SECRET=keepthissecret
+#
+## Username for admin panel
+##TURN_ADMIN_USER=admin
+#
+## Password for admin panel
+##TURN_ADMIN_SECRET=changeme
+#
+## HTTP(s) port for acess to admin panel
+##TURN_ADMIN_PORT=8443
+#
+## Type of TURN(s)/STUN. Can be turn or turns.
+##TURN_TYPE=turns
+#
+## Annonce FQDN or IP address of turn server
+##TURN_HOST=8.8.8.8
+#
+## TLS/TCP/UDP turn port for connection
+##TURN_PORT=5349
+#
+## Transport for stun/turn connection. Can be tcp or udp.
+##TURN_TRANSPORT=tcp
+#
+## RTP start port for turn/turns connections
+##TURN_RTP_MIN=10000
+#
+## RTP end port for turn/turns connections
+##TURN_RTP_MAX=11000
+#
--- a/env.example
+++ b/env.example
@@ -294,43 +294,47 @@ JIBRI_LOGS_DIR=/config/logs
 
 # Redirects HTTP traffic to HTTPS. Only works with the standard HTTPS port (443).
 #ENABLE_HTTP_REDIRECT=1
-#
-## Use TURN for P2P connections
-##TURN_ENABLE_P2P=0
-#
-## Use TURN for JVB (bridge mode) connections
-##TURN_ENABLE_JVB=0
-#
-## Realm to be used for the users with long-term credentials mechanism or with TURN REST API
-##TURN_REALM=realm
-#
-## Secret for connect to TURN server
-##TURN_SECRET=keepthissecret
-#
-## Username for admin panel
-##TURN_ADMIN_USER=admin
-#
-## Password for admin panel
-##TURN_ADMIN_SECRET=changeme
-#
-## HTTP(s) port for acess to admin panel
-##TURN_ADMIN_PORT=8443
-#
-## Type of TURN(s)/STUN. Can be turn or turns.
-##TURN_TYPE=turns
-#
-## Annonce FQDN or IP address of turn server
-##TURN_HOST=8.8.8.8
-#
-## TLS/TCP/UDP turn port for connection
-##TURN_PORT=5349
-#
-## Transport for stun/turn connection. Can be tcp or udp.
-##TURN_TRANSPORT=tcp
-#
-## RTP start port for turn/turns connections
-##TURN_RTP_MIN=10000
-#
-## RTP end port for turn/turns connections
-##TURN_RTP_MAX=11000
-#
+
+# Use TURN for P2P and JVB (bridge mode) connections
+#TURN_ENABLE=0
+
+# Realm to be used for the users with long-term credentials mechanism or with TURN REST API
+#TURN_REALM=realm
+
+# Secret for connect to TURN server
+#TURN_SECRET=keepthissecret
+
+# Type of TURN(s)/STUN. Can be turn or turns.
+#TURN_TYPE=turns
+
+# Annonce FQDN/IP address of the turn server via XMPP server (XEP-0215).
+# If empty or not set, variable DOCKER_HOST_ADDRESS will be used by default.
+#TURN_HOST=turn.example.com
+
+# Public IP address for an instance of turn server.
+# If empty or not set, will be detected and set dynamically when a container start.
+#TURN_PUBIC_IP=144.144.144.144
+
+# TLS/TCP/UDP turn port for connection
+#TURN_PORT=5349
+
+# Transport for stun/turn connection. Can be tcp or udp.
+#TURN_TRANSPORT=tcp
+
+# RTP start port for turn/turns connections
+#TURN_RTP_MIN=10000
+
+# RTP end port for turn/turns connections
+#TURN_RTP_MAX=11000
+
+# Enable admin web access. If enabled, please set variables below.
+#TURN_ADMIN_ENABLE=0
+
+# Username for admin panel
+#TURN_ADMIN_USER=admin
+
+# Password for admin panel
+#TURN_ADMIN_SECRET=changeme
+
+# HTTP(s) port for acess to admin panel
+#TURN_ADMIN_PORT=8443
--- a/env.example
+++ b/env.example
@@ -319,10 +319,10 @@ JIBRI_LOGS_DIR=/config/logs
 #TURN_TRANSPORT=tcp
 
 # RTP start port for turn/turns connections
-#TURN_RTP_MIN=10000
+#TURN_RTP_MIN=16000
 
 # RTP end port for turn/turns connections
-#TURN_RTP_MAX=11000
+#TURN_RTP_MAX=17000
 
 # Enable admin web access. If enabled, please set variables below.
 #TURN_ADMIN_ENABLE=0
--- a/env.example
+++ b/env.example
@@ -304,9 +304,6 @@ JIBRI_LOGS_DIR=/config/logs
 # Secret for connect to TURN server
 #TURN_SECRET=keepthissecret
 
-# Type of TURN(s)/STUN. Can be turn or turns.
-#TURN_TYPE=turns
-
 # Annonce FQDN/IP address of the turn server via XMPP server (XEP-0215).
 # If empty or not set, variable DOCKER_HOST_ADDRESS will be used by default.
 #TURN_HOST=turn.example.com
--- a/env.example
+++ b/env.example
@@ -310,7 +310,7 @@ JIBRI_LOGS_DIR=/config/logs
 
 # Public IP address for an instance of turn server.
 # If empty or not set, will be detected and set dynamically when a container start.
-#TURN_PUBIC_IP=144.144.144.144
+#TURN_PUBLIC_IP=144.144.144.144
 
 # TLS/TCP/UDP turn port for connection
 #TURN_PORT=5349
--- a/prosody/Dockerfile
+++ b/prosody/Dockerfile
@@ -39,6 +39,10 @@ RUN \
     && rm -rf /tmp/usr /var/cache/apt
 
 RUN \
+    curl -4so /prosody-plugins/mod_turncredentials.lua \
+    https://raw.githubusercontent.com/netaskd/mod_turncredentials/master/mod_turncredentials.lua
+
+RUN \
     sed -i s/hook/hook_global/g /prosody-plugins/mod_auth_token.lua \
     && patch -d /usr/lib/prosody/modules/muc -p0 < /prosody-plugins/muc_owner_allow_kick.patch
 
--- a/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
+++ b/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
@@ -5,7 +5,7 @@ http_default_host = "{{ .Env.XMPP_DOMAIN }}"
 {{ if .Env.TURN_ENABLE | default "0" | toBool }}
 turncredentials_secret = "{{ .Env.TURN_SECRET | default "keepthissecret" }}";
 turncredentials = {
-  { type = "{{ .Env.TURN_TYPE | default "turns" }}",
+  { type = "turns",
     host = "{{ .Env.TURN_HOST | default .Env.DOCKER_HOST_ADDRESS }}",
     port = {{ .Env.TURN_PORT | default "3478" }},
     transport = "{{ .Env.TURN_TRANSPORT | default "tcp" }}"
--- a/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
+++ b/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
@@ -2,11 +2,11 @@ admins = { "{{ .Env.JICOFO_AUTH_USER }}@{{ .Env.XMPP_AUTH_DOMAIN }}" }
 plugin_paths = { "/prosody-plugins/", "/prosody-plugins-custom" }
 http_default_host = "{{ .Env.XMPP_DOMAIN }}"
 
-{{ if or (.Env.TURN_ENABLE_P2P | default "0" | toBool) (.Env.TURN_ENABLE_JVB | default "0" | toBool) }}
+{{ if .Env.TURN_ENABLE | default "0" | toBool }}
 turncredentials_secret = "{{ .Env.TURN_SECRET | default "keepthissecret" }}";
 turncredentials = {
   { type = "{{ .Env.TURN_TYPE | default "turns" }}",
-    host = "{{ .Env.TURN_HOST | default "8.8.8.8" }}",
+    host = "{{ .Env.TURN_HOST | default .Env.DOCKER_HOST_ADDRESS }}",
     port = {{ .Env.TURN_PORT | default "3478" }},
     transport = "{{ .Env.TURN_TRANSPORT | default "tcp" }}"
   }
--- a/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
+++ b/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
@@ -2,6 +2,17 @@ admins = { "{{ .Env.JICOFO_AUTH_USER }}@{{ .Env.XMPP_AUTH_DOMAIN }}" }
 plugin_paths = { "/prosody-plugins/", "/prosody-plugins-custom" }
 http_default_host = "{{ .Env.XMPP_DOMAIN }}"
 
+{{ if or (.Env.TURN_ENABLE_P2P | default "0" | toBool) (.Env.TURN_ENABLE_JVB | default "0" | toBool) }}
+turncredentials_secret = "{{ .Env.TURN_SECRET | default "keepthissecret" }}";
+turncredentials = {
+  { type = "{{ .Env.TURN_TYPE | default "turns" }}",
+    host = "{{ .Env.TURN_HOST | default "8.8.8.8" }}",
+    port = {{ .Env.TURN_PORT | default "3478" }},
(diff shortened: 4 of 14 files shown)


prosody/Dockerfile
```diff
@@ -38,6 +38,10 @@ RUN \
     && apt-cleanup \
     && rm -rf /tmp/usr /var/cache/apt
 
+RUN \
+    curl -4so /prosody-plugins/mod_turncredentials.lua \
+    https://raw.githubusercontent.com/netaskd/mod_turncredentials/master/mod_turncredentials.lua
```
saghul: Why do we need your forked version?

turn.yml
```diff
@@ -0,0 +1,27 @@
+version: '3'
+
+services:
+    # coturn TURN server project
+    turn:
+        image: jitsi/turn
```
saghul: let's call it coturn, same way we name each image after the project it uses, for the most part.

turn/rootfs/defaults/docker-entrypoint.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/ash
+# make certs if not exist
+if [[ ! -f /etc/ssl/cert.crt || ! -f /etc/ssl/cert.key  ]]; then
```
saghul: These certs should be in /config

turn/rootfs/defaults/docker-entrypoint.sh
```diff
@@ -0,0 +1,38 @@
...
+--verbose \
+--prod \
+--no-tlsv1 \
+--no-tlsv1_1 \
+--log-file=stdout \
+--listening-port=${TURN_PORT:-5349} \
+--tls-listening-port=${TURN_PORT:-5349} \
+--alt-listening-port=${TURN_PORT:-5349} \
+--alt-tls-listening-port=${TURN_PORT:-5349} \
+--cert=/etc/ssl/cert.crt \
```
saghul: ditto

turn/rootfs/defaults/docker-entrypoint.sh
```diff
@@ -0,0 +1,38 @@
...
+--pkey=/etc/ssl/cert.key \
+--min-port=${TURN_RTP_MIN:-10000} \
+--max-port=${TURN_RTP_MAX:-11000} \
+--no-stun \
+--use-auth-secret \
+--static-auth-secret=${TURN_SECRET:-keepthissecret} \
+--no-multicast-peers \
+--realm=${TURN_REALM:-realm} \
+--external-ip=$(curl -4k https://icanhazip.com 2>/dev/null) \
```
saghul: This should probably be DOCKER_HOST_ADDRESS. On a related note, we should probably rename it...

turn/rootfs/defaults/docker-entrypoint.sh
```diff
@@ -0,0 +1,38 @@
...
+--relay-ip=$(hostname -i) \
+--listening-ip=$(hostname -i) \
+--web-admin \
```
saghul: Why enable the web admin?

turn/rootfs/defaults/docker-entrypoint.sh
```diff
@@ -0,0 +1,38 @@
...
+--pkey=/etc/ssl/cert.key \
+--min-port=${TURN_RTP_MIN:-10000} \
+--max-port=${TURN_RTP_MAX:-11000} \
+--no-stun \
+--use-auth-secret \
+--static-auth-secret=${TURN_SECRET:-keepthissecret} \
+--no-multicast-peers \
+--realm=${TURN_REALM:-realm} \
+--external-ip=$(curl -4k https://icanhazip.com 2>/dev/null) \
+--relay-ip=$(hostname -i) \
```
saghul: don't defaults Just Work (TM) ?

turn/rootfs/defaults/docker-entrypoint.sh
```diff
@@ -0,0 +1,38 @@
...
+--listening-ip=$(hostname -i) \
+--web-admin \
+--web-admin-ip=$(hostname -i) \
+--web-admin-port=${TURN_ADMIN_PORT:-8443} \
+--no-cli \
+--cli-password=${TURN_ADMIN_SECRET:-changeme}
```
saghul: why do we need to specify a password, if `--no-cli` is specified?

web/rootfs/defaults/config.js
```diff
@@ -227,7 +227,7 @@ var config = {
     // minParticipants: 2,
 
     // Use XEP-0215 to fetch STUN and TURN servers.
-    // useStunTurn: true,
+    // useStunTurn: true, //jvb
```
saghul: In the near future we are going to get rid of this cfg and get it directly from the package, so modifying it is not a great idea. I think we should just use a single setting for now, which would apply to both.

@damencho Do you have an idea on how to better handle this?

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -2,6 +2,17 @@ admins = { "{{ .Env.JICOFO_AUTH_USER }}@{{ .Env.XMPP_AUTH_DOMAIN }}" }
 plugin_paths = { "/prosody-plugins/", "/prosody-plugins-custom" }
 http_default_host = "{{ .Env.XMPP_DOMAIN }}"
 
+{{ if or (.Env.TURN_ENABLE_P2P | default "0" | toBool) (.Env.TURN_ENABLE_JVB | default "0" | toBool) }}
+turncredentials_secret = "{{ .Env.TURN_SECRET | default "keepthissecret" }}";
+turncredentials = {
+  { type = "{{ .Env.TURN_TYPE | default "turns" }}",
+    host = "{{ .Env.TURN_HOST | default "8.8.8.8" }}",
```
saghul: 8.8.8.8 is not a valid default, is it? Then this should fail, if the variable is empty.

saghul (changes requested):
Thanks a lot for getting this effort started! ❤️ Left some comments.

prosody/Dockerfile
netaskd: Original doesn't support prosody 0.11.X
Changes can be found here:  https://github.com/netaskd/mod_turncredentials/commit/76fabc32159580fac34ab627a016bcb9aa3c0697
Also, @damencho said (https://github.com/jitsi/docker-jitsi-meet/issues/22#issuecomment-564626682) he has plan to push PR his version to jitsi-meet repo. I think we can wait and insert "official" plugin when it will done.

turn/rootfs/defaults/docker-entrypoint.sh
netaskd: Just it makes log coturn cleaner. Otherwise warning will show. Is not affect functionality.

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
netaskd: yep, I'll change it to DOCKER_HOST_ADDRESS

turn.yml
netaskd: IMO, actually, I would prefer to stay it "turn" because it fully describes the functionality of the service. Coturn is a "coturn TURN server project", so I don't see any reasons to rename it. WDYS

turn/rootfs/defaults/docker-entrypoint.sh
netaskd: ack

turn/rootfs/defaults/docker-entrypoint.sh
netaskd: ack

turn/rootfs/defaults/docker-entrypoint.sh
netaskd: I think it needs more complicated then just `DOCKER_HOST_ADDRESS`.

turn/rootfs/defaults/docker-entrypoint.sh
netaskd: need check, usually the best way is to strictly described interfaces what do you want to use.

turn/rootfs/defaults/docker-entrypoint.sh
netaskd: it conveniently for manage turn connections.

prosody/Dockerfile
saghul: Got it, thanks!

turn/rootfs/defaults/docker-entrypoint.sh
saghul: Ok.

turn.yml
saghul: Ok.

turn/rootfs/defaults/docker-entrypoint.sh
saghul: Maybe we can call it TURN_PUBLIC_IP ? It could default to that, if not set.

saghul: But also: doesn't coturn detect its own  public IP if not set?

turn/rootfs/defaults/docker-entrypoint.sh
saghul: Manage in what way? I don't think we have ever used it in production servers.  I'd  say it should be an opt-in feature.

turn/rootfs/defaults/docker-entrypoint.sh
saghul: Then maybe let's set it to "NotReallyUs3d" or something, instead of reusing the admin secret.

netaskd: ack

web/rootfs/defaults/config.js
damencho: Hum, probably I will enable that by default in jitsi-meet installation in my PR of adding coturn there by default.

turn/rootfs/defaults/docker-entrypoint.sh
damencho: I think I had also set those in my test scenario, I will also test this once I got to this.

turn/rootfs/defaults/docker-entrypoint.sh
netaskd: the doc says it doesn't: https://github.com/coturn/coturn/blob/540ef5fd6e82faad0159ab24e2df92b4701cb62f/README.turnserver#L364

saghul: 👍

turn/rootfs/defaults/docker-entrypoint.sh
netaskd: I've just tested, so `--relay-ip` doesn't need actually.

turn/rootfs/defaults/docker-entrypoint.sh
netaskd: fixed

turn/rootfs/defaults/docker-entrypoint.sh
netaskd: fixed

turn/rootfs/defaults/docker-entrypoint.sh
netaskd: fixed

turn/rootfs/defaults/docker-entrypoint.sh
netaskd: fixed

web/rootfs/defaults/config.js
netaskd: now TURN_ENABLE flag change both options

README.md
```diff
@@ -344,6 +348,32 @@ Variable | Description | Example
 
 For setting `GOOGLE_APPLICATION_CREDENTIALS` please read https://cloud.google.com/text-to-speech/docs/quickstart-protocol section "Before you begin" from 1 to 5 paragraph.
 
+### TURN(S) server
+For enable turn server for P2P and JVB connections, please add to the variable `GLOBAL_MODULES` string `turncredentials` and set variables below
```
saghul: I think it would be nice if we did this automatically.

README.md
```diff
@@ -344,6 +348,32 @@ Variable | Description | Example
...
+
+Variable | Description | Default value
+--- | --- | ---
+`TURN_ENABLE` | Use TURN for P2P and JVB (bridge mode) connections | 0
+`TURN_REALM` | Realm to be used for the users with long-term credentials mechanism or with TURN REST API | realm
+`TURN_SECRET` | Secret for connect to TURN server | keepthissecret
+`TURN_TYPE` | Type of TURN(s) (turn/turns) | turns
```
saghul: Since we create TLS certificates always, we could hardcode this to `turns`

turn.yml
```diff
@@ -0,0 +1,29 @@
+version: '3'
+
+services:
```
saghul: This service needs the volume for /config

turn/rootfs/defaults/docker-entrypoint.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/ash
+# create config dir if not exists
+[ ! -d /config ] && mkdir /config
```
saghul: No need IIRC, docker does this because /config is a volume.

turn/Dockerfile
```diff
@@ -0,0 +1,11 @@
+ARG VERSION
+FROM instrumentisto/coturn:${VERSION:-latest}
+
+RUN apk add --no-cache openssl
+
+ADD ./rootfs/defaults/docker-entrypoint.sh /docker-entrypoint.sh
+
+ENTRYPOINT ["/docker-entrypoint.sh"]
```
saghul: Define /config as a volume

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -2,6 +2,17 @@ admins = { "{{ .Env.JICOFO_AUTH_USER }}@{{ .Env.XMPP_AUTH_DOMAIN }}" }
 plugin_paths = { "/prosody-plugins/", "/prosody-plugins-custom" }
 http_default_host = "{{ .Env.XMPP_DOMAIN }}"
 
+{{ if .Env.TURN_ENABLE | default "0" | toBool }}
+turncredentials_secret = "{{ .Env.TURN_SECRET | default "keepthissecret" }}";
+turncredentials = {
+  { type = "{{ .Env.TURN_TYPE | default "turns" }}",
+    host = "{{ .Env.TURN_HOST | default .Env.DOCKER_HOST_ADDRESS }}",
```
saghul: Shouldn't the default be TURN_PUBLIC_IP ?

web/rootfs/etc/cont-init.d/10-config
```diff
@@ -92,6 +92,13 @@ if [[ ! -f /config/config.js ]]; then
              -e "s#// transcribingEnabled:.*#transcribingEnabled: true,#" \
              /config/config.js
     fi
+
+    if [[ $TURN_ENABLE -eq 1 || "$TURN_ENABLE" == "true" ]]; then
```
saghul: We only check for 1 for the other ENABLE type settings...

saghul (changes requested):
Did another review round!

web/rootfs/etc/cont-init.d/10-config
netaskd: not for all. I don't see nothing bad to check `1` and `true`.

README.md
netaskd: yep, good idea

README.md
netaskd: ok

turn.yml
netaskd: ask

turn/rootfs/defaults/docker-entrypoint.sh
netaskd: ask

turn/Dockerfile
netaskd: ack

web/rootfs/etc/cont-init.d/10-config
saghul: Conincidentally it's only on the ones you added ;-) Ok then.

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
netaskd: prosody doesn't know about TURN_PUBLIC_IP in case TURN_PUBLIC_IP is not set from env file (got dynamically)

web/rootfs/etc/cont-init.d/10-config
netaskd: Yep ;)
I look at it from the point of view variables in configmap of k8s. So, `true` is more comfortable to use there.

saghul: Yeah, makes sense!

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
saghul: We can detect it in the init file and re-export it, perhaps?

netaskd: I'll try to explain my view on it. `TURN_HOST` is an announced FQDN for reach cluster of turn. It has use inside prosody for provide XEP-0215.
`TURN_PUBLIC_IP` is a exactly external IP address of turn instance. 
When we use `TURN_HOST`="turn.example.com", conditionally, `A` record can be `turn.example.com TURN_PUBLIC_IP1, TURN_PUBLIC_IP2, TURN_PUBLIC_IP3` and client (that got `TURN_HOST` via XEP-0215) goes to this scope.

netaskd: On the other hand, if it's just a single setup. TURN_HOST can be DOCKER_HOST_ADDRESS (if not set) and TURN_PUBLIC_IP is dynamically set.

netaskd: I'm not completely sure of the latest comment.. I guess, we should remove default set for "TURN_HOST" and make it force needed, because if TURN_PUBLIC_IP and DOCKER_HOST_ADDRESS is different, setup will broken. WDYT?

saghul: Not sure I follow you here. I do understand that you could set a hostname. But if you don't, the IP that should be there is the publicc IP of the TURN server, right?

On hindsight DOCKER_HOST_ADDRESS was not a great name. I'll probably rename it to JVB_PUBLIC_IP in the future, which is a better name.

So, to me, it makes sense that the default is the TURN public IP, and not anything else.

prosody/Dockerfile
sapkra:
```dockerfile
ADD https://raw.githubusercontent.com/netaskd/mod_turncredentials/master/mod_turncredentials.lua /prosody-plugins/mod_turncredentials.lua
```
Can you not just use this?

turn/Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+EXPOSE 5349 8443 10000:11000/udp
+
```
sapkra: It's recommended to run containers without root. Can you please add a line like I did it in my OpenShift PR and test it, otherwise I have to change it in my PR. (FYI OpenShift will launch containers with random UIDs and all users are always member of the root group)
Would be great to have new image already ready to run as an unprivileged user.

```dockerfile
USER 1001
```

prosody/Dockerfile
damencho: I'm very close to preparing my PR for adding turn support by default in jitsi-meet and the module will be part of it and will be installed with jitsi-meet once jitsi-meet-prosody package is installed.

damencho: https://github.com/jitsi/jitsi-meet/pull/4959

turn/rootfs/defaults/docker-entrypoint.sh
```diff
@@ -0,0 +1,42 @@
...
+--log-file=stdout \
+--listening-port=${TURN_PORT:-5349} \
+--tls-listening-port=${TURN_PORT:-5349} \
+--alt-listening-port=${TURN_PORT:-5349} \
+--alt-tls-listening-port=${TURN_PORT:-5349} \
+--cert=/config/cert.crt \
+--pkey=/config/cert.key \
+--min-port=${TURN_RTP_MIN:-10000} \
+--max-port=${TURN_RTP_MAX:-11000} \
+--no-stun \
```
mjattiot: Why stun is disabled ?

env.example
```diff
@@ -294,3 +294,44 @@ JIBRI_LOGS_DIR=/config/logs
...
+
+# Public IP address for an instance of turn server.
+# If empty or not set, will be detected and set dynamically when a container start.
+#TURN_PUBIC_IP=144.144.144.144
+
+# TLS/TCP/UDP turn port for connection
+#TURN_PORT=5349
+
+# Transport for stun/turn connection. Can be tcp or udp.
+#TURN_TRANSPORT=tcp
```
mjattiot: I am not sure we should be make it configurable. Coturn can both listen on udp, tcp, tcp/tls and dtls on the same port. We should let all these listeners opened so clients could try connecting using UDP and if a firewall block, fallback to TCP.

prosody/Dockerfile
pierreozoux: Yup, here is the official file:

https://raw.githubusercontent.com/jitsi/jitsi-meet/master/resources/prosody-plugins/mod_turncredentials.lua

env.example
```diff
@@ -294,3 +294,44 @@ JIBRI_LOGS_DIR=/config/logs
...
+# Secret for connect to TURN server
+#TURN_SECRET=keepthissecret
+
+# Annonce FQDN/IP address of the turn server via XMPP server (XEP-0215).
+# If empty or not set, variable DOCKER_HOST_ADDRESS will be used by default.
+#TURN_HOST=turn.example.com
+
+# Public IP address for an instance of turn server.
+# If empty or not set, will be detected and set dynamically when a container start.
+#TURN_PUBIC_IP=144.144.144.144
```
Craeckie:
```suggestion
#TURN_PUBLIC_IP=144.144.144.144
```

Craeckie (changes requested):

prosody/Dockerfile
```diff
@@ -38,6 +38,10 @@ RUN \
     && apt-cleanup \
     && rm -rf /tmp/usr /var/cache/apt
 
+RUN \
+    curl -4so /prosody-plugins/mod_turncredentials.lua \
```
sliard: you need to add : apt-get install curl

env.example
```diff
@@ -294,3 +294,44 @@ JIBRI_LOGS_DIR=/config/logs
...
+
+# TLS/TCP/UDP turn port for connection
+#TURN_PORT=5349
+
+# Transport for stun/turn connection. Can be tcp or udp.
+#TURN_TRANSPORT=tcp
+
+# RTP start port for turn/turns connections
+#TURN_RTP_MIN=10000
```
mtyurt, line 322: JVB default port is 10000 as well, wouldn't it conflict?

jkroepke:
Can I help here? I have free time...

sliard:
For information I finish a version that integrate a turn server into web container.
Why insert coturn inside web container and not create a new one ? It's for use the same SSH certificate.
https://github.com/sliard/docker-jitsi-meet/tree/add-ext-turn

It's not fair to create a new MR because it's base on this one at the begining.
I have one last problem, it works only with network_mode host into docker compose.

netaskd:
> Can I help here? I have free time...

@jkroepke it's really crazy sugession, but I will. My fork has A LOT of improvments for docker-jitsi-meet community, but I really do not have time for create PR and provide each of them. May be we somehow can realise it together?

turn/Dockerfile
netaskd: Could you please make PR around it, when this PR will merged?

env.example
netaskd: Feel free to make PR, when this PR will be merged.

env.example
netaskd: changed to 16k-17k

prosody/Dockerfile
netaskd: fixed in 5753213

env.example
netaskd: fixed in cb8552a

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
netaskd: I suggest to use only `TURN_HOST` for spreading hostname via XEP-0215. In this case we are suppose that:
a) the turn instance only one, and we exactly know how to reach it from outside.
b) the turn instance is multiple, but behind proxy. so, proxy knows how to reach each of them.

The `TURN_HOST` is mandatory variable, otherwise do not enable turn functions.

WDYT?

ccoenen:
One thing that's currently missing is documentation: which ports one would need to open additionally. This should be added here:

https://github.com/jitsi/docker-jitsi-meet#external-ports

netaskd:
> One thing that's currently missing is documentation: which ports one would need to open additionally. This should be added here:
> 
> https://github.com/jitsi/docker-jitsi-meet#external-ports

added in 14d717c

turn/Dockerfile
sapkra: That's fine for me because my PR is also not merged yet. I have to rebase it anyway if this PR becomes merged first which I think will be the case.

holtgrewe:
Hi, any news on this?

mihaiplesa:
Having this in will be great.

holtgrewe:
ping?

README.md
```diff
@@ -47,6 +47,9 @@ If you want to use jibri too, first configure host as described in JItsi BRoadca
 and then run Docker Compose as follows: ``docker-compose -f docker-compose.yml -f jibri.yml up -d``
 or to use jigasi too: ``docker-compose -f docker-compose.yml -f jigasi.yml -f jibri.yml up -d``
 
+If you want to enable TURN server, configure it and run Docker Compose as
```
luixxiul, line 50: This change needs to be applied to https://github.com/jitsi/handbook/blob/master/docs/devops-guide/docker.md (cf: #601)

holtgrewe: Is there anything I can do to support his PR towards being merged? I'd really like to vacate Zoom with all of its problems for Jitsi Meet...

netaskd:
As I see we have only one unresolved conversation around it. So, if @saghul have a couple of minutes to take a look this, I beleive we will fastly resolve and merge it.

holtgrewe:
@saghul could you have a look at this?

*Only you can save mankind from Zoom meetings* :)

dsantos21:
Hi! Although I'm not really versed on all needed to make this work on docker, I could test it in my on premises installation. Would it help? We are using docker, but setting up TURN without docker. Yes... Our JMS doesn't currently have a TURN server... :(

citosid:
Thank you so much for this work! I look forward to see this! Jitsi community really rocks!

prosody/Dockerfile
goacid: You could also just 
ADD ADD https://hg.prosody.im/prosody-modules/raw-file/tip/mod_turncredentials/mod_turncredentials.lua /prosody_plugins
And avoid adding curl in base image

prosody/Dockerfile
goacid: Official : 
`ADD https://hg.prosody.im/prosody-modules/raw-file/tip/mod_turncredentials/mod_turncredentials.lua /prosody-plugins/mod_turncredentials.lua`

holtgrewe:
Hi, is there anything that could be done here to move this forward?

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -2,6 +2,17 @@ admins = { "{{ .Env.JICOFO_AUTH_USER }}@{{ .Env.XMPP_AUTH_DOMAIN }}" }
...
+turncredentials_secret = "{{ .Env.TURN_SECRET | default "keepthissecret" }}";
+turncredentials = {
+  { type = "{{ .Env.TURN_TYPE | default "turns" }}",
+    host = "{{ .Env.TURN_HOST | default "8.8.8.8" }}",
+    port = {{ .Env.TURN_PORT | default "3478" }},
+    transport = "{{ .Env.TURN_TRANSPORT | default "tcp" }}"
+  }
+}
+{{ end }}
+
```
goacid: The credential must be set in the prosody.cfg.lua files and not in the jitsi-meet.cfg.lua.
In jitsi-meet.cfg.lua you have to add the "turncredantials" in the module_enabled section of the meet.jitsi Virtualhost.

nestorconde:
I've worked on this from the latest docker-jitsi-meet adding all the changes of the PR except for two: 

1. It does not contain the changes to the README of the original PR since that page content has been moved somewhere else  
2. It changes the name of the turn docker container (from jitsi/turn which wasn't found to instrumentisto/coturn). 

I attach the patch which is mostly the same of the PR except for these two changes. I'd also be able to provide the terraform script that created the machines I'm testing in (in Digital Ocean). [turn.diff](https://github.com/jitsi/docker-jitsi-meet/files/4842147/turn.txt)

**Problems**

Jitsi works but I'm having some problems to test whether the turn server is working. 

I can't block the port 10000/udp with ufw. I've tried to block ports with ufw but it doesn't seem to have any effect [perhaps due to this?](https://stackoverflow.com/questions/49549834/ufw-firewall-is-not-working-on-ubuntu-in-digitalocean/49563279#49563279).

The other thing I've tried is to checking chrome://webrtc-internals (since I haven't been able to block 10000 this is not as useful as it could be). When I test with several participants I see in the some of the tabs "iceServers: [stun:meet-jit-si-turnrelay.jitsi.net:443]", but this is the same I see in a server where I have not applied the patch

@netaskd Any way I could go on from here in order to provide help to advance this pull request?

First option I can think of is to try other ways to block port 10000. Do other iptables management utilities have the same problem as ufw with respect to docker?, any advice on how to do block 10000/udp without stumbling on this problem?

holtgrewe:
> First option I can think of is to try other ways to block port 10000. Do other iptables management utilities have the same problem as ufw with respect to docker?, any advice on how to do block 10000/udp without stumbling on this problem?

Hi, does this help?

https://stackoverflow.com/questions/49549834/ufw-firewall-is-not-working-on-ubuntu-in-digitalocean/49563279#49563279

baccenfutter:
Docker generally doesn't play well out-of-the-box with firewall configuration tools such as UFW, shorewall, etc. Such tools generate IPTABLES rules based on some "easier" configuration format, while Docker itself leverages IPTABLES to "magically" configure NAT for the containers. If both tools are not aware of each other, they will overwrite or shadow each-other's IPTABLE rules.

Both UFW and Shorwall have builtin support for Docker in their latest versions which is disabled by default. Enabling this feature will make UFW "docker-aware" and preserve Dockers IPTABLE rules when run.

However, it is still rather complicated to block incoming connections to Docker via IPTABLE rules, due to the fact that NAT is processed before the FILTER rules, meaning that packets to Docker-configured ports are redirected into the corresponding container before the packets run through the FILTER chain of the host - therefor never hitting the rule.

See: https://2.bp.blogspot.com/-skGmyfTzXGI/U84PI9EbjkI/AAAAAAAAD6Q/kjjfG17_yo8/s1600/Linux-Iptables-firewall-schema.png

The correct way to prevent access into a Docker container is to explicitly state `127.0.0.1` in the `--port` argument when running a container, such as

```
docker run --port 127.0.0.1:10000:10000 ...
```

This will only allow access to port 10000 when the connection is coming from the Docker host itself and will deny access when coming from elsewhere.

parruc:
I also need this. How can I help the development? What's still open?

nestorconde:
@baccenfutter What I did was modify the docker-compose.yml at the root of the project by changing
```
jvb
   ...
    ports:
-      '${JVB_PORT}:${JVB_PORT}/udp'
+      '127.0.0.1:${JVB_PORT}:${JVB_PORT}/udp'
```

Is that equivalent to what you suggested? (I don't know enough about docker and docker-compose to evaluate whether it is). 

In any case after doing that I was having problems with the audio/video from the videoconference which only worked momentarily and then stopped working. Perhaps there's some unexpected side effect to that way of restricting the port?, if that's not the case then my patch is not working when 10000/udp is blocked.

@holtgrewe I did not try that because it seemed to me that I'd be changing more things than strictly needed (it would affect potentially other things). In order to evaluate whether the patch is working I need a test that isolates the changes so I wanted to only block 10000/udp and leave everything else the same.

@netaskd @saghul There's a lot of interest in this PR and some people would like to help but we don't have the information on how to do it. As far as I understand there's just [this](https://github.com/jitsi/docker-jitsi-meet/pull/163#pullrequestreview-331539367) change pending review but the turncredentials code in netaskd branch have changed since the time the review was requested. Could you provide some guidance if helping is possible? If it isn't it would be useful to know that too ;)

turn/rootfs/defaults/docker-entrypoint.sh
```diff
@@ -0,0 +1,42 @@
...
+--use-auth-secret \
+--static-auth-secret=${TURN_SECRET:-keepthissecret} \
+--no-multicast-peers \
+--realm=${TURN_REALM:-realm} \
+--listening-ip=$(hostname -i) \
+--external-ip=${TURN_PUBLIC_IP} \
```
vkruoso, line 39: As per https://hub.docker.com/r/instrumentisto/coturn they recommend using the ```detect-external-ip``` tool to automatically get the external IP. I did not check how that works, but I'm not sure as well to add a dependency on ```icanhazip.com``` to get the TURN server up a good thing.

netaskd:
Actually, the script `derect-external-ip` do the same ;)
See https://github.com/instrumentisto/coturn-docker-image/blob/master/rootfs/usr/local/bin/detect-external-ip.sh

turn/rootfs/defaults/docker-entrypoint.sh
netaskd: Actually, the script derect-external-ip do the same ;)
See https://github.com/instrumentisto/coturn-docker-image/blob/master/rootfs/usr/local/bin/detect-external-ip.sh

vkruoso: Yeah, should have checked it before. Awesome. I've been testing this and works great. My use case is on kubernetes and Im having some issues with networking, but that is another story.

pandel:
Read through all of this but didn't get it... will this also support adding an external TURN server without starting an additional coturn docker container which isn't needed in that case?

hermanbanken:
What is the status of this PR, what discussion is still pending?

I'm also working on integrating Jitsi, and we really need to support TURN for corporate VPNs. The setup we are using is to split traffic with Envoy & also terminate TLS there. That way both Jitsi's NGINX/web container & CoTURN don't need to do TLS offloading.

example envoy.yaml configuration:
<details>
# docker run -v $PWD:/etc/ssl/private -v $PWD:/data envoyproxy/envoy:v1.14.4 /usr/local/bin/envoy -c /data/envoy.yaml --mode validate
#
admin:
  access_log_path: /tmp/admin_access.log
  profile_path: /tmp/envoy.prof
  address:
    socket_address: { address: 0.0.0.0, port_value: 9901 }

static_resources:
  listeners:
  # Main listener
  - address:
      socket_address: { address: 0.0.0.0, port_value: 4430 }
    # TLS Inspector parses the SNI & ALPN settings
    listener_filters:
    - name: "envoy.filters.listener.tls_inspector"
      typed_config: {}
    # One filter chain per SNI & ALPN
    filter_chains:
      # shard-1 HTTP
      - filter_chain_match: # shard-1 http
          server_names: ["shard-1.eu.example.org"]
          application_protocols: ["http/1.1", "h2"]
        transport_socket: &tmpl_transport_socket
          name: envoy.transport_sockets.tls
          typed_config:
            "@type": type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.DownstreamTlsContext
            common_tls_context:
              tls_certificates:
              - certificate_chain: { filename: "/etc/ssl/private/tls.crt" }
                private_key: { filename: "/etc/ssl/private/tls.key" }
        filters: &tmpl_http_filters
          - name: envoy.filters.network.http_connection_manager
            typed_config:
              "@type": type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
              access_log:
                - name: envoy.access_loggers.file
                  typed_config:
                    "@type": type.googleapis.com/envoy.extensions.access_loggers.file.v3.FileAccessLog
                    path: /tmp/http_access.log
              stat_prefix: ingress_http
              route_config:
                virtual_hosts:
                - name: default
                  domains: "*"
                  routes:
                  - match: { prefix: "/" }
                    route: { cluster: service_shard1_nginx }
              http_filters:
                - name: envoy.filters.http.router
      # shard-1 TURN
      - filter_chain_match:
          server_names: ["shard-1.eu.example.org"]
        transport_socket: *tmpl_transport_socket
        filters:
          - name: envoy.filters.network.tcp_proxy
            typed_config:
              "@type": type.googleapis.com/envoy.extensions.filters.network.tcp_proxy.v3.TcpProxy
              stat_prefix: tcp
              cluster: service_shard1_turn
      # fallback HTTP (redirects to Application Level router)
      - filter_chain_match:
          server_names: ["*.example.org"]
        transport_socket: *tmpl_transport_socket
        filters:
          - name: envoy.filters.network.http_connection_manager
            typed_config:
              "@type": type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
              access_log:
                - name: envoy.access_loggers.file
                  typed_config:
                    "@type": type.googleapis.com/envoy.extensions.access_loggers.file.v3.FileAccessLog
                    path: /tmp/http_access.log
              stat_prefix: ingress_http
              route_config:
                virtual_hosts:
                - name: default
                  domains: "*"
                  routes:
                  - match: { prefix: "/" }
                    route: { cluster: service_fallback }
              http_filters:
                - name: envoy.filters.http.router
      # fallback for health check
      - filter_chain_match:
        transport_socket: *tmpl_transport_socket
        filters:
          - name: envoy.filters.network.http_connection_manager
            typed_config:
              "@type": type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
              stat_prefix: direct_http
              route_config:
                virtual_hosts:
                - name: default
                  domains: "*"
                  routes:
                  - match: { prefix: "/" }
                    direct_response:
                      status: 200
                      body:
                        inline_string: "Fallback"
              http_filters:
                - name: envoy.filters.http.router
  clusters:
  - name: service_fallback
    connect_timeout: 0.25s
    type: strict_dns
    lb_policy: round_robin
    load_assignment:
      cluster_name: service_fallback
      endpoints:
      - lb_endpoints:
        - endpoint:
            address:
              socket_address:
                address: nginx-internal
                port_value: 80
  - name: service_shard1_nginx
    connect_timeout: 0.25s
    type: strict_dns
    lb_policy: round_robin
    load_assignment:
      cluster_name: service_shard1_nginx
      endpoints:
      - lb_endpoints:
        - endpoint:
            address:
              socket_address:
                address: nginx.shard-1.svc.cluster.local
                port_value: 80
  - name: service_shard1_turn
    connect_timeout: 0.25s
    type: strict_dns
    lb_policy: round_robin
    load_assignment:
      cluster_name: service_shard1_turn
      endpoints:
      - lb_endpoints:
        - endpoint:
            address:
              socket_address:
                address: turn.shard-1.svc.cluster.local
                port_value: 5349

</details> 

I'm also working on an Envoy xDS API endpoint that serves this configuration dynamically. That would ease adding shards drastically in our Kubernetes setup.

nestorconde:
I've documented in [the community](https://community.jitsi.org/t/docker-jitsi-meet-and-firewalls/80155) my tests with a simple firewall and:

1.  The setup described in this PR 
2.  Multiplexing (for turn to work when its ports are blocked by the firewall) based on [the docs](https://jitsi.github.io/handbook/docs/devops-guide/turn). My exact setup for multiplexing can be seen [here](https://github.com/jitsi/docker-jitsi-meet/pull/667#issuecomment-702335852)

As far as I can see under with this setup for the videoconferences to work it's enough if the firewall allows 80, 443 (both outbound and inbound) and 10000 outbound.

I know this PR does not include the multiplexing bit but I think this contributes to the conversation and the multiplexing part is an interesting direction for turn in docker-jitsi-meet to pusue

turn/rootfs/defaults/docker-entrypoint.sh
```diff
@@ -0,0 +1,42 @@
...
+# set coturn web-admin access
+if [[ "${TURN_ADMIN_ENABLE}" == "1" || "${TURN_ADMIN_ENABLE}" == "true" ]]; then
+  turnadmin -A -u ${TURN_ADMIN_USER:-admin} -p ${TURN_ADMIN_SECRET:-changeme}
+  export TURN_ADMIN_OPTIONS="--web-admin --web-admin-ip=$(hostname -i) --web-admin-port=${TURN_ADMIN_PORT:-8443}"
+fi
+
+# run coturn server with API auth method enabled.
+turnserver -n ${TURN_ADMIN_OPTIONS} \
+--verbose \
+--prod \
```
hamletmun, line 21: https://github.com/coturn/coturn/commit/520e172b22c7b35fd8e90ed113f2464281b15ad6
> --no-software-attribute	Production mode: hide the software version.

almereyda: That's a good spot!

turn/Dockerfile
```diff
@@ -0,0 +1,13 @@
+ARG VERSION
+FROM instrumentisto/coturn:${VERSION:-latest}
```
almereyda, line 2: This has been deprecated: https://github.com/instrumentisto/coturn-docker-image/commit/0a4a7f0963118d91110c51d49abd88a28c2ba7d9

It is suggested to switch to the upstream image [`coturn/coturn`](https://hub.docker.com/r/coturn/coturn) ([🔗](https://github.com/coturn/coturn/tree/master/docker/coturn)).

turn/rootfs/defaults/docker-entrypoint.sh
almereyda: This has been adopted for the upstream `coturn` image, too.

https://github.com/coturn/coturn/blob/master/docker/coturn/rootfs/usr/local/bin/detect-external-ip.sh

See previous comment https://github.com/jitsi/docker-jitsi-meet/pull/163/files#r802157850

VidarHUN:
Hi Folks,

This PR is still alive or somebody working on it? What kind of job is left to do?

Also, I don't understand this PR. As I can see the main goal of this PR is to make it happen to configure an external TURN server for Jitsi and you have created a new image which will be a coTURN server for testing or if somebody wants to use it. Am I right? If it's true then I would like to help you to finish this, because from the commits and because of my lack of knowledge with the configuration I can't see another way to configure Jitsi containers to use TURN servers.

If we can configure Jitsi to use TURN server, then there is a way to somehow force the relay through TURN with p2p connections? Can I just put a zero or false value into an env variable and force it?

SimJoSt:
@VidarHUN @netaskd @vorburger checking in to get some clarity on this.
We struggled with turn integration the last time we deployed it. A native solution would be great.

satonotdead:
Should be great to have. Thanks to all the contributors :)

satonotdead:
I'm agree, a native solution should be great for everyone :)

ccoenen:
Dear github-actions bot, please kindly FUCK OFF.

Either this work is generally unwelcome or outdated, then some human should just say so and close this.

Or it is valid, then some human should look at it and evaluate what is still missing to get this merged.

Stalebots, however, don't add anything here. Which is why they are generally frowned upon. This is not some obscure unreproducible error where the original reporter does not reply for months. This is a (IMO) nearly finished pull request with little activity. Because it's nearly finished, it's sitting there. That does not mean that the contribution should be flushed down the toilet because someone once thought putting a robot in charge would be a neat idea.

saghul:
Why do you think you are entitled to tell others what to do in their projects?

PR #166, opened 2019-12-17
added bind_adress because of reverse proxy

BlackWolfDEsign:
only changes for reverse proxy setups

added:
HTTP_BIND and HTTPS_BIND

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,8 +5,8 @@ services:
     web:
         image: jitsi/web
         ports:
-            - '${HTTP_PORT}:80'
-            - '${HTTPS_PORT}:443'
+            - '${HTTP_BIND}:${HTTP_PORT}:80'
+            - '${HTTPS_BIND}:${HTTPS_PORT}:443'
         volumes:
             - ${CONFIG}/web:/config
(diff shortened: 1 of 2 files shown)


saghul:
Can you please explain why this is needed?

BlackWolfDEsign:
container binded to 0.0.0.0 = container reachable from the internet: container --> www
binded to 127.0.0.1 = only reachable from server: container --> server(reverse proxy) --> www

saghul:
I still don’t get it, sorry. By default we bind on any interface and then expose the port with Docker. Are you using host mode or something?

BlackWolfDEsign:
No, I just pulled jitsi-meet and started it. (without hostmode or something)

Without a binding address, jitsi-meet is bound to every interface and can therefore be reached via [IP]:PORT, **as you wrote**. 

But I want to bind to a fixed interface, and that was the easiest way (for me). Then i can map the [IP]:PORT to jitsi.domain.de and jitsi is only reachable via this address.

saghul:
Ah I think I get it now. I’d like some changes though:

- A single option used in both cases
- Please call it HTTP_INTERFACE

saghul:
Did you intend to close this?

Issue #13, opened 2018-09-20
Record and stream


Volont84:
Hi!
It is a great solution easy to use and work just fine.
But i have an issue with recording. As far i understand it is not possible to record or stream your conference. I tried to create jibri docker image but no luck for now. Maybe there is some another way to record?

dimm0:
Do I understand it right that we can send the video and audio streams from jitsi to ffmpeg, which will generate a resulting file/stream for us?

https://unix.stackexchange.com/questions/233832/merge-two-video-clips-into-one-placing-them-next-to-each-other

aaronkvanmeerten:
That's incorrect.  The jitsi videobridge does not generate video streams.  It simply forwards UDP packets.

PR #169, opened 2020-01-03
Simple k8s example

dimm0:
Added a minimal single-pod example for deploying jitsi to kubernetes cluster


sapkra:
@saghul recommended to have all examples in an `examples` directory.

```
examples
-- kubernetes
-- openshift
-- others
```

saghul:
Thanks for this! I'll try to give it a test soon. @aaronkvanmeerten @jmacelroy can you PTAL too?

HubbeKing:
Since the containers use a bunch of the same environment variables (`XMPP_DOMAIN`, `XMPP_AUTH_DOMAIN`, `XMPP_INTERNAL_MUC_DOMAIN`, `JVB_BREWERY_MUC`,`TZ`), it would be better to define them in a ConfigMap and pull them into the containers using envFrom and configMapRef. This way, you can have them all in one place, just like how the existing docker-compose.yml file pulls them all from a .env file.

Likewise, since you're already defining `JVB_AUTH_PASSWORD` and `JICOFO_AUTH_PASSWORD` in Secrets, you should also define `JVB_AUTH_USER` and `JICOFO_AUTH_USER` in those Secrets, to make it easier to adjust these values without making mistakes.

jmacelroy:
> Since the containers use a bunch of the same environment variables (`XMPP_DOMAIN`, `XMPP_AUTH_DOMAIN`, `XMPP_INTERNAL_MUC_DOMAIN`, `JVB_BREWERY_MUC`,`TZ`), it would be better to define them in a ConfigMap and pull them into the containers using envFrom and configMapRef. This way, you can have them all in one place, just like how the existing docker-compose.yml file pulls them all from a .env file.
> 
> Likewise, since you're already defining `JVB_AUTH_PASSWORD` and `JICOFO_AUTH_PASSWORD` in Secrets, you should also define `JVB_AUTH_USER` and `JICOFO_AUTH_USER` in those Secrets, to make it easier to adjust these values without making mistakes.

I've typically kept secrets in secret store and not added associated non-secret config. Not every user can view the secrets and it seems like a trade-off. I'm fine leaving it for however the committer would like the example to be. In practice I imagine the example will always be slightly modified for the specific cluster.

Also, I think I could argue either way on the configmap issue in this case. As an example I think having the separate copy/paste makes it more clear what service uses what. In practice I'd probably templatize and use a single helm chart with the values.yaml defining the same for all env vars or separate configmaps per service.

Great thoughts and suggestions but I'm ok with implementer deciding in this case.

examples/kubernetes/README.md
```diff
@@ -0,0 +1,26 @@
+# Install guide for kubernetes
+
+This guide will deploy jitsi in the most simple way: as several containers in a single pod. This is enough to start in case your hardware is enough. If you need to scale components to severa instance, you'll have to modify it to use several services and pods.
```
jmacelroy: Small typo, severa->several instances.

Bellusterra:
@dimm0 Thank you.

dimm0:
Ok great!
Yeah, I tried to keep it as simple and minimal as possible, so that everyone can add to it

Bellusterra:
Your instructions look very clear and easy to follow, which is great for a novice like myself :). Speaking of adding to it, I know the jibri docker was recently released and was thinking that would make a great addition... it might take me a month/year but would be happy to try and figure that out if it doesn’t exist already?

dimm0:
It's actually on my todo list, and I started adding it to my install, but was stopped by a bunch of requirements like adding sound driver modules to the node.. Can it be simplified?
(we can talk in https://rocket.nautilus.optiputer.net if you want, to not pollute the thread here)

dimm0:
I looked at jibri, and it doesn't seem very useful now... Only one session can be recorded, lots of pre-setup... I think I'll wait for something better to be implemented.

aaronkvanmeerten:
Jibri was originally meant to be used on individual instances, so the docker image is only useful as a proof of concept I fear.  Part of the problem is the sound driver modules which need a device per jibri, as well as the CPU/memory required to run the chrome as a meeting participant.  I suggest running the jibris outside docker, and exposing the prosody client port to allow them to be discoverable, if you're looking to run more than a single recording.

dimm0:
Why can't we simply dump the sound and video stream from jitsi core into a file? (I know nothing about how it's made inside unfortunately...)

aaronkvanmeerten:
Absolutely a reasonable question.  The main answer is that the jitsi videobridge does no transcoding, and no formatting/rendering.  It is simply an SFU (selective forwarding unit).  So whatever is going to do the recording would need to receive all relevant packets, decrypt and render them. Also, since we get streams from each participant, we'd need to choose how to composite these into a screen.  The best tool for all of these jobs together is a browser (since we already do this in the jitsi-meet client), so that's how we got here.

dimm0:
Seems like ffmpeg can do this...
https://unix.stackexchange.com/questions/233832/merge-two-video-clips-into-one-placing-them-next-to-each-other

Just need to run it with proper params depending on the number of participants, and pack in a container :)

dimm0:
Moving discussion here https://github.com/jitsi/docker-jitsi-meet/issues/13

AliMD:
@dimm0 I'm sure you dont test at all (Specially more than 2 users) 😞 
Its not work at all and I found many bugs in enviroment!

AliMD:
@dimm0 I'm sure you dont test it at all (Specially more than 2 users) 😞
Its not work at all and I found many bugs in enviroment!

@jmacelroy Please dont merge before testing, I took more than 20 hours to fix it! And steel finding bugs in production!

saghul:
@AliMD The examples (anything other than the docker-compose file) are 100% community maintained and are self-serving. If you found problems, please send a PR!

dimm0:
Works fine for me with 10 people

PR #170, opened 2020-01-04
fix letsencrypt renewal

kegato:
When mounting the `/etc/letsencrypt` directory into a volume the letsencrypt renewal does not succeed because the init script only copies the renewal script into `/etc/cron.monthly` if there is no certificate yet.

Also it's better to use cron.daily since `certbot renew` does only renew the certificate if it's close to expire.

### DIFF
--- a/web/rootfs/etc/cont-init.d/10-config
+++ b/web/rootfs/etc/cont-init.d/10-config
@@ -18,7 +18,16 @@ if [[ $DISABLE_HTTPS -ne 1 ]]; then
                 -d $LETSENCRYPT_DOMAIN \
                 --agree-tos \
                 --email $LETSENCRYPT_EMAIL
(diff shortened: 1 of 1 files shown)


saghul:
Thanks for the PR! Won't this exhaust the weekly renewalls quota? Or is certboot smart enough not to do the actual renew every day?

kegato:
Calling `certbot renew` checks if a certificate is close to expiry and only renews it if that's the case.
[By default](https://github.com/certbot/certbot/blob/e048da1e389ede7a52bd518ab4ebf9a0b18bfafc/certbot/certbot/_internal/constants.py#L141) all certificates which expire in less 30 days are renewed.

So there won't be any quota problems 😄 .

saghul:
Great news then!

Issue #174, opened 2020-01-23
Jibri in docker container doesn't work


volo-zyko:
This is the same issue as jitsi/jibri#231. Chrome inside the container tries to open "https://meet.jitsi" while should be directed to "https://external.fqdn".

Proposed in the above issue solution probably should work. Related code is https://github.com/jitsi/jibri/blob/d566436459b5c1f1772af550d5d4c3357ceb2cee/src/main/kotlin/org/jitsi/jibri/api/xmpp/XmppApi.kt#L276 but it's not completely clear how to preserve the old behavior. In some setups it will need to use "external.fqdn" and in some setups - `xmpp_domain` (which proably will be the same as "external.fqdn").

saghul:
Are you using the latest version of the images? When I tested it before merging it did work. Can you share your `.env` file?

volo-zyko:
I have the latest images.

Here is [`.env`](https://gist.github.com/volo-zyko/18e458e0d552c8fcf41d5adc69c75167) file.

saghul:
I don't see any obvious mistake. I'll try to repro and let you know.

netaskd:
@volo-zyko 
> Chrome inside the container tries to open "https://meet.jitsi" while should be directed to "https://external.fqdn".

For get properly direct to external.fqdn you have to set `XMPP_DOMAIN=external.fqdn` and so on, because  
https://github.com/jitsi/docker-jitsi-meet/blob/6234a18dc69e81ff01cb52a5dc89534ee4f93273/env.example#L281

original in config.json: https://github.com/jitsi/jibri/blob/d566436459b5c1f1772af550d5d4c3357ceb2cee/resources/debian-package/etc/jitsi/jibri/config.json#L49

volo-zyko:
After some more thinking about this problem I changed all `jitsi.meet` strings in the above `.env` file to `external.fqdn`. Also, I changed `LDAP_FILTER` to `(|(mail=%U@external.fqdn)(mail=%U@external.fqdn2))` (thanks to the suggestion in  jitsi/docker-jitsi-meet#173). Now, I can login with the full email from domain `external.fqdn` and users in `external.fqdn2` can only login with `@external.fqdn2` omitted.

Still Jibri cannot record video. I can see in logs that chrome tries to open `https://external.fqdn` which is wrong and should be `https://video-chat.external.fqdn`.

volo-zyko:
@netaskd 
[...]

PR #176, opened 2020-01-24
docker-compose: add the right IP for the XMPP_DOMAIN in /etc/hosts

volo-zyko:
This must fix the issue I described [here](https://github.com/jitsi/docker-jitsi-meet/issues/174#issuecomment-578316073). In short, if HTTPS is disabled then it must be enabled on the external IP. And since Chrome uses only HTTPS then we must direct it to that external IP.

I must warn that I didn't test this code (I only manually added a line to `/etc/hosts` inside container) but it's rather trivial.

### DIFF
--- a/jibri.yml
+++ b/jibri.yml
@@ -12,6 +12,8 @@ services:
         devices:
             - /dev/snd:/dev/snd
         environment:
+            - DISABLE_HTTPS
+            - DOCKER_HOST_ADDRESS
             - XMPP_AUTH_DOMAIN
             - XMPP_INTERNAL_MUC_DOMAIN
             - XMPP_RECORDER_DOMAIN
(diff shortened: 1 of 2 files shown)


saghul:
I think this is a bug in Jibri and it should be fixed there. It should have some “public URL base” setting, assuming the domain is reachable directly is kinda wrong.

I’m not a fan of the proposed solution because I have plans to rename DOCKER_HOST_ADDRESS and give it a more JVB related name, since that’s what it was supposed to be.

volo-zyko:
Ok, I don't mind. Then close this PR.

volo-zyko:
On the other hand you confirm that this is a bug and currently this PR is the simplest workaround for it. So, even if the PR is rejected it might be worth to mention it in the readme file.

saghul:
Let’s keep this open so we can at least remember. Also, there might be some other way to do this workaround.

saghul:
Fixed in Jibri, latest unstable has this fixed as of https://github.com/jitsi/docker-jitsi-meet/pull/1046

PR #178, opened 2020-01-27
Adds support of reservation REST API base URL .env input (fixes #175)

### LINKED ISSUE #175: Jicofo's reservation.rest.BASE_URL cannot be specified as ENV
Currently it's not possible to set the Base URL of Jicofo's reservation system from the `.env` file.
Instead `org.jitsi.impl.reservation.rest.BASE_URL=http://reservation.example.com` must be added manually to `.jitsi-meet-cfg/jicofo/sip-communicator.properties`, which get's resetted sometimes.
There is definitely some workarounds, but it'd be most elegant, if this feature was implemented, so that adding the URL to the `.env` file would to the trick.

kheir95:
Enables setting the Base URL of Jicofo's reservation system from the .env file.
Instead of adding `org.jitsi.impl.reservation.rest.BASE_URL=http://reservation.example.com` manually to .jitsi-meet-cfg/jicofo/sip-communicator.properties, which get's resetted sometimes.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -123,6 +123,7 @@ services:
             - JIBRI_BREWERY_MUC
             - JIBRI_PENDING_TIMEOUT
             - TZ
+            - JICOFO_RESERVATION_REST_BASE_URL
         depends_on:
             - prosody
         networks:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -118,12 +118,12 @@ services:
             - JICOFO_COMPONENT_SECRET
             - JICOFO_AUTH_USER
             - JICOFO_AUTH_PASSWORD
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -123,6 +123,7 @@ services:
             - JIBRI_BREWERY_MUC
             - JIBRI_PENDING_TIMEOUT
             - TZ
+            - JICOFO_RESERVATION_REST_BASE_URL
```
saghul: small nit: do you mind moving this up, together with the other Jicofo variables? Thank you!

kheir95: Sure and done! :)

timonegk:
Do you mind adding the documentation for this variable to the table in the README?

PR #181, opened 2020-01-31
enable websockets

aaronkvanmeerten:
web: enable websockets in config.js
doc: document new ENABLE_WEBSOCKETS variable

### DIFF
--- a/README.md
+++ b/README.md
@@ -388,6 +388,7 @@ Variable | Description | Default value
 `JIGASI_TRANSCRIBER_ADVERTISE_URL` | Jigasi post to the chat an url with transcription file | true
 `DISABLE_HTTPS` | Disable HTTPS, this can be useful if TLS connections are going to be handled outside of this setup | 1
 `ENABLE_HTTP_REDIRECT` | Redirects HTTP traffic to HTTPS | 1
+`ENABLE_WEBSOCKETS` | Enabled xmpp traffic over websockets | 1
 `LOG_LEVEL` | Controls which logs are output from prosody and associated modules | info
 
 ### Running behind NAT or on a LAN environment
--- a/env.example
+++ b/env.example
@@ -148,6 +148,9 @@ XMPP_SERVER=xmpp.meet.jitsi
 # Internal XMPP server URL
 XMPP_BOSH_URL_BASE=http://xmpp.meet.jitsi:5280
 
+# Flag to enable websockets
+XMPP_WEBSOCKETS=true
+
 # Internal XMPP domain for authenticated services.
 XMPP_AUTH_DOMAIN=auth.meet.jitsi
 
--- a/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
+++ b/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
@@ -8,6 +8,7 @@ http_default_host = "{{ .Env.XMPP_DOMAIN }}"
 {{ $JWT_ALLOW_EMPTY := .Env.JWT_ALLOW_EMPTY | default "0" | toBool }}
 {{ $JWT_AUTH_TYPE := .Env.JWT_AUTH_TYPE | default "token" }}
 {{ $JWT_TOKEN_AUTH_MODULE := .Env.JWT_TOKEN_AUTH_MODULE | default "token_verification" }}
+{{ $ENABLE_WEBSOCKETS := .Env.ENABLE_WEBSOCKETS | default "1" | toBool }}
 
 {{ if and $ENABLE_AUTH (eq $AUTH_TYPE "jwt") .Env.JWT_ACCEPTED_ISSUERS }}
 asap_accepted_issuers = { "{{ join "\",\"" (splitList "," .Env.JWT_ACCEPTED_ISSUERS) }}" }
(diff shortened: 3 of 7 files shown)


prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -8,6 +8,7 @@ http_default_host = "{{ .Env.XMPP_DOMAIN }}"
 {{ $JWT_ALLOW_EMPTY := .Env.JWT_ALLOW_EMPTY | default "0" | toBool }}
 {{ $JWT_AUTH_TYPE := .Env.JWT_AUTH_TYPE | default "token" }}
 {{ $JWT_TOKEN_AUTH_MODULE := .Env.JWT_TOKEN_AUTH_MODULE | default "token_verification" }}
+{{ $ENABLE_WEBSOCKETS := .Env.ENABLE_WEBSOCKETS | default "1" | toBool }}
```
saghul: I think we should default it to 0, keeping the current defaults.

env.example
```diff
@@ -148,6 +148,9 @@ XMPP_SERVER=xmpp.meet.jitsi
 # Internal XMPP server URL
 XMPP_BOSH_URL_BASE=http://xmpp.meet.jitsi:5280
 
+# Flag to enable websockets
+XMPP_WEBSOCKETS=true
```
saghul: ENABLE_WEBSOCKET?

web/rootfs/defaults/config.js
```diff
@@ -40,6 +40,9 @@ var config = {
     // BOSH URL. FIXME: use XEP-0156 to discover it.
     bosh: '//jitsi-meet.example.com/http-bind',
 
+    // Websocket URL
+    websocket: '//jitsi-meet.example.com/xmpp-websocket',
```
saghul: Did you copy this from the current config.js on meet?

web/rootfs/etc/cont-init.d/10-config
```diff
@@ -66,6 +66,7 @@ if [[ ! -f /config/config.js ]]; then
     sed -i \
         -e "s#jitsi-meet.example.com#$XMPP_DOMAIN#g" \
         -e "s#bosh:.*#bosh: '/http-bind',#" \
+        -e "s#websocket:.*#websocket: 'wss://'+window.location.host+'/xmpp-websocket',#" \
```
saghul: This won’t work on mobile. A relative URL should do though.

saghul:
Please re-target the PR to dev.

gchallen:
Thanks so much for submitting this patch! I found it very helpful in getting XMPP websockets to work.

However, I needed to make a few additional changes. My full solution for both XMPP and Colibri websockets is summarized [in my forum post](https://community.jitsi.org/t/solved-docker-compose-setup-for-local-development-using-websockets/23367/3). The parts that are relevant to this patch are the changes to `jitsi-meet.cfg.lua`, and the resulting requirements to modify `docker-compose.yml`.

Specifically: I had to set `consider_websocket_secure = true` and `cross_domain_websocket = true` at the top level of `jitsi-meet.cfg.lua`. These changes seem like they should definitely be included, since nginx is terminating wss and modifying the hostname.

I also configured `trusted_proxies` as described in [here](https://prosody.im/doc/http). I'm not 100% sure this is required, but adding it creates other problems, since you need to fix the IP address of the nginx proxy. This is doable, but then requires additional changes to`docker-compose.yml` as described [here](https://community.jitsi.org/t/solved-docker-compose-setup-for-local-development-using-websockets/23367/3).

Finally, there was an error in the `config.js` concerning a `websocket` option that seems to be no longer supported.

**Overall I really support finishing and merging this patch.** In my experience Jitsi works much (much) better with XMPP websockets available, and it would be great to enable this behavior out of the box without all of the pain and fiddling I encountered.

saghul:
Thanks for the feedback @gchallen! That's the goal here, but we're quite busy at the moment so this will have to wait a bit...

web/rootfs/defaults/config.js
sapkra: `serviceUrl` should be used instead which can be used for xmpp and websocket endpoints

```javascript
// websockets
serviceUrl: 'wss://server.com/xmpp-websocket'

// bosh
serviceUrl: '//server.com/xmpp-websocket'
```

jkroepke:
@saghul Can I help?

README.md
```diff
@@ -388,6 +388,7 @@ Variable | Description | Default value
 `JIGASI_TRANSCRIBER_ADVERTISE_URL` | Jigasi post to the chat an url with transcription file | true
 `DISABLE_HTTPS` | Disable HTTPS, this can be useful if TLS connections are going to be handled outside of this setup | 1
 `ENABLE_HTTP_REDIRECT` | Redirects HTTP traffic to HTTPS | 1
+`ENABLE_WEBSOCKETS` | Enabled xmpp traffic over websockets | 1
```
luixxiul, line 391: This change needs to be applied to https://github.com/jitsi/handbook/blob/master/docs/devops-guide/docker.md (cf: #601)

saghul:
Superseded by https://github.com/jitsi/docker-jitsi-meet/pull/502

PR #184, opened 2020-02-04
jicofo: set owner jicofo rights for /config directory

mjattiot:


saghul:
HI there! Thanks for sending in a patch! Is there a reson (other than the annoying warning) for this? Wouldn't we also need it for the jvb?

mjattiot:
Well I am not sure it's an annoying warning : the java process is not allowed to read the configuration so any changes won't be applied, will they ?
Anyway jvb had already been patched the same way by you I think .. :) 
https://github.com/jitsi/docker-jitsi-meet/commit/bb1d386ebcacbf50d11095a16166032f2992ee9a#diff-0f226375ce91ffa55099fe1bceae5d6a

saghul:
Oh, LoL, let's be consistent then! :-)

saghul:
Landed in 6eebabd, thanks! (I cherry-picked the commit to save you a rebase, since we develop against the dev branch).

As for the config, it's loaded fine, the fact that the directory is not writtable is no big deal, but squelching warnings is A Good Thing (TM) so thanks for noticing and fixing it!

mjattiot:
Cool, thank you for your explanation and your quick reaction !

PR #190, opened 2020-02-16
prosody: fix restart loop on rolling deployment

sapkra:


saghul:
Cheers!

saghul:
Hey @sapkra, I had to revert this, sorry. It breaks the default installation. Prosody fails to start because it cannot write to /tmp. I tried changing it to /var/run/ (which also makes more sense) but same thing, and I ran out of time for debugging.

sapkra:
Mhh right...I have an idea how to fix it because I   had to change a lot of permissions anyway.

My suggestion:
I will update and split my OpenShift PR soon and we will apply this change afterwards or within the new PR which includes all permission changes.

saghul:
Sounds good!

PR #197, opened 2020-03-13
add mkdir -p ~/.jitsi-meet-cfg/... to README.md

vorburger:
`podman-compose up -d` fails for me when I do not first create these directories


saghul:
While not needed for a plain Docker setup, these won't hurt. Thank you!

cognitive-arch:
This command is not working on windows. i want to setup jisti on docker linux based from windows but im not be able to create this directory cloud you please help me @saghul

saghul:
You can skip that line, it’s not necessary in all cases IIRC

PR #200, opened 2020-03-13
Clarify HTTP vs HTTPS in README (fixes #199)

### LINKED ISSUE #199: Remove http as it seems to be broken and only https actually works
I followed the README which currently suggests to _"Access the web UI at https://localhost:8443 (or http://localhost:8000 for HTTP)"_ and so when I got the self signed SSL cert warning on https naively used http as documented, but got:

On Chrome: _Cannot use microphone for an unknown reason. Cannot read property 'getUserMedia' of undefined_

On Firefox: _Cannot use microphone for an unknown reason. navigator.mediaDevices is undefined._

When I used https these errors went away and it seemed to work.

It looks like in recent modern browsers WebRTC only works on https? (I'm a WebRTC noob, just guessing.)

To avoid confusing for future users running into this, how about we remove http? I'll propose a PR.

vorburger:
@saghul re. #199

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,10 @@ follow these steps:
 * Create a ``.env`` file by copying and adjusting ``env.example``.
   * `cp env.example .env`
 * Run ``docker-compose up -d``.
-* Access the web UI at ``https://localhost:8443`` (or ``http://localhost:8000`` for HTTP, or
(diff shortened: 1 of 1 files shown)


saghul:
Thank you!

Issue #201, opened 2020-03-13
Containers don't work with Podman instead of Docker


vorburger:
I personally prefer using https://podman.io instead of Docker (because containers dont' run as root, on the host; although you can still be root inside the container..), and when I tried that with this project (FYI they have a `podman-compose` which, in my experience, is reasonably compatible with `docker-compose`) but I've noticed that the images of this project don't yet work with Podman instead of Docker. The web container for example failed with the error below (I hadn't even checked the others.)

The short-term workaround is, of course, to just use Docker instead of Podman for now, but I thought I'd at least just let you know about this by filing this issue here. 

Glancing over this project, I've noticed open PRs #192 and #126, it's possible they help with this.

@saghul more of an FYI

```
[s6-init] making user provided files available at /var/run/s6/etc...                                                                                                                                                 
[s6-init] ensuring user provided files have correct perms...                                                                                                                                                         
[fix-attrs.d] applying ownership & permissions fixes...                                                                                                                                                              
[fix-attrs.d] done.                                                                                                                                                                                                  
[cont-init.d] executing container initialization scripts...                                                                                                                                                          
[cont-init.d] 01-set-timezone: executing...                                                                                                                                                                          
[cont-init.d] 01-set-timezone: exited 0.                                                                                                                                                                             
[cont-init.d] 10-config: executing...                                                                                                                                                                                
mkdir: cannot create directory '/config/nginx': Permission denied                                                                                                                                                    
mkdir: cannot create directory '/config/keys': Permission denied                                                                                                                                                     
generating self-signed keys in /config/keys, you can replace these with your own keys if required                                                                                                                    
Generating a RSA private key                                                                                                                                                                                         
........................++++                                                                                                                                                                                         
....................................................................................................++++                                                                                                             
writing new private key to '/config/keys/cert.key'                                                                                                                                                                   
req: Can't open "/config/keys/cert.key" for writing, No such file or directory                                                                                                                                       
Can't open /config/nginx/dhparams.pem for writing, No such file or directory                                                                                                                                         
140389928026176:error:02001002:system library:fopen:No such file or directory:../crypto/bio/bss_file.c:74:fopen('/config/nginx/dhparams.pem','w')                                                                    
140389928026176:error:2006D080:BIO routines:BIO_new_file:no such file:../crypto/bio/bss_file.c:81:                                                                                                                   
cp: cannot create regular file '/config/nginx/nginx.conf': No such file or directory                                                                                                                                 
[...]

PR #204, opened 2020-03-13
use :Z (selinux) for volume mounts (re. #201)

vorburger:
This helps with, but doesn't entirely fix just yet,
running this project on Fedora using `podman up -d` instead of Docker.

This should not break regular `docker-compose` on systems without SELinux.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,9 +8,9 @@ services:
             - '${HTTP_PORT}:80'
             - '${HTTPS_PORT}:443'
         volumes:
-            - ${CONFIG}/web:/config
-            - ${CONFIG}/web/letsencrypt:/etc/letsencrypt
-            - ${CONFIG}/transcripts:/usr/share/jitsi-meet/transcripts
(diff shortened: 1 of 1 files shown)


vorburger:
@saghul this partially helps with #201, but doesn't entirely fix it yet.

You'll probably want to double check that it does not break your regular Docker (it should not).

jmhunter:
FWIW - I just tried using docker-compose to install jitsi (CentOS 7, selinux enabled) and it was not successful until I applied this patch.

Before the patch it was unable to create directories as below. After the patch, all works great!

```
[cont-init.d] 10-config: executing...
mkdir: cannot create directory '/config/nginx': Permission denied
mkdir: cannot create directory '/config/keys': Permission denied
```

vorburger:
@saghul given the independent verification" (above) of this one, do you want to merge it?

digital-codes:
I have the same setup (centos 7.7. with selinux). The :Z is required to make it work.

sapkra:
Closing because already applied in master.

Issue #209, opened 2020-03-16
gcp shell : read_bio_DHparams("/config/nginx/dhparams.pem") failed


jmikedupont2:
running in the google cloud shell web preview
docker logs docker-jitsi-meet_web_1 shows
nginx: [emerg] PEM_read_bio_DHparams("/config/nginx/dhparams.pem") failed (SSL: error:0906D06C:PEM routines:PEM_read_bio:no start line:Expecting: DH PARAMETERS)

https://github.com/nginx-proxy/nginx-proxy/issues/1226

saghul:
This runs on container startup: https://github.com/jitsi/docker-jitsi-meet/blob/master/web/rootfs/etc/cont-init.d/10-config#L43

Can you check the container logs? Has nginx fully started?

PR #211, opened 2020-03-16
workaround #209

jmikedupont2:
Today I setup docker jitsi on the google cloud shell web preview and got the website working but no the audio or video stream.
I got it running up to the point I could connect with authenticated login via google, but the audio was not working and the app did not work, presumably because they did not handle the login via google that redirects to the sso/oauth. That will be a job for another day, I did change the https port to 8888 because that was an option. 

I encountered one issue that the certificate for nginx did not work and the workaround was to generate it again and bounce the server,
ideally you would do that before starting, just after you copy the configuration.
https://github.com/jitsi/docker-jitsi-meet/issues/209

My workaround is documented in my updated version here. Hope this helps someone. I don't expect this pull request to be merged because of the port change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,9 @@ follow these steps:
 * Create a ``.env`` file by copying and adjusting ``env.example``, and create required `CONFIG` directories
   * `cp env.example .env`
   * `mkdir -p ~/.jitsi-meet-cfg/{web/letsencrypt,transcripts,prosody,jicofo,jvb}`
+  * `openssl dhparam -out ~/.jitsi-meet-cfg/web/nginx/dhparams.pem 2048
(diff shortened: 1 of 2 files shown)


saghul:
I think the size for DHparamss is higher by default so it may take up to a ocuple of minutes for it to be generated. Maybe that's your problem?

jmikedupont2:
Interesting. Letting the user generate it at least gives the feedbAck.
Otherwise noobs will need some instructions. Will try agaon and look for
it.

saghul:
No follow-up, closing.

jmikedupont2:
Its is ok. I need more time to look Into this

PR #222, opened 2020-03-18
Fixed URL that contains dash

Tyldar:
If you create an instance with a subdomain containing a dash, it will not be reachable. This commit fixes it

### DIFF
--- a/web/rootfs/defaults/meet.conf
+++ b/web/rootfs/defaults/meet.conf
@@ -6,7 +6,7 @@ root /usr/share/jitsi-meet;
 index index.html
 error_page 404 /static/404.html;
 
-location ~ ^/([a-zA-Z0-9=\?]+)$ {
+location ~ ^/([a-zA-Z0-9=\?\-]+)$ {
     rewrite ^/(.*)$ / break;
 }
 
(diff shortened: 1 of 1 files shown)


web/rootfs/defaults/meet.conf
```diff
@@ -6,7 +6,7 @@ root /usr/share/jitsi-meet;
 index index.html
 error_page 404 /static/404.html;
 
-location ~ ^/([a-zA-Z0-9=\?]+)$ {
+location ~ ^/([a-zA-Z0-9=\?\-]+)$ {
```
QuentinBrosse: Can you please add `_` too?

Tyldar: Your desires are orders m8

saghul:
@damencho WDYT?

damencho:
I think that we should be using what is currently in jitsi-meet. Isn't that covered already? 
https://github.com/jitsi/jitsi-meet/blob/master/doc/debian/jitsi-meet/jitsi-meet.example#L72

damencho:
This is the rule we use in production.

saghul:
Yes indeed, I’ll update to that.

saghul:
Fixed in https://github.com/jitsi/docker-jitsi-meet/pull/245

PR #227, opened 2020-03-18
Update images to buster

### LINKED ISSUE #219: Update to Debian "buster"
According to [Jitsi Call on 3/9](https://www.youtube.com/watch?v=UfYRt20um4U), https://meet.jit.si uses [epoll](https://prosody.im/doc/network_backend).

This has not yet been incorporated into `docker-jitsi-meet`.

There are some important epoll fixes in [Prosody 0.11.3](https://blog.prosody.im/prosody-0.11.3-released/)

Stretch-backports (which `docker-jitsi-meet` currently uses) only [provides 0.11.2](https://packages.debian.org/stretch-backports/prosody).

Buster is the [current stable Debian release](https://www.debian.org/releases/stable/).
Buster-backports [provides 0.11.4](https://packages.debian.org/buster-backports/prosody).

gauthierl:
This pull request update all images to Debian Buster.

I've compiled a Openssl 1.0 in the prosody image cause I didn't find any other way to get Luacrypto install correctly.

The images are not well tested but seems to work.  
At least the one I use : jicofo, jvb, web, prosody.

I didn't make any change for using epoll.

Fix #219 ?

### DIFF
--- a/base-java/Dockerfile
+++ b/base-java/Dockerfile
@@ -4,6 +4,6 @@ FROM ${JITSI_REPO}/base
 RUN	\
 	mkdir -p /usr/share/man/man1 && \
 	apt-dpkg-wrap apt-get update && \
-	apt-dpkg-wrap apt-get install -y openjdk-8-jre-headless && \
+	apt-dpkg-wrap apt-get install -y openjdk-11-jre-headless && \
 	apt-cleanup
 
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:stretch-slim
+FROM debian:buster-slim
 
 ARG JITSI_RELEASE=stable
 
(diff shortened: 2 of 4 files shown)


saghul:
Thanks for getting this started! Isn’t OpenSSL 1.0.x installable anymore? Is so we should migrate from using that package before migrating to buster I think.

@damencho WDYT?

gauthierl:
Yes,  libssl1.0-dev is not available anymore. I guess you should go away from luacrypto that seem to be deprecated for jwt. I will not able to do that.

saghul:
Yep we need to do that. No worries, thanks for your help so far!

saghul:
I had a chat with @bbaldino on Monday and we are not really readdy to recomment the usse of Java 11 in production. Jibri may not even work at all. We currently run 8.

So, we'll need to build some confidence here before we can go this route, and this is specially relevant since Buster doesn't provide Java 8, so we cannot do the change in batches.

So, I'm going to keep the issue open but close this, for the time being.

toby63:
Any news on this?

Could you open issues to track potential changes of the encryption library/method, necessary changes for a new Java version etc.?

Issue #226, opened 2020-03-18
Incorrect instruction to add user to prosody


digital-codes:
Instructions in readme need update:
user registration is disabled by default in prosody, so command
prosodyctl --config /config/prosody.cfg.lua register username meet.jitsi passsword
doens not work. Please change instruction to use command adduser instead of register, like
prosodyctl --config /config/prosody.cfg.lua adduser xyz@meet.jitsi
then provide password on command line.

daaaaan:
this also fixed my issue #216

saghul:
I have never been able to reproduce this. Have you made changes to the Prosody config in any way?

digital-codes:
Not that I'm aware of. I just enabled the internal authentication in the environment file. 
But maybe the issue some others brought up about "no response from register" is related to something else: the docu says "register" is a batch command, so you wouldn't expect a response unless it  fails (which it did in my  case)

saghul:
@digital-codes Can you please give the latest images a try? That's 4101-2. Prosody was updated there.

Issue #216, opened 2020-03-16
Secure Domain - Internal Auth - Not Authenticating


daaaaan:
Having followed the docs for deployment of secure domain with internal authentication. 

The only changes I've really made are as follows : 

```

HTTP_PORT=80
HTTPS_PORT=443
PUBLIC_URL="https://video.fqdn.live" << with corresponding lets encrypt enabled
DOCKER_HOST_ADDRESS=167.172*****
ENABLE_AUTH=1
ENABLE_GUESTS=1
AUTH_TYPE=internal

```

Then I've run 
```

docker-compose exec prosody /bin/bash
prosodyctl --config /config/prosody.cfg.lua register username meet.jitsi passsword

```

Then headed to video.fqdn.live/test, which was working without auth, and hit the "Password Required" dialog. 

I've tried 

```
username:password
username@meet.jitsi:password

```

So far no joy, console shows : 

```

CONNECTION FAILED: connection.passwordRequired

```

Prosody logs show : 

```

New BOSH session, assigned it sid '7e81b464-8d6a-424e-b0b2-8fbc0cbe114d'

```

But nothing further after failed logon. 

I've rebuilt the docker images and volumes a few times, wiped the config folder etc 

Any help appreciated. 🤷‍♂

daaaaan:
Had a look a bit further down the prosody logs and this is hiding up there : 

```

 mv: cannot stat '/config/data/*.crt': No such file or directory
prosody_1  | mv: cannot stat '/config/data/*.key': No such file or directory
prosody_1  | [cont-init.d] 10-config: exited 0.

```
Likely root cause?

saghul:
Did you change your .env configuration after starting contaainers? If so you must detroy your CONFIG directory and start again so it gets recreated.

daaaaan:
> Did you change your .env configuration after starting contaainers? If so you must detroy your CONFIG directory and start again so it gets recreated.

Yes, this has been done. Also with a docker-compose down --rmi="all" --volumes as well to get rid of any dangling volume files.

saghul:
Can you paste the generated prosody configuration?

daaaaan:
This is fixed with PR #228 

Its a documentation issue. Wrong Prosody command.

sukubo:
No of this fixes do it... still cant add user y prosodyctl... another cuestion, some can put an own valid signe certificate, not a let's encrypt?

amazzoccone:
[...]

PR #228, opened 2020-03-18
Update README.md

daaaaan:
Fixes incorrect documentation re : internal auth

Fixes : #226 #216

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,7 +288,7 @@ In order to do that, first execute a shell in the corresponding container:
 
 Once in the container, run the following command to create a user:
 
-``prosodyctl --config /config/prosody.cfg.lua register username meet.jitsi passsword``
+``prosodyctl --config /config/prosody.cfg.lua adduser user@meet.jitsi``
 
 The command then asks for a password interactively.
 


saghul:
That used to work... what’s the difference between register and adduser?

saghul:
Hum, the docs specify both work: https://prosody.im/doc/creating_accounts

Moreover we use register for creating internal accounts which are internally needed...

vidschofelix:
`prosodyctl --config /config/prosody.cfg.lua register username meet.jitsi`
also works

wojas:
Even if the `register` command actually works, this part of the README is incorrect for `register`:

> The command then asks for a password interactively.

It's also unclear that `username` and `password` in the example are placeholders, and that `meet.jitsi` is the default (or even hardcoded?) in the Docker Compose setup.

saghul:
I tested this on a fresh install yesterday evening and it works just fine. Any chance you've made changes to the prosody setup?

wojas:
The remark about "The command then asks for a password interactively." threw me off. With that extra line, it's not clear that `password` on the commandline is a placeholder, because why would you get prompted for a password if you just specified it?

The command also does not output anything, which makes it seem like it's not working, and does not appear in the list of commands in the help output.

When there are three PRs regarding this specific part of the README (this, #249 and #274), please accept that something is unclear about that part of the documentation.

saghul:
Agreed, I'll fix the docs right away.

PR #232, opened 2020-03-19
allow whitespaces in room names

fusselkater:
As the welcome page allows whitespaces in room names, nginx should handle whitespaces correctly. Added whitepsace in location regex in meet.conf

### DIFF
--- a/web/rootfs/defaults/meet.conf
+++ b/web/rootfs/defaults/meet.conf
@@ -6,7 +6,7 @@ root /usr/share/jitsi-meet;
 index index.html
 error_page 404 /static/404.html;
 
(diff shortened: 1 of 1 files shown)


saghul:
URLs must be percent encoded so that would be %20. No spaces.

fusselkater:
That is correct. The welcome page redirects to /test%20channel if I type in "test channel" but the url encoding seems to be decoded before nginx parses the location rules. So in the location regex it has to be a whitespace.

fusselkater:
Sorry, i forgot to describe the problem. The problem is, if i type "test channel" in the welcome page, I get a 404 Error as the location rule does not match.

saghul:
We are tracking the allowed roon names regex in other open issues, please take a look.

PR #235, opened 2020-03-19
A Helm chart for jitsi-meet

Zempashi:
Add a kubernetes helm chart for jitsi-meet

Featured:
- quite as par with the main docker-compose file
- a separate subchart for prosody, that's the easiest way to disable prosody completely and use an external prosody server
- multiple option to handle the public IP of the jvb (kubernetes could be quite complicated regarding network setup)

Non-featured:
- other components (such as sip gayteway, I don't think the range port could be easily achievable in kubernetes)
- full feature prosody chart, but interface is generic enough to be one day a dependency to an official prosody chart
- the release process to made it available via helm repository (https://helm.sh/docs/topics/chart_repository/#share-your-charts-with-others)

Tested:
 Tested options 2, 3, 4, not the 1 (https://github.com/Zempashi/docker-jitsi-meet/tree/k8s-helm/examples/helm)

### DIFF
--- a/examples/helm/README.md
+++ b/examples/helm/README.md
@@ -1,0 +1,120 @@
+# jitsi-meet
+
+[jitsi-meet](https://jitsi.org/jitsi-meet/) Secure, Simple and Scalable Video Conferences that you use as a standalone app or embed in your web application.
+
+## TL;DR;
+
+```console
+$ git clone https://github.com/jitsi/docker-jitsi-meet
+$ helm install ./docker-jitsi-meet/examples/helm
+```
+
+## Introduction
+
+This chart bootstraps a jitsi-meet deployment, like the official [one](https://meet.jit.si).
+
+## Different topology
+
+To be able to do video conferencing with other people, the jvb component should be reachable by all participants (eg: a public IP).
+Thus the default behaviour of advertised the internal IP of jvb, is not really suitable in many cases.
+Kubernetes offers multiple possibilities to work around the problem. Not all options are available depending on the Kubernetes cluster setup.
+The chart tries to make all options available without enforcing one.
+
+### Option 1: service of type `LoadBalancer`
+
+This requires a cloud setup that enables a Loadbalancer attachement.
+This could be enabled via values:
+
+```yaml
+jvb:
+  service:
+    type: LoadBalancer
+
+  # Depending on the cloud, publicIP cannot be know in advance, so deploy first, without the next option.
+  # Next: redeploy with the following option set to the public IP you retrieved from the API.
+  publicIP: 1.2.3.4
+```
+
+In this case you're not allowed to change the `jvb.replicaCount` to more than `1`, UDP packets will be routed to random `jvb`, which would not allow for a working video setup.
+
+### Option 2: NodePort and node with Public IP or external loadbalancer
+
+```yaml
+jvb:
+  service:
+    type: NodePort
+  # It may be required to change the default port to a value allowed by Kubernetes (30000-32768)
+  UDPPort: 30000
+  TCPPort: 30443
+
+  # Use public IP of one of your node, or the public IP of a loadbalancer in front of the nodes
+  publicIP: 1.2.3.4
+```
+
+In this case you're not allowed to change the `jvb.replicaCount` to more than `1`, UDP packets will be routed to random `jvb`, which would not allow for a working video setup.
+
+### Option 3: hostPort and node with Public IP
+
+Assuming that the node knows the PublicIP it holds, you can enable this setup:
+
+```yaml
+jvb:
+  useHostPort: true
+  # This option requires kubernetes >= 1.17
+  useNodeIP: true
+```
+
+In this case you can have more the one `jvb` but you're putting you cluster at risk by having it directly exposed on the Internet.
+
+### Option 4: Use ingress TCP/UDP forward capabilities
+
+In case of an ingress capable of doing tcp/udp forwarding (like nginx-ingress), it can be setup to forward the video streams.
+
+```yaml
+# Don't forget to configure the ingress properly (separate configuration)
+jvb:
+  # 1.2.3.4 being one of the IP of the ingress controller
+  publicIP: 1.2.3.4
+
+```
+
+Again in this case, only one jvb will work in this case.
+
+### Option 5: Bring your own setup
+
+There are multiple other possibilities combining the available parameters, depending of your cluster/network setup.
+
+
+
+## Configuration
+
+The following table lists the configurable parameters of the jisti-meet chart and their default values.
+
+Parameter | Description | Default
+--- | --- | ---
+`imagePullSecrets` | List of names of secrets resources containing private registry credentials | `[]`
+`enableAuth` | Enable authentication | `false`
+`enableGuests` | Enable guest access | `true`
+`jicofo.replicaCount` | Number of replica of the jicofo pods | `1`
+`jicofo.image.repository` | Name of the image to use for the jicofo pods | `jitsi/jicofo`
+`jicofo.extraEnvs` | Map containing additional environment variables for jicofo | '{}'
+`jicofo.livenessProbe` | Map that holds the liveness probe, you can add parameters such as timeout or retries following the Kubernetes spec | A livenessProbe map
+`jicofo.readinessProbe` | Map that holds the liveness probe, you can add parameters such as timeout or retries following the Kubernetes spec | A readinessProbe map
+`jicofo.xmpp.user` | Name of the XMPP user used by jicofo to authenticate | `focus`
+`jicofo.xmpp.password` | Password used by jicofo to authenticate on the XMPP service | 10 random chars
+`jicofo.xmpp.componentSecret` | Values of the secret used by jicofo for the xmpp-component | 10 random chars
+`jvb.service.enabled` | Boolean to enable os disable the jvb service creation | `false` if `jvb.useHostPort` is `true` otherwise `true`
+`jvb.service.type` | Type of the jvb service | `ClusterIP`
+`jvb.UDPPort` | UDP port used by jvb, also affects port of service, and hostPort | `10000`
+`jvb.TCPPort` | TCP port used by jvb, also affects port of service, and hostPort | `4443`
+`jvb.extraEnvs` | Map containing additional environment variables to jvb | '{}'
+`jvb.xmpp.user` | Name of the XMPP user used by jvb to authenticate | `jvb`
+`jvb.xmpp.password` | Password used by jvb to authenticate on the XMPP service | 10 random chars
+`jvb.livenessProbe` | Map that holds the liveness probe, you can add parameters such as timeout or retries following the Kubernetes spec | A livenessProbe map
+`jvb.readinessProbe` | Map that holds the liveness probe, you can add parameters such as timeout or retries following the Kubernetes spec | A readinessProbe map
+`web.httpsEnabled` | Boolean that enabled tls-termination on the web pods. Useful if you expose the UI via a `Loadbalancer` IP instead of an ingress | `false`
+`web.httpRedirect` | Boolean that enabled http-to-https redirection. Useful for ingress that don't support this feature (ex: GKE ingress) | `false`
+`web.extraEnvs` | Map containing additional environment variable to web pods | '{}'
+`web.livenessProbe` | Map that holds the liveness probe, you can add parameters such as timeout or retries following the Kubernetes spec | A livenessProbe map
+`web.readinessProbe` | Map that holds the liveness probe, you can add parameters such as timeout or retries following the Kubernetes spec | A readinessProbe map
+`tz` | System Time Zone | `Europe/Amsterdam`
--- a/examples/helm/templates/jicofo/deployment.yaml
+++ b/examples/helm/templates/jicofo/deployment.yaml
@@ -1,0 +1,66 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "jitsi-meet.jicofo.fullname" . }}
+  labels:
+    {{- include "jitsi-meet.jicofo.labels" . | nindent 4 }}
+spec:
+  replicas: {{ .Values.jicofo.replicaCount }}
+  selector:
+    matchLabels:
+      {{- include "jitsi-meet.jicofo.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      labels:
+        {{- include "jitsi-meet.jicofo.selectorLabels" . | nindent 8 }}
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/jicofo/configmap.yaml") . | sha256sum }}
+        checksum/secret: {{ include (print $.Template.BasePath "/jicofo/xmpp-secret.yaml") . | sha256sum }}
+    spec:
+    {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+    {{- end }}
+      serviceAccountName: {{ include "jitsi-meet.serviceAccountName" . }}
+      securityContext:
+        {{- toYaml .Values.jicofo.podSecurityContext | nindent 8 }}
+      containers:
+        - name: {{ .Chart.Name }}
+          securityContext:
+            {{- toYaml .Values.jicofo.securityContext | nindent 12 }}
+          image: "{{ .Values.jicofo.image.repository }}:{{ default .Chart.AppVersion .Values.jicofo.image.tag }}"
+          imagePullPolicy: {{ pluck "pullPolicy" .Values.jicofo.image .Values.image | first }}
+          envFrom:
+          - secretRef:
+              name: {{ include "call-nested" (list . "prosody" "prosody.fullname") }}-jicofo
+          - configMapRef:
+              name: {{ include "call-nested" (list . "prosody" "prosody.fullname") }}-common
+          - configMapRef:
+              name: {{ include "jitsi-meet.jicofo.fullname" . }}
+          ports:
+            - name: http
+              containerPort: 80
+              protocol: TCP
+          {{- with .Values.jicofo.livenessProbe }}
+          livenessProbe:
+          {{- toYaml . | nindent 12 }}
+          {{- end }}
+          {{- with .Values.jicofo.readinessProbe }}
+          readinessProbe:
+          {{- toYaml . | nindent 12 }}
+          {{- end }}
+          resources:
+            {{- toYaml .Values.jicofo.resources | nindent 12 }}
+
+      {{- with .Values.jicofo.nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+    {{- with .Values.jicofo.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+    {{- end }}
+    {{- with .Values.jicofo.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+    {{- end }}
--- a/examples/helm/templates/jicofo/deployment.yaml
+++ b/examples/helm/templates/jicofo/deployment.yaml
@@ -13,9 +13,15 @@ spec:
     metadata:
       labels:
         {{- include "jitsi-meet.jicofo.selectorLabels" . | nindent 8 }}
+      {{- range $label, $value := mergeOverwrite .Values.global.podLabels .Values.jvb.podLabels }}
+        {{ $label }}: {{ $value }}
+      {{- end }}
       annotations:
         checksum/config: {{ include (print $.Template.BasePath "/jicofo/configmap.yaml") . | sha256sum }}
         checksum/secret: {{ include (print $.Template.BasePath "/jicofo/xmpp-secret.yaml") . | sha256sum }}
+      {{- range $annotation, $value := mergeOverwrite .Values.global.podAnnotations .Values.jvb.podAnnotations }}
+        {{ $annotation }}: {{ $value }}
+      {{- end }}
     spec:
     {{- with .Values.imagePullSecrets }}
       imagePullSecrets:
--- a/examples/helm/templates/jicofo/xmpp-secret.yaml
+++ b/examples/helm/templates/jicofo/xmpp-secret.yaml
@@ -1,0 +1,11 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "call-nested" (list . "prosody" "prosody.fullname") }}-jicofo
+  labels:
+    {{- include "jitsi-meet.jicofo.labels" . | nindent 4 }}
+type: Opaque
+data:
+  JICOFO_AUTH_USER: '{{ .Values.jicofo.xmpp.user | b64enc }}'
+  JICOFO_AUTH_PASSWORD: '{{ default (randAlphaNum 10) .Values.jicofo.xmpp.password | b64enc }}'
+  JICOFO_COMPONENT_SECRET: '{{ default (randAlphaNum 10) .Values.jicofo.xmpp.componentSecret | b64enc }}'
--- a/examples/helm/templates/jvb/service.yaml
+++ b/examples/helm/templates/jvb/service.yaml
@@ -18,6 +18,7 @@ spec:
       {{- end }}
       protocol: UDP
       name: rtp-udp
+    {{- if and .Values.jvb.enableTCP }}
     - port: {{ .Values.jvb.TCPPort }}
       targetPort: rtp-tcp
       {{- if eq .Values.jvb.service.type "NodePort" }}
@@ -25,6 +26,7 @@ spec:
       {{- end }}
       protocol: TCP
       name: rtp-tcp
+    {{- end }}
   {{- with .Values.jvb.service.externalIPs }}
   externalIPs:
   {{ toYaml . | indent 2 | trim }}
--- a/examples/helm/templates/jvb/service.yaml
+++ b/examples/helm/templates/jvb/service.yaml
@@ -1,0 +1,34 @@
+{{- if or (and (kindIs "invalid" .Values.jvb.service.enabled) (not .Values.jvb.useHostPort)) .Values.jvb.service.enabled }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "jitsi-meet.jvb.fullname" . }}
+  labels:
+    {{- include "jitsi-meet.jvb.labels" . | nindent 4 }}
+spec:
+  type: {{ .Values.jvb.service.type }}
+  {{- with .Values.jvb.service.LoadbalancerIP }}
+  loadBalancerIP: {{ . }}
+  {{- end }}
+  ports:
+    - port: {{ .Values.jvb.UDPPort }}
+      targetPort: rtp-udp
+      {{- if eq .Values.jvb.service.type "NodePort" }}
+      nodePort: {{ .Values.jvb.UDPPort }}
+      {{- end }}
+      protocol: UDP
+      name: rtp-udp
+    - port: {{ .Values.jvb.TCPPort }}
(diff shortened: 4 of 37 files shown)


funkyfuture:
my impression is that this is suited for a separate repository. because it uses a completely other too for deployment.

Zempashi:
I'm fine with that. I initially wanted to made a PR to https://github.com/helm/charts, but as state in the main readme of the repo:

```
Nov 13, 2019 | At Helm 3's public release, new charts are no longer accepted to stable or incubator.  Patches to existing charts may continue to be submitted by the  community, and (time permitting) reviewed by chart OWNERS for acceptance
```
There is also a branch call k8s-helm which is in early stage of development (so early that I decide to start from sratch, sorry...)

The main advantages of being in this repo:
- as it use the same docker images, it's easier to have a CI pipeline, that release helm chart in sync with new image release (for deployment in kubernetes the usage of `latest` is strongly discouraged, as kubernetes may decide to move container to other nodes, resulting of having diffferent images/unsolicited upgrade in a set of normally identical containers.)
- PR that change API (env variable), could change both deployment in the same time

The main advantages of being in a separated repo:
-  if CI toolchain size increases, it could be easier/clearer to split it between repo, and if bug apply on helm, it doesn't prevent the release of docker images. 
- PR that change API, will be smaller (so more reviewable) and fewer as it need an extra PR to fix the other repo, then need for synchronous release of both project.

There is also less technical aspect: do maintainers of this repo want to be exposes to issue being reported for chart bug, regression/improvement breaking the chart, kubernetes, etc...
I already saw few PR likely related to kubernetes/openshift (non-root in docker user, which is enforceable on kubernetes).

The border is blur (from what is docker related, what is kubernetes-related), but I would totally understand a negative answer. The best alternative would be then to have a `jitsi/kubernetes-jitsi-meet` so that related projects are close together, limiting the chance people would issue PR that break API ignoring the existence of a related project. (which could be less maintained than the docker-jitsi-meet for exemple)

saghul:
We have an examples directory now, where this could go. However, it shoulld follow thee docker-compose behavior a bit more. That is, it users shouldn't provide their own prosody chart / instance / whathaveyou, it should be provided out of the box.

Zempashi:
Hmm I'm not sure I understood wrongly. Just to clarify, there is a prosody provided out of the box, it just "as a dependecy". I use that trick to be able to disable the prosody part completely and let the user to provide their own instance (but that's definitely not the default, the default come with a built-in prosody)

So when you say:
```
That is, it users shouldn't provide their own prosody chart 
```
You mean:
```
users must not provide their own prosody chart 
```
they must use the one that is built-in ? I did this because I saw a "XMPP_SERVER" option.
I'm perfectly fine with that, It is not perfectly clear to me, so prefer to ask ;)

I will move the helm as `examples/helm` if that seems correct to you ?

saghul:
> Hmm I'm not sure I understood wrongly. Just to clarify, there is a prosody provided out of the box, it just "as a dependecy". I use that trick to be able to disable the prosody part completely and let the user to provide their own instance (but that's definitely not the default, the default come with a built-in prosody)

Ah, I had misunderstood, this sounds fine!

> I will move the helm as `examples/helm` if that seems correct to you ?

Yeah, that sounds great!

Zempashi:
Done !

examples/helm/README.md
```diff
@@ -0,0 +1,124 @@
+# jitsi-meet
+
+[jitsi-meet](https://jitsi.org/jitsi-meet/) Secure, Simple and Scalable Video Conferences that you use as a standalone app or embed in your web application.
+
+## TL;DR;
+
+```console
+$ git clone https://github.com/jitsi/docker-jitsi-meet
+$ helm install ./docker-jitsi-meet/k8s/helm
```
saghul: Thiss needs an update doesn't it?

Zempashi: Ah good catch !

rouja:
Hi,

It seems that I arrived to late with my chart : https://github.com/rouja/jitsi-chart :(

@Zempashi it seems that your chart is better than mine. What do you think about include an option to use cert-manager ?

Zempashi:
Hehe sorry ;)
Yes, not sure what do you mean by an option to use cert-manager ? It's already compliant in a way it let you annotate the ingress, here is the value file I use:

```
fullnameOverride: jitsi
web:
  ingress:
    enabled: true
    annotations:
      cert-manager.io/cluster-issuer: letsencrypt-prod
    tls:
      - hosts:
        - jitsi.redacted.com
        secretName: jitsi-tls-certificate
    hosts:
    - jitsi.redacted.com
jicofo:
  xmpp:
    password: "<redacted>"
    componentSecret: "<redacted>"

jvb:
  xmpp:
    password: "<redacted>"
  useHostPort: true
  useNodeIP: true
```
As you see it use cert-manager with letsencrypt but could be any cert-manager Issuer or ClusterIssuer. 
I saw you include `Certificate` object (which in my case is created/managed by cert-manager itself via the ingress object). Do you have a setup where cert-manager can't read ingress object ?
Or maybe you didn't enable Ingress at all ? (but cert-manager can't work without an Ingress ?)
I'm a bit confused and don't know what to do...

rouja:
@Zempashi I didn't knew this feature of cert-manager ^^ It's much simpler like that... 

"ah with a Self-signed CA, or DNS challenge you don't really need an ingress, I see a use case here, can you confirm ?"
Yes, I discovered the usage of cert-manager with a mail project that do not use ingress because it's only smtp and imap.

Thanks a lot to point me to this use case.

rouja:
Hi,

I did some test of your chart and I have some observations :

1- Depending of the number of helm deploy you do to adjust your configuration and if you don't fix any password, it seems that you can be in an incoherent situation where the value of JVB_AUTH_PASSWORD differs between jvb and prosody which result in error on the web front

2- The web pod take a long time to start because it generates a certificate which is probably not used in our context. What do you think about to put DISABLE_HTTPS: "1" by default on the web pod to avoid this behaviour ?

3- What is your opinion about using an externalIP on the service to expose jvb ?
```
NAME                   TYPE        CLUSTER-IP      EXTERNAL-IP    PORT(S)                      AGE
jitsi-jitsi-meet-jvb   ClusterIP   XXX.XXX.XX.XX   XX.XXX.XX.XX   10000/UDP,4443/TCP           30m
```

Zempashi:
@rouja,

- 1/ Yes, this is a problem with helm that occur with many chart (postgresql for example), but random string are chosen for security sake.
Many chart also has support for "existingSecret", to use an existing secret deployed with another tool. This one does not (yet) support this.
I still wondering how to fix this issue (hardcoding password ? for me this looks like a bad idea as every installation on the world get the same password)

- 2 Indeed, will do that

- 3 That's a good Idea, I'll add an option for that

Zempashi:
@rouja 
- DISABLE_HTTPS: "1", was supposedly the default, except that I did a typo on variable name, that's now fixed (the option is httpEnabled, and is by default to false)

- Options for externalIPs added. 
Warning: I'm not able to test this fully on my cluster. 
TL;DR; If you use this option you need to know what you are doing.

My concern is about setting web.service.type=Loadbalancer (the cluster should allocated "a loadbalancer and an IP" (<= cloud magics here) automatically and set externalIPs field. If you set also web.service.externalIPs either kubernetes don't let you do that, or obey and overwrite the one he wants to put (then you loose connectivity to you LoadbalancerIP). The last possibility is that kubernetes take some time to allocate the loadbalancer and then merge the IP to the IP list (which would be great), but then if you upgrade you will remove the loadbalancer IP from the list.

Zempashi:
@rouja some reading about secrets: https://github.com/helm/charts/issues/5167, I still don't see an easy option here. May be you smarter than me

ananace:
It'd be helpful if the chart supported disabling TCP - since multi-protocol loadbalancers is something that Kubernetes doesn't allow yet - unless you're on a supported cloud platform.
MetalLB has a workaround in place that let you support it anyway, though it's rather messy to set up and probably better left as an exercise to the user - involves sorting protocols into separate duplicate services with an added annotation and an explicit IP.

Something like this would at least let you set the TCP port to `null` in order to disable TCP, which isn't the most beautiful solution to that issue, but it works;
[0001-Support-disabling-TCP-Port.patch.txt](https://github.com/jitsi/docker-jitsi-meet/files/4389496/0001-Support-disabling-TCP-Port.patch.txt)

Zempashi:
> Additionally, prosody is lacking a hash annotation for the current config/secrets, so it's not restarted if the secrets or configuration changes during a helm upgrade 

=> yes, and it's difficult to be done due to the fact that it's in subchart, so I need to move it the template directory and add some {{- if }}{{- end }} (which I wanted to avoid in first shot) to allow disabling prosody entirely

rouja:
Hi,

@Zempashi I think that there is not easy solution right now. I think we have different ways to address the problem :

- As suggest @ananace, we can use a hash annotation on deployments to force a restart on all component all the time
- We can remove the magic with secrets automatically generated and let them empty by default with a failed if people do not configure them
- Add a variable like a "lock_secrets" to tell helm to not modify them

I have no more idea...

vitobotta:
@Zempashi I have installed Jitsi Meet with your Helm chart and enabled auth / disable guests. When I create a meeting I am required, as expected, to enter username and password, but where do I set them? From the values.yaml I am not sure which settings are relevant. Thanks!

rouja:
Hi @vitobotta 

I'm not sure that there is a setting for that.  In order to create a user / password, I   run this command in the prosody pod : 
```
prosodyctl --config /config/prosody.cfg.lua register USER DOMAIN PASSWORD
```
By the way, I'm not sure if there is the persistence for this.

vitobotta:
@rouja Gotcha. Thanks

Zempashi:
will look at persistence for prosody as soon as possible

sudheendrakeshav:
@Zempashi I am trying to host this on AKS and Option 1 it's not working out well for me as it throws me an error saying kubernetes cannot create an external load Balancer with multiple protocols. Would be really helpful for your suggestion on which option to proceed. Could you provide me some guidance on setting this up with Option 4 for jvb also please?

gynter:
@Zempashi, thanks for that Helm chart, you're da MVP :boom: ! I set it up on DigitalOcean managed Kubernetes. I configured LB as [Nginx ingress with cert-manager](https://www.digitalocean.com/community/tutorials/how-to-set-up-an-nginx-ingress-with-cert-manager-on-digitalocean-kubernetes). Disabling `jvc.TCPport` is required. I also created PVC for Prosody `/config/data` (to make accounts permanent) and ConfigMap & volume for Jitsi Web `/defaults/config.js` so I can overwrite default config settings (like language etc...).

@Zempashi, if possible, please push the `appVersion` to current in Your PR branch (also rebase with upstream `master` branch) since there have been new releases.

Also, there's a tiny mistake in the `values.yaml` (or `templates/web/ingress.yaml:31`, depends how you ment it to work). One solution is to modify `values.yaml` like this (hosts should be a list of hosts):

```
web:
    ...
     hosts:
-      - host: jitsi.local
-        paths: []
+      - jitsi.local
```

Zempashi:
@gynter glad to hear that ! I think with the couple of reports people did here, the most reasonable action is to disable TCP by default. Will do storage for prosody (and move to a statefulset, that make more sense). Thanks for pointing the `/defaults/config.js`, there is indeed lot of tunable here that I was ingoring existence. My stance here, till I fully understand the config options, would do nothing except adding couple options ('extraVolumes`, `extraVolumeMounts`, etc..) to let people work around. 
Thanks also for reporting the ingress problem, will address that too.

Try to do that now, stay tune.

Zempashi:
@gynter Done, tests and comments are more than welcome !

gynter:
@danjenkins Since I have only one node in cluster I used NodePort. For UDPport I specified 30300, TCPport set to false. Then opened 30300 UDP to all IPv4 from DigitalOcean control panel (Networking -> Firewalls) - by default it should be open for new nodes, but I like to keep things clean and secure ;) . For JVB PublicIP I specified this node (not external LoadBalancer) IP Address. That way Jitsi web (client, front-end) connects directly to NodeIP:30300 (JVB, back-end). For namespace I just used `jitsi` not `default`

If you have multinode cluster then I recommend to provision JVB on only one specific node and use this node IP. You can't have both TCP & UDP protocol load balancing on same external LoadBalancer (nginx-ingress) and I haven't found a way to set-up multiple external LoadBalancers (i.e different ngixn-ingress for TCP and UDP) in DO managed Kubernetes. I also tried to force Jitsi and JVB to use TCP only and tried to load balance the TCP JVB traffic via nginx-ingress, but failed and fell back to UDP and NodePort.

Also see my [values-do.yaml.txt](https://github.com/jitsi/docker-jitsi-meet/files/4450923/values-do.yaml.txt).

examples/helm/templates/jicofo/xmpp-secret.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Secret
+metadata:
+  name: {{ include "call-nested" (list . "prosody" "prosody.fullname") }}-jicofo
+  labels:
+    {{- include "jitsi-meet.jicofo.labels" . | nindent 4 }}
+type: Opaque
+data:
+  JICOFO_AUTH_USER: '{{ .Values.jicofo.xmpp.user | b64enc }}'
+  JICOFO_AUTH_PASSWORD: '{{ default (randAlphaNum 10) .Values.jicofo.xmpp.password | b64enc }}'
+  JICOFO_COMPONENT_SECRET: '{{ default (randAlphaNum 10) .Values.jicofo.xmpp.componentSecret | b64enc }}'
```
MichielDeMey, line 11: You specified `component_password` in the `values.yaml`, but you are using `componentSecret` here.
Caused an authentication issue in my cluster because it will use a generated password instead of the defined one. 🙂

rouja:
Hi,

@Zempashi I just made some tests and the persistence for prosody seems OK. I also tested without TCP and every thing seems to work.

examples/helm/templates/jicofo/xmpp-secret.yaml
Zempashi: Indeed, I will fix that ;)

ChrisTomAlx:
Same here tested without TCP using nginx ingress controller and a loadbalancer for jvb. Giving the loadbalancerIP and the publicip a static ip address I created using gcloud compute addresses (Make sure the region the static address is created is same as the region of the cluster)

This works on GCP as expected so far.

Thanks a lot for your great work @Zempashi and also everyone else involved at jitsi. I hope to contribute back soon once I dive a little deeper.

Cheers :)

examples/helm/templates/jicofo/xmpp-secret.yaml
Zempashi: And now fixed...

vkruoso:
Thanks @Zempashi for this work. How are your thoughts on using the chart for a HA deployment? I did test your chart and my issue is with jvb needing to have its public IP as its config. This is a problem to scale it on multiple nodes. It seems that the jvb scaling can be done locally simply by adding more jvb processes with non-conflicting configs, and the regional scaling should be done with octo.

I think that deploying like the official jitsi service, with multiple jvb's in different reagions, we might need to a different approach than Kubernetes.

examples/helm/templates/jvb/service.yaml
```diff
@@ -0,0 +1,34 @@
+{{- if default (not .Values.jvb.useHostPort) .Values.jvb.service.enabled }}
```
aderuelle, line 1: Setting  `jvb.service.enabled` to `false` doesn't work as I expected : the test is evaluated as `true` and the resource is still created.

I think there's some confusion around `default` usage, see for example this [Helm issue](https://github.com/helm/helm/issues/3308#issuecomment-585349501).

A workaround bellow (doesn't handle the case of `jvb.service` set to `null` though)

```suggestion
{{- if and .Values.jvb.service.enabled (not .Values.jvb.useHostPort) }}
```

Zempashi:
@vkruoso HA is a difficult things to achieve. Talking only about jvb, (prosody is a totally different story).
The only way I think it's possible now, is to have a kubernetes with node and public attach to it (preferably via 1to1 NAT and security groups or equivalent). Then use can use `useHostPort` and `useNodeIP` (option 3). I have achieved a working multiple jvb setup this way on a toy cluster with public IP on nodes (bad bad bad for security). So the right answer might be effectively that kubernetes is not (yet ?) a proper HA tool for jitsi. (But still, having a helm chart is quite handy considering how easy it is to have a cluster up and running on Google or Amazon or whatelse)

EDIT: link to my branch updated, to see the rendered markdown. I already wrote few words about multiple jvb.

PS: I write the PR trying to let people decide what best network setup for their own use case. I was also hoping that someone find a better alternative than the one I found.

examples/helm/templates/jvb/service.yaml
Zempashi: Argggg, stupid helm documentation, https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function. This is hugely misleading. So now I believe that all the default statement are wrong. Will check that soon

saghul:
Agreed on the fact that it need to be perfect to start with. I’d be ok with merging the current state, since it’s an example and not an exhaustive setup.

Is this ready, given the above?

Zempashi:
@saghul let me check @aderuelle problem (could also hide a lot more problems)

vkruoso:
Thanks for your response @Zempashi. I agree that the chart is very good and can be merged as it is actually working great, since HA is a different story.

bhack:
I think that could be still very interesting to have a new specific HA issue after this is merged.

Zempashi:
@saghul last issue fixed and branch rebased: ok to merge !

kpeiruza:
Hi @Zempashi, thanks for you great work on this Chart.

I wanted to create a Helm as well, but got quite overloaded at work and just made Jibri work on K8s so far, which is the only key piece I'm missing on your Helm chart. It's able to work in machines without /dev/snd (migrated from ALSA to Pulseaudio), so I don't need an init container loading snd-aloop (which in my personal K8s provider wasn't available).

I've been working on a Docker image that lets JVB to configure itself what's it's IP address, based on "master" branch so I can get cool Octo-Bridges working, but I'm running them as several nodes with just Docker on them. My next step was to migrate them to Kubernetes.

In order to avoid problems with Ingress, I was thinking in using a DaemonSet for JVB, so you get the real hostPort instead of the behaviour of a nodePort or a LoadBalancer, where you can't control which node is going to answer your request once you have more than one.

Fancy working together to get a fully-featured Jitsi Helm Chart? :-)

Zempashi:
hi @kpeiruza 

I should admit that Jibri is one of the component that is really obscure to me. I don't really understand what it is trying to achieve. Looking a jibri README.md, requirements make it hard for docker. I would suggest that work could go in a separate PR on it's own (but I'm not affiliated to jitsi in any way).

JVB is the tricky part of porting jitsi to kubernetes: what you describe with daemonset and hostPort is done similarily in the helm chart with hostPort and nodeAffinity, however it let's you control the number of jvb as opposed to daemonset that made you run as many jvb as you have nodes.

This chart has been suitable for merging (@saghul does it ?), so adding more unfinished work will delay the "General Availability" of the chart. 
But would be happy to help on new PR(s)

thapakazi:
@saghul **let's merge this,** 
we need this asap for our flight :rocket:  to the mars  :neckbeard: ...just kidding.  :rofl:

guice:
Count me as a +1 for this. I'm about to download it and try it out.

ChrisTomAlx:
If you choose to use nodeIP instead of public IP (useful for jvb scaling). Remember to allow the UDP port through firewall and also add the following extra envs to your jvb deployment using values.yaml.
```yaml
    - name: JVB_STUN_SERVERS
      value: stun.l.google.com:19302,stun1.l.google.com:19302,stun2.l.google.com:19302
    - name: JVB_TCP_HARVESTER_DISABLED
      value: "true"
    - name: JVB_OPTS
      value: "--apis=xmpp,rest"
    - name: ENABLE_STATISTICS
      value: "true"
```

Another thing I would like to mention is not to use podaffinity and podantiaffinity within the chart because as per kubernetes docs, clusters with greater than 100 nodes take a long time to satisfy these conditions. And if ever jibri comes into this chart there will be a whole lot of nodes that pop in and out of existence simultaneously

Zempashi:
@ChrisTomAlx 
the default value of `JVB_TCP_HARVESTER_DISABLED` is now `1` (tcp disabled by default), should I move it to `true` ?

ChrisTomAlx:
@Zempashi 

Stun server is the main one. I am assuming `1` and `true` are the same so there is not much to worry there.

I did not test it with 100+ nodes. But the docs suggest against using podaffinity and podantiaffinity. The first blockquote [here](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity) says :- 

> Note: Inter-pod affinity and anti-affinity require substantial amount of processing which can slow down scheduling in large clusters significantly. We do not recommend using them in clusters larger than several hundred nodes.

It does say several hundred nodes, so I guess its alright unless the cluster is massive. Still thought I would let you know this could be an issue for someone deploying it on that kind of scale. Guess that can be taken up as a future enhancement. Or maybe it could be added to the readme to make users aware of such a problem that they may not be expecting

ananace:
Usually, (anti-)affinities are left up to the user to configure, but as jvb will not work with multiple instances on the same node - unless a bunch of other configuration is done, perhaps it'd be better to simply default to providing an anti-affinity and adding a note to both the chart readme as well as the notes.

But generally, if you're working with a massive cluster you tend to be aware of these limitations, and know to look for them.

kpeiruza:
Once you reach that cluster size, it makes sense to deploy several shards.Prosody is single threaded, you don't want to have 1 single prosody+ 1 jicofo for 100 videobridges with an average of 50 users/bridge.Jicofo and prosody will be the bottleneck there (and even the max connections on the webserver).Regards,KennethOn Jun 15, 2020 11:52, Christy Thomas Alex <notifications@github.com> wrote:
@Zempashi
Stun server is the main one. I am assuming 1 and true are the same so there is not much to worry there.
I did not test it with 100+ nodes. But the docs suggest against using podaffinity and podantiaffinity. The first blockquote here says :-

Note: Inter-pod affinity and anti-affinity require substantial amount of processing which can slow down scheduling in large clusters significantly. We do not recommend using them in clusters larger than several hundred nodes.

It does say several hundred nodes, so I guess its alright unless the cluster is massive. Still thought I would let you know this could be an issue for someone deploying it on that kind of scale. Guess that can be taken up as a future enhancement. Or maybe it could be added to the readme to make users aware of such a problem that they may not be expecting

ChrisTomAlx:
@ananace @kpeiruza 
I agree with both of you. This just needs to be mentioned somewhere in the readme. That would be enough. Atleast that is my humble opinion.

thapakazi:
guys, comon, lets merge this :panda_face:

LorbusChris:
:shipit:

examples/helm/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Option 1: LoadbalancerIP
+
+This requires a cloud setup that enables a Loadbalancer attachement.
+This could be enabled via values:
+
+```yaml
+jvb:
+  service:
+    type: LoadbalancerIP
```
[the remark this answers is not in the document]
Zempashi: Thanks for spotting typo !

Zempashi:
Rebase/updated to last stable images

kuzm1ch:
@Zempashi  
could you also take a look at this error pls
```
cd examples/
helm dep up ./helm
...
...Successfully got an update from the "stable" chart repository
Update Complete. ⎈Happy Helming!⎈
Saving 1 charts
Dependency prosody did not declare a repository. Assuming it exists in the charts directory
Error: Dependency prosody has an invalid version/constraint format: improper constraint: 
```
it can be fixed by removing these lines. 
```
# dependencies:
#   - name: prosody
#     condition: prosody.enabled
```
If a chart has been already in chart subdirectory, maybe there is no need to describe it as a dependency to avoid helm error.

kuzm1ch:
fyi: this helm chart was deployed on GKE with option 2(NodePort and node with Public IP). It's working(except of error with helm dep up which don't affect helm install).

```
helm: v3.2.4
k8s: v1.16.9-gke.6
```

Zempashi:
I just add a `version: '*'`, so that dep build/dep up work as expected. This chart as currently no dependency (may be one day prosody could live it's own life separated from jitsi, but today it's really tied up to docker-jitsi), so you don't realy need the `helm dep` commands.
Anyway; it's always better to be compliant, so thanks for reporting the issue !

examples/helm/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Option 1: Loadbalancer
+
+This requires a cloud setup that enables a Loadbalancer attachement.
+This could be enabled via values:
+
+```yaml
+jvb:
+  service:
+    type: Loadbalancer
```
AkeemMcLennon, line 31:
```suggestion
    type: LoadBalancer
```

There is a typo here for the service type

Zempashi: Shit, I already tried to fix this typo, replaced by another typo.

examples/helm/templates/jicofo/deployment.yaml
```diff
@@ -0,0 +1,66 @@
...
+          envFrom:
+          - secretRef:
+              name: {{ include "call-nested" (list . "prosody" "prosody.fullname") }}-jicofo
+          - configMapRef:
+              name: {{ include "call-nested" (list . "prosody" "prosody.fullname") }}-common
+          - configMapRef:
+              name: {{ include "jitsi-meet.jicofo.fullname" . }}
+          ports:
+            - name: http
+              containerPort: 80
```
petarGitNik, line 42: jicofo is XMPP component, should it expose port 80, or ports 5222 and 5347?

kuzm1ch:
strange why this helm chart is still not merged...

examples/helm/templates/jicofo/deployment.yaml
Zempashi: Indeed, that'a mistake. There a port 8888, that I found open. I use this port in the readiness/liveness probe (but ignoring it's purpose).

Zempashi:
@kuzm1ch I don't know either. 

Looks like @saghul is busy. Or something missing, but I don't know what.

I'm now considering putting the chart on my company repository https://github.com/enix/helm-charts
But this is sub-obtimal as bugfixes should hit this repo first, not an extra repository with unrelated stuff.

zzvara:
It would be nice if this repo would merge.

xrow:
@saghul Can you give this topic some love? I think we need to know what comes next. If you merge, we can propose more chagens based on that.

0x46616c6b:
A nice addition for the Helm Chart would be the possibility to add labels and annotations per Pod. So for example to adjust Sidecars (Istio/Linkerd) or set Annotations for log processing.

Zempashi:
@0x46616c6b tell me if you find last commit useful (or not) ;)

0x46616c6b:
Really cool and so fast! Thanks a lot! We will investigate why it not working with Istio per default and maybe come back with a solution. 

Can I give you another hint or maybe I can add this as PR in your branch: I'm maintaining a jitsi-meet-exporter for Prometheus. Would be great to have them also in the Helm Chart as a Option: https://github.com/systemli/prometheus-jitsi-meet-exporter

Zempashi:
@0x46616c6b, If you can PR it, that's best !

saghul:
PSA: https://github.com/jitsi/docker-jitsi-meet/issues/991

saghul:
Closing as this will be moving to https://github.com/jitsi-contrib If anyone wants to be added to the org to work on this, please drop a comment here.

sapkra:
https://github.com/jitsi-contrib/helm
@Zempashi I invited you to this repository. All the other community members can create PRs to improve the chart as fast as possible.

Zempashi:
Thanks, sorry lacking a bit of time this week ;)

PR #238, opened 2020-03-19
ldap: fix anonymous binds without breaking template rendering (Fixes #234)

### LINKED ISSUE #234: ${CONFIG}/prosody/saslauthd.conf template is not filled
After configuring LDAP configuration in the .env file, all authentication fails.
The contents of /config/saslauthd.conf in the prosody container are as follows:

    {{ if eq (.Env.AUTH_TYPE | default "internal") "ldap" }}
    ldap_bind_dn: {{ .Env.LDAP_BINDDN }}
    ldap_bind_pw: {{ .Env.LDAP_BINDPW }}
    ldap_filter: {{ .Env.LDAP_FILTER | default "uid=%u" }}
    ldap_version: {{ .Env.LDAP_VERSION | default "3" }}
    ldap_auth_method: {{ .Env.LDAP_AUTH_METHOD | default "bind" }}
      {{ if .Env.LDAP_USE_TLS | default "0" | toBool }}
    ldap_tls_key: /config/certs/{{ .Env.XMPP_DOMAIN }}.key
    ldap_tls_cert: /config/certs/{{ .Env.XMPP_DOMAIN }}.crt
        {{ if .Env.LDAP_TLS_CHECK_PEER | default "0" | toBool }}
    ldap_tls_check_peer: yes
    ldap_tls_cacert_file: {{ .Env.LDAP_TLS_CACERT_FILE | default "/etc/ssl/certs/ca-certificates.crt" }}
    ldap_tls_cacert_dir: {{ .Env.LDAP_TLS_CACERT_DIR | default "/etc/ssl/certs" }}
        {{ end }}
        {{ if .Env.LDAP_TLS_CIPHERS }}
    ldap_tls_ciphers: {{ .Env.LDAP_TLS_CIPHERS }}
        {{ end }}
      {{ end }}
    {{ end }}

It looks as if a templating step wasn't triggered during `docker-compose up`.
The logs of the prosody container don't contain anything about a malformed saslauthd.conf file.

After manual configuration of the saslauthd.conf file, everything works as expected.

What did I do wrong? Is this expected?

elor:
Commit e92a00ca199ab4f64acc39a245f354164ce8f13a took the first steps towards fixing `<no value>` expansions during rendering for LDAP_BINDDN and LDAP_BINDPW.

In our deployment, this breaks template rendering if LDAP_BINDDN is not defined.
Only the first two lines of the template are evaluated.

A simple solution is provided in this pull request: expand LDAP_BINDDN to a default value and cast it to a boolean, similar to line 14 of the same file.

This fixes #234

### DIFF
--- a/prosody/rootfs/defaults/saslauthd.conf
+++ b/prosody/rootfs/defaults/saslauthd.conf
@@ -1,7 +1,7 @@
 {{ if eq (.Env.AUTH_TYPE | default "internal") "ldap" }}
 ldap_servers: {{ .Env.LDAP_URL }}
 ldap_search_base: {{ .Env.LDAP_BASE }}
-{{ if .Env.LDAP_BINDDN }}
+{{ if .Env.LDAP_BINDDN | default "" | toBool }}
 ldap_bind_dn: {{ .Env.LDAP_BINDDN }}
 ldap_bind_pw: {{ .Env.LDAP_BINDPW }}
 {{ end }}


saghul:
Do you have it not defined or defined as empty?

elor:
I did not define it, since it's an anonymous bind.
That's how I understand env.example

Quote from env.example, LDAP_BINDDN:
https://github.com/jitsi/docker-jitsi-meet/blob/76acc6502734d9bca5fa78b67102cfbc7a4ad29c/env.example#L104

In contrast: LDAP_BASE must be defined, but can be empty:
https://github.com/jitsi/docker-jitsi-meet/blob/76acc6502734d9bca5fa78b67102cfbc7a4ad29c/env.example#L101

What I take from this:
1. LDAP_BINDDN and LDAP_BINDPW defined? --> pass them to saslauthd.conf
2. LDAP_BINDDN not defined? --> anonymous bind. Omit them in saslauthd.conf

Option 2 is currently not working in master (rendering fails). It's also not working in the latest docker image (`<no value>` problem).

elor:
P.S. The proposed one-line change enables both methods:
Not defining LDAP_BINDDN as well as leaving it empty have the same result --> anonymous bind.

From what I can see, it's a non-conflicting commit without regression.
Current behavior is preserved, but it's more fault-tolerant.

timonegk:
Unfortunately, this seems to break non-anonymous bind.

timonegk:
Without ` | toBool`, it works fine for me.

elor:
What's your resulting saslauthd.conf file?
Is there an error message during prosody startup during template parsing via `/etc/cont-init.d/10-config` -> `tpl` -> `frep`?

My commit should have no effect on both anonymous and non-anonymous binds, but somehow it does.

For anonymous binds, it converts empty (`no value`) -> empty (`""`) -> empty (`false`)-> if-condition not fulfilled.
All of those are "empty values" in terms of the template language, which don't trigger the if-condition.

If LDAP_BINDDN is defined on the other hand, this becomes:
not empty (`some string`) -> still not empty (`some string`)-> not empty (`true`) -> if-condition fulfilled.

So, according to logic, my commit does nothing, but our deployment systems beg to differ and prefer not to render the template to completion without it.

Feel free to revert the merge as needed.

https://github.com/jitsi/docker-jitsi-meet/issues/234#issuecomment-602183580 contains some more of my digging.

timonegk:
Appareantly, `toBool` always evaluates to false in my test cases. Please verify the following behavior:
In the template: `{{ .Env.TEST | default "" | toBool }}`
Run: `TEST=examplevalue tpl templatefile` and `tpl templatefile`. For me, both return false.

timonegk:
Okay, the problem is that `toBool` (https://github.com/subchen/frep/blob/master/func.go#L58) uses `parseBool` which actually parses a boolean, instead of checking whether the passed value is 'truthy' (see https://golang.org/pkg/strconv/#ParseBool). In our case, we want to check if the value is set. To do that, the `toBool` must be omitted.
Check for example the template `{{ if .Env.TEST | default "" }}Conditional string{{ end }}` and run again with of both the commands in my previous comment.

Issue #220, opened 2020-03-17
Traefik Deployment questions


Dextroz:
Hello 😃 

I am looking into deploying Jitsi using Traefik. I am looking for basic audio and video call functionality.

For this, do I need to deploy `jigasi` ?

Furthermore:
1. If using Traefik, do I need to uncomment the `PUBLIC_URL` environment variable?
2. I don't require any recording/streaming of calls/video etc, how do I ensure this is disabled?

I assume I will need to ensure the relevant ports are open and forwarded correctly too.

Thanks in advance :-)

jakesmolka:
Hey! I don't know if you got your setup running and just what to get more into it - or you seek help getting your setup running. If it is the latter, I can give you my jitsi + traefik setup as example how it works for me.

From the perspective of my experience and not being an expert, regarding your questions: 
- I don't run `jigasi`. In fact, I didn't know what it was until now, because I just used the basic [docker-compose file](https://github.com/jitsi/docker-jitsi-meet/blob/master/docker-compose.yml) provided here.
- I uncommented and added my value to the `PUBLIC_URL` in the `.env` file. But I'm not sure if it really is necessary.
- I would say, make sure `ENABLE_RECORDING` stays commented (i.e. not set to `1`) and you don't accidentally deploy `jibri`. Again, this is not the case if you only the the basic docker-compose file.

Dextroz:
Hi @jakesmolka,

Thanks for replying. Would you able to share your example please? 

Thanks.

saghul:
We have an examples directory in the dev branch, you are welcome to contribute your config with a PR so we can add it there!

SebastianKobs:
Hi, working example(at least for me) with traefik.


- Remove the port-bindings for the web service. 
- Add your external network (e.g. web) in the network-sections 
- Add your labels.  

I've added comments to the relevant parts, otherwise it's the default.
Make sure to uncomment and add your host ip in .env : 
````
DOCKER_HOST_ADDRESS=1.2.3.4
````
Example docker-compose.yml with traefik(still running on traefik 1.7). 
````
version: '3'

services:
    # Frontend
    web:
        image: jitsi/web
        #traefik
        #ports:
        #    - '${HTTP_PORT}:80'
        #    - '${HTTPS_PORT}:443'
        #traefik end
        volumes:
            - ${CONFIG}/web:/config
            - ${CONFIG}/web/letsencrypt:/etc/letsencrypt
[...]

PR #258, opened 2020-03-21
minimal example to run jitsi behind traefik

SebastianKobs:
Hi, 
adding to my comment in https://github.com/jitsi/docker-jitsi-meet/issues/220
I've added an minmal example for traefik.
Kind regards
Seb

### DIFF
--- a/examples/traefik/README.md
+++ b/examples/traefik/README.md
@@ -1,0 +1,14 @@
+# Basic configuration to use with the traefik reverse proxy
+
+Note: Tested with traefik 1.7
+
+- Remove the port-binds for the web service
+- The provided example assumes an external network with the name "web"
+- I've added comments where to adjust the docker-compose.yml, but I assume that you know your stuff, if you're running traefik.
+
+Uncomment and set DOCKER_HOST_ADDRESS in .env  
+Im pretty sure, that this is mandatory for the docker-setup and should be clearer in the original README. Could be the proxying, didn't investigate further.
+
+````env
+DOCKER_HOST_ADDRESS=136.243.57.154
+````
--- a/examples/traefik/README.md
+++ b/examples/traefik/README.md
@@ -2,13 +2,13 @@
 
 Note: Tested with traefik 1.7
 
-- Remove the port-binds for the web service
-- The provided example assumes an external network with the name "web"
-- I've added comments where to adjust the docker-compose.yml, but I assume that you know your stuff, if you're running traefik.
+- If you're running behind traefik, I consider it a better practice to remove the port-binds for the web service
+- The provided example assumes an external network with the named "web". This is the network you most likely have created while setting up traefik.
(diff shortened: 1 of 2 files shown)


examples/traefik/README.md
```diff
@@ -0,0 +1,14 @@
+# Basic configuration to use with the traefik reverse proxy
+
+Note: Tested with traefik 1.7
+
+- Remove the port-binds for the web service
+- The provided example assumes an external network with the name "web"
+- I've added comments where to adjust the docker-compose.yml, but I assume that you know your stuff, if you're running traefik.
```
funkyfuture: this is not appropriate for a documentation.

examples/traefik/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+Uncomment and set DOCKER_HOST_ADDRESS in .env  
+Im pretty sure, that this is mandatory for the docker-setup and should be clearer in the original README. Could be the proxying, didn't investigate further.
```
funkyfuture: do not guess what is needed and / or correct!

SebastianKobs: You're right on this. Didn't work for me without it, doesn't mean it's mandatory. Will change that.

examples/traefik/README.md
SebastianKobs: I've assumed(wrongly?) that people who wants to run jitsi behind traefik, know a bit about setting it up and just added the steps which I've done. I can write a complete setup-guide including setting up traefik from the getgo, if you consider this more helpful. Not sure, if the traefik-setup-part would be out of scope?

SebastianKobs:
>  ... move the exposure of web ports (as defined now in `docker-compose.yml`) to a separate compose file, i have no idea whether the maintainers would accept that.

You can leave them in, but I consider it a better practice to not expose them on the host, unless absolutely needed. As for maintaining it, I'll happily update the example when needed, but will see if I can come up with a better way. Thanks for your helpfull comments.

examples/traefik/README.md
funkyfuture: the point is that you're writing in the singular first person, unlike the rest of the document. also punctuation is missing, there are unnecessary line breaks in the markup. just use a hyperlink to reference traefik, don't point out the obvious ("you know your stuff, if …")

funkyfuture: yes, an introduction to traefik would be out of scope.

funkyfuture:
exposing unneeded ports to the public is not a good idea. it might cause confusion and makes them unavailable for other purposes.

are you aware of this Docker-Compose feature: https://docs.docker.com/compose/extends/ ?

SebastianKobs:
Yes I'm aware of it. Could use that, but this example stems from my own setup. I changed the wording in the README to include your suggestions,.

saghul:
Cheers!

PR #261, opened 2020-03-22
Adds a check for certbot's success and exits in case of a failure

funkyfuture:


saghul:
I think a better fix would be to `set -e` for the script. Whatever fails we want the whole thing to fail...

funkyfuture:
i agree. but i'd leave the pausing for the mentioned reasons in place. that okay?

saghul:
Makes sense. The `-e` is a follwop-up change that should be applied everywhere really.

saghul:
Thank you!

funkyfuture:
thank you for maintaining!

PR #262, opened 2020-03-22
Add helm options for namespace and node ports

chclaus:


sapkra:
There is also a new pull request which will add a helm chart: https://github.com/jitsi/docker-jitsi-meet/pull/235

sapkra:
Can you create a PR at our new helm repository if this is not already implemented? https://github.com/jitsi-contrib/helm

PR #192, opened 2020-02-21
misc: make all containers run without root

sapkra:
I had no time to test all containers again.

Untested:
- Jigasi
- Jibri (I'm sure it will not work on OpenShift/OKD -> see TODOs)
- Everything else has to be retested I think - that's a lot including LDAP, JWT etc. 

Changes:
- All containers are running with USER 1001 (all users are always in the root group)
- The containers have to run with random user IDs which is the case in OpenShift/OKD for example
- **BREAKING**: Ports have been changed to ports which can be used by non privileged users

PR #273, opened 2020-03-24
prosody: fix restart loop on rolling deployment

sapkra:
This PR is the revive of the reverted PR #190. The permission problem should be fixed now.

@saghul Should I change the path like you commented here? https://github.com/jitsi/docker-jitsi-meet/pull/190#issuecomment-589728732
I could try using the default `/var/run/prosody/prosody.pid`. WDYT?

FYI: This PR is currently based on #192.


saghul:
re-targeted for master, you caught me mid-change, I'm deprecating the dev branch and cutting out a release today, if things go ok.

saghul:
> I could try using the default `/var/run/prosody/prosody.pid`. WDYT?

Makes sense!

saghul:
Alright, I'm going to give this a test, thanks a lot for this @sapkra !

saghul:
Just a cosmetic thing: is there a way to make this look a bit better? `I have no name!@25b911148469:/$` :-)

saghul:
Just gave it a quick test and things look good! Nice work!

sapkra:
> > I could try using the default `/var/run/prosody/prosody.pid`. WDYT?
> 
> Makes sense!

I will try that. Then I will not need the permission change anymore I guess.



> Just a cosmetic thing: is there a way to make this look a bit better? `I have no name!@25b911148469:/$` :-)

That's part of #192. I've chosen the user 1001 because it's default and also the OpenShifts docs are saying that I should use a numeric user. But what I will do is dynamically adding the user ID to the passwd and assigning a username.

### OpenShift Docs
Lastly, the final `USER` declaration in the `Dockerfile` should specify the user ID (numeric value) and not the user name. This allows OpenShift Container Platform to validate the authority the image is attempting to run with and prevent running images that are trying to run as root, because running containers as a privileged user exposes potential security holes.
https://docs.openshift.com/container-platform/3.11/creating_images/guidelines.html#use-uid

sapkra:
Closing this because everything should be fixed in #192 now.

PR #274, opened 2020-03-24
README.md: correct command to add user

wojas:
There is no `register` command. The actual command is `adduser`, with different commandline arguments.


wojas:
Just noticed this is a duplicate of PR #228

PR #279, opened 2020-03-24
prosody: fix ldap config template

timonegk:
https://github.com/jitsi/docker-jitsi-meet/pull/238 caused a regression in the prosody ldap template that resulted in breaking non-anonymous ldap login. `toBool` converts values that are not in `1, t, T, TRUE, true, True` to false. More details can be found here: https://github.com/jitsi/docker-jitsi-meet/pull/238.

### DIFF
--- a/prosody/rootfs/defaults/saslauthd.conf
+++ b/prosody/rootfs/defaults/saslauthd.conf
@@ -1,7 +1,7 @@
 {{ if eq (.Env.AUTH_TYPE | default "internal") "ldap" }}
 ldap_servers: {{ .Env.LDAP_URL }}
 ldap_search_base: {{ .Env.LDAP_BASE }}
-{{ if .Env.LDAP_BINDDN | default "" | toBool }}
+{{ if .Env.LDAP_BINDDN | default "" }}
 ldap_bind_dn: {{ .Env.LDAP_BINDDN }}
 ldap_bind_pw: {{ .Env.LDAP_BINDPW }}
 {{ end }}


saghul:
Cheers!

prosody/rootfs/defaults/saslauthd.conf
```diff
@@ -1,7 +1,7 @@
 {{ if eq (.Env.AUTH_TYPE | default "internal") "ldap" }}
 ldap_servers: {{ .Env.LDAP_URL }}
 ldap_search_base: {{ .Env.LDAP_BASE }}
-{{ if .Env.LDAP_BINDDN | default "" | toBool }}
+{{ if .Env.LDAP_BINDDN | default "" }}
```
sapkra: Do we really need the default value?

timonegk: I think adding the default value was the entire point of #238.

timonegk:
Could you rebuild the docker hub images, please?

prosody/rootfs/defaults/saslauthd.conf
sapkra: I was trying to understand the problem but was not able to figure it out. The following commands are returning nothing so I can't see the problem with `<no value>`.
`LDAP_BINDDN` is unset on my machine.

```bash
echo "{{ if .Env.LDAP_BINDDN | default "" }}ldap_bind_dn: {{ .Env.LDAP_BINDDN }}{{ end }}" | frep -
```

```bash
echo "{{ if .Env.LDAP_BINDDN }}ldap_bind_dn: {{ .Env.LDAP_BINDDN }}{{ end }}" | frep -
```

timonegk: I see. So maybe #238 and therefore this PR are entirely useless?

sapkra: It seems so or I just don't understand the problem.

PR #287, opened 2020-03-25
prosody: update to latest stable version

saghul:
Use the official repositories so we don't lag behind in fixes.

### DIFF
--- a/prosody/Dockerfile
+++ b/prosody/Dockerfile
@@ -1,8 +1,13 @@
 ARG JITSI_REPO=jitsi
 FROM ${JITSI_REPO}/base
 
+ADD https://prosody.im/files/prosody-debian-packages.key /tmp/prosody.key
+
 RUN \
-    apt-dpkg-wrap apt-get update \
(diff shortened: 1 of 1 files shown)


saghul:
@damencho @aaronkvanmeerten WDYT?

saghul:
YOLO.

sapkra:
@saghul I've built the master branch last night and noticed that prosody is installed from the stretch-backports which is still at version 0.11.2 but the recent prosody version is 0.11.5.

When starting the new container I got:
```bash
startup             info	Hello and welcome to Prosody version 0.11.2
```

saghul:
Hum, I'll take a look, thanks for the heads up! I think I know the problem: the `-t stretch-backports`. I needed to add it back for lua-sec, but that undid my update. Oh well!

saghul:
Fixed in https://github.com/jitsi/docker-jitsi-meet/commit/93ba77021b3604df380174635a1267e01b04a18c

PR #291, opened 2020-03-25
Add build instructions to README

camisatx:
- Remove `Limitations` section from table of contents since it is not longer at the bottom
- Add table of contents section for `Build Instructions` and `ToDo`
- Add section with instructions for how to build the images

### DIFF
--- a/README.md
+++ b/README.md
@@ -439,7 +439,7 @@ The docker images can be built by running the `make` command in the main reposit
 
 If you are on the unstable branch, build the images with `FORCE_REBUILD=1 JITSI_RELEASE=unstable make`.
 
-Before deployment, make sure you remove the config file. You are now able to run `docker-compose up` as usual.
+You are now able to run `docker-compose up` as usual.
 
 ## TODO
 
--- a/README.md
+++ b/README.md
@@ -439,6 +439,8 @@ The docker images can be built by running the `make` command in the main reposit
 
 If you are on the unstable branch, build the images with `FORCE_REBUILD=1 JITSI_RELEASE=unstable make`.
 
+Before deployment, make sure you remove the config file. You are now able to run `docker-compose up` as usual.
+
 ## TODO
 
 * Support container replicas (where applicable).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -430,6 +431,16 @@ succeed. If your users are coming in over the Internet (and not over LAN), this
 The public IP address is discovered via [STUN]. STUN servers can be specified with the ``JVB_STUN_SERVERS``
 option.
 
+## Build Instructions
+
+Building your own images allows you to edit the configuration files of each image individually, providing more customization for your deployment.
```
saghul: Can you remove the extra space before the B?

README.md
```diff
@@ -430,6 +431,16 @@ succeed. If your users are coming in over the Internet (and not over LAN), this
...
+
+The docker images can be built by running the `make` command in the main repository folder. If you need to overwrite existing images from the remote source, use `FORCE_REBUILD=1 make`.
```
saghul: Ditto

README.md
```diff
@@ -430,6 +431,16 @@ succeed. If your users are coming in over the Internet (and not over LAN), this
...
+
+If you are on the unstable branch, build the images with `FORCE_REBUILD=1 JITSI_RELEASE=unstable make`.
+
+Before deployment, make sure you remove the config file. You are now able to run `docker-compose up` as usual.
```
saghul: What config file? You should need to remove anything.

README.md
camisatx: I think the space before `Build` is required for the md formatting, and I don't think there is a space before `Building`. The formatting in the diff may make it look like there is a space ;)

camisatx: Unless you are talking about the line 435? But that would conflict with the other titles in this doc.

README.md
camisatx: Ah, that was in regard to any changes done in `.env`, which requires the removal of `.jitsi-meet-cfg`. But that doesn't matter if we are only building the docker images here, so out of scope for the build process itself. Agree?

saghul: Agreed!

Disregard the spacing thing, I see it ok now

camisatx:
Okay I think we are good! Look good with you?

saghul:
Cheers!

PR #302, opened 2020-03-26
jvb: don't configure STUN mapping if DOCKER_HOST_ADDRESS is set

albe:
This works around #294


albe:
/cc @saghul 

alternatively add some documentation in the .env.example to outcomment STUN servers if the docker public address is set?

saghul:
There are cases where you need both, so we cannot go in this direction. The problem that needs solving in the bridge is that it sshouldn't fail to start if it fails the STUN checks.

PR #313, opened 2020-03-28
Add example traefik v2

ruby232:


saghul:
Thank you! Why do we need another one? Can’t the existing one be updated?

ruby232:
I've added a new one because if someone's still using Traefik 1.

saghul:
I don't know much about Traefik, but unless v1 is stilll supported and somewhat of an LTS, I'd say we only want to have examples for the latest 
.

ruby232:
OK, I'm going to create a new pull request.

saghul:
Oh, I can't reopen. Can you make a new PR again please? Sorry for the trouble.

ruby232:
@saghul not problem

PR #319, opened 2020-03-28
Example Traefik v2

ruby232:
Change the Traefik example to version 2

### DIFF
--- a/examples/traefik/README.md
+++ b/examples/traefik/README.md
@@ -1,10 +1,11 @@
 # Basic configuration to use with the traefik reverse proxy
 
-Note: Tested with traefik 1.7
+Note: Tested with traefik 2.1.3
 
 - When running behind traefik, it's a better practice to remove the port-binds for the web service.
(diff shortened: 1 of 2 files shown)


SebastianKobs:
Wouldn't it be better, to add the traeffk v2 docker-compose.yml as a separate file? Me and others are still running on 1.7.

sapkra:
This was his first approach: https://github.com/jitsi/docker-jitsi-meet/pull/313
My thoughts on this were: I don't know the traefik community but if someone is still using an older version he can still look at the file history.

SebastianKobs:
"Someone" is an understatement. The migration from v1.7 to v2 is not trivial if you have a large number of hosts or projects to upgrade. Traefik 1.7.24 is still actively maintained and available as an official image. So my votes still goes to keeping both examples.

ruby232:
@SebastianKobs I agree, but really i don't know what do..
* Reopen #313?
Please let me know

saghul:
Looks like a 1.7 releasse was made just 5 days ago: https://github.com/containous/traefik/releases/tag/v1.7.24 Let's have both then.

PR #322, opened 2020-03-28
Improve wording of README.md

frispete:
My first little contribution. Hope, you like it..

### DIFF
--- a/README.md
+++ b/README.md
@@ -373,7 +373,7 @@ Variable | Description | Example
 `GC_CLIENT_ID` | `client_id` from Google Cloud Credetials
 `GC_CLIENT_CERT_URL` | `client_x509_cert_url` from Google Cloud Credetials
 
-For setting the Google Cloud Credentials please read https://cloud.google.com/text-to-speech/docs/quickstart-protocol section "Before you begin" from 1 to 5 paragraph.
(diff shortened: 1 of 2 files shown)


sapkra:
Can you please fix the same wording issues in the `.env` file too? 🙂

frispete:
> Thank you for the contribution.

With pleasure. I would love to contribute to this valuable project in these times. Just consider this a warm-up. Hopefully getting my [issues](#294) sorted out soon, I'm ready to provide more valuable contributions. An example of integrating with an existing nginx/certbot infrastructure is first and foremost here.

> Can you please fix the same wording issues in the `.env` file too? :slightly_smiling_face:

Given, you had `env.example` in mind, here we go...

While at it, I'm inclined to remove the trailing full stops from `env.example`, but that would produce some churn. Opinions?

sapkra:
> While at it, I'm inclined to remove the trailing full stops from env.example, but that would produce some churn. Opinions?

I also think they are quite annoying, but @saghul has to decide that.

saghul:
Thanks @frispete ! Regarding full-stops, I like docs to be gramatiaclly correct! :-)

sapkra:
@saghul But these are no sentences, but rather headings, which grammatically have no full-stops.

frispete:
Thanks for the merge, @saghul. 

While I really support your request of being grammatically correct, it's rather disturbing for casual readers due to attempting to make fragments of sentences **look** correct, while they are not.

PR #323, opened 2020-03-29
examples: add example to run with docker stack

### LINKED ISSUE #306: change the network name for docker stack to work
To achieve a more extensible deployment but without the fuzz of a kubernetes one I want to deploy using `docker stack`.

Docker itself doesnt source this `.env` file, nor has a param like `--env-file` to load those variables, so I pre-process the `docker-compose.yml` file with
```bash
docker-compose config > deployment.yml
```

Then, knowing that docker complains about the network name I remove the dot
```bash
sed -i 's/meet\.jitsi/meet_jitsi/g' deployment.yml
```

Afther this I simply deploy the stack
```bash
docker stack deploy --compose-file deployment.yml jitsi
```

And voila :)

Maybe this little trick of using docker-compose for interpolating is known to anyone arround, but it just took me a couple of hours to figure out why `docker stack` was eating out those env variables.

Should this be merged in the docs or the examples?

Also, can we change the network name to something `/^[a-zA-Z-_]+$/` compliant?

dinigo:
resolves #306

### DIFF
--- a/examples/docker-stack/README.md
+++ b/examples/docker-stack/README.md
@@ -1,0 +1,24 @@
+# DOCKER STACK
+
+In order to make jitsi docker-compose config  work with docker stack  we have to take in consideration several things
+1. docker doesnt handle dots (`.`) in the network names due to the internal DNS
+2. docker-compose interpolates the configuration set in the `.env` file while docker itself doesn't
+
+## 1. Fixing the network
+
+If you examine the main `docker-compose.yml` file, yo will realize that you can set a name for docker to handle and 
+another (alias) for docker to expose. Whith this _alias_ trick we can make XMPP support our network 'domain'
+
+## 2. Interpolating the configuration
+
+Start with the commands as stated in the main README.md. Simply copy the example config into your personal `.env`
+file and customize it. Once you have it ready generate the deployment with all the interpolated values with
+```
+docker-compose config > deployment.yml
+```
+If you examine the `deployment.yml` you will realize all your variables are set already
+
+Now you can start your docker stack like you would normaly do
+```
+docker stack deploy --compose-file deployment.yml jitsi
+```
--- a/examples/docker-stack/docker-compose.yml
+++ b/examples/docker-stack/docker-compose.yml
@@ -1,0 +1,166 @@
+version: '3'
+
+services:
+    # Frontend
+    web:
+        image: jitsi/web
+        ports:
+            - '${HTTP_PORT}:80'
+            - '${HTTPS_PORT}:443'
+        volumes:
+            - ${CONFIG}/web:/config
+            - ${CONFIG}/web/letsencrypt:/etc/letsencrypt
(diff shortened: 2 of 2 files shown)


dinigo:
```
ID                  NAME                IMAGE                  NODE                DESIRED STATE       CURRENT STATE         ERROR               PORTS
g6dkxb1bietu        jitsi_web.1         jitsi/web:latest       swarm               Running             Running 3 hours ago
zi4icxe1gcmi        jitsi_prosody.1     jitsi/prosody:latest   swarm               Running             Running 3 hours ago
r28c3qm7u4dn        jitsi_jvb.1         jitsi/jvb:latest       swarm               Running             Running 3 hours ago
s0k82gql7xwt        jitsi_jicofo.1      jitsi/jicofo:latest    swarm               Running             Running 3 hours ago
```

Just had a chat with a friend using it

saghul:
Why ddo we need to duplicate the docker-compose.yml file? How about providing a bash script wich creates the required yaml file for Docker Stack?

dinigo:
righ

port22:
Did anyone have success here, yet? I'm pulling my hair off trying to get it working on swarm. 

What are the Hostnames in the config for swarm? Like, those:

```
      XMPP_AUTH_DOMAIN: auth.meet.jitsi
      XMPP_DOMAIN: meet.jitsi
      XMPP_SERVER: xmpp.meet..jitsi
      XMPP_GUEST_DOMAIN: guest.meet.jitsi
      XMPP_INTERNAL_MUC_DOMAIN: internal-muc.meet.jitsi
      XMPP_RECORDER_DOMAIN: recorder.meet.jitsi
```

I don't get that thing with the aliases and the domain suffixes. Why aren't they just talking by service hostname? This works with swarm and docker-compose.

nginx wasn't the best choice either. It's not able to do dynamic dns lookups (at least not like you configured it without `resolver`). If any container restarts, upstream is dead. 
Also it makes no sense that Etherpad *MUST* be running when nginx starts, just because nginx too dumb to ignore. In my opinion, Caddy would've been a better choice. More even, to be able to connect the stack with an already running traefik, like I have.

side note:
```
set -a
source .env
set +a
```
doing that, variable expansion will work with `docker stack deploy`.

dinigo:
yes. there's a pull request open with the method. basically you alias the shared network when needed

tu-nv:
The PR does not work for me. I see the error `xmpp.meet.jitsi: Name or service not known` in` jvb`. 

Also tried to not using network alias but using the service name (i.e.: set `XMPP_SERVER=prosody`), and this DOES work with `docker-compose up -d`, but does not work with swarm (error: `prosody: Name or service not known`).  I tried `ping prosody` from other container and it works.

The docker overlay network just bug me everytime I need it :disappointed:

hamletmun:
What about using the overlay network?
```
networks:
    meet_jitsi:
	driver: overlay
        attachable: true
```

PR #332, opened 2020-03-30
Maximize S/N ratio of env.example

frispete:
Here's a PR as being discussed. 

@saghul You might want to have a second look, please.

Advantages:
 * comments are consistent across the whole file now
 * easier to read/comprehend
 * spares 62 bytes :wink:

Disadvantages:
 * some churn for everybody (who diffs .env with this file, e.g. me)

I've documented the antidote for the latter issue in the commit message.
There's a chance, that people, who diff things, also read git logs. :smiley:

### DIFF
--- a/env.example
+++ b/env.example
@@ -2,23 +2,23 @@
 # Basic configuration options
 #
 
-# Directory where all configuration will be stored.
+# Directory where all configuration will be stored
 CONFIG=~/.jitsi-meet-cfg
 
-# Exposed HTTP port.
+# Exposed HTTP port
 HTTP_PORT=8000
 
-# Exposed HTTPS port.
+# Exposed HTTPS port
 HTTPS_PORT=8443
 
-# System time zone.
+# System time zone
 TZ=Europe/Amsterdam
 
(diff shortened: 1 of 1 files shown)


saghul:
This is why I started adding periods:

```
- # A comma separated list of APIs to enable when the JVB is started. The default is none.
+ # A comma separated list of APIs to enable when the JVB is started. The default is none
```

I agree that when there is only one sentence it makes sense to drp it, but it doesn't when there is 2. So I added it everywhere for consistency and to keep my "OCD" at bay.

frispete:
I would express it this way:
```
# A comma separated list of APIs to enable when the JVB is started [default: none]
```

sapkra:
Wouldn't it make sense to remove the comments completely from the `.env` file? In my opinion, we don't need the docs for the variables in two places.

frispete:
I think, they're useful, since it avoids switching between two documents, while mastering a pretty sophisticated setup (it still fails for me for some constellations, e.g. where the Android app is involved). It can and should be brief, of course.

PR #341, opened 2020-03-31
etherpad: use official image and making skin full width

sapkra:
depends on:
* https://github.com/ether/etherpad-lite/pull/3765
* https://github.com/ether/etherpad-lite/pull/3873

@saghul Should I also add the etherpad env vars to the README? I don't think it would makes sense to basically copy the docs of etherpad.


flecno:
I think it is enough and better to link to the README from etherpad.

sapkra:
@saghul All my three etherpad PRs are merged now.

luixxiul:
This needs rebase.

etherpad.yml
```diff
@@ -3,8 +3,11 @@ version: '3'
 services:
     # Etherpad: real-time collaborative document editing
     etherpad:
-        image: jitsi/etherpad:latest
+        image: etherpad/etherpad
```
saghul, line 6: OK, let's do this. Can you please pin the image? LGTM otherwise.

saghul (commented):
Bonus nachos is you make the skin configurable while we are at it :-)

sapkra:
@saghul The skin is configurable by default or do you mean that I should remap the env vars? If so, what should be the default value for both `SKIN_NAME` and `SKIN_VARIANTS`? Shall we use `SKIN_VARIANTS=full-width-editor` as the default?

saghul:
> @saghul The skin is configurable by default or do you mean that I should remap the env vars? If so, what should be the default value for both `SKIN_NAME` and `SKIN_VARIANTS`? Shall we use `SKIN_VARIANTS=full-width-editor` as the default?

Sorry I missed your reply. Yes please!

sapkra:
@saghul Done.

saghul (approved):

Issue #310, opened 2020-03-27
Kubernetes example - jvb doesn't work 


lelvisl:
Hi.
I tried to deploy yours Kubernetes example, vide calls only works in p2p mode.
When add third member (and trying to switch to jvb - everybody sees himself. 
When I disable p2p in client (or in browser) - same. 
[jvb.log](https://github.com/jitsi/docker-jitsi-meet/files/4395740/jvb.log)

d1ngd0:
I am seeing this error repeatedly in the logs
```
Error: Could not find or load main class the
```

lelvisl:
how i can fix this?

obeleh:
same issue here :(

obeleh:
@dimm0 could you have a look?

dimm0:
Either way I have no idea how to debug this. I know nothing about jitsi internals

Schnitzel:
I struggled with the same, the problem is here:
https://github.com/jitsi/docker-jitsi-meet/blob/ebfa142d384466fefc7cd224cdca50df9cfe40c7/examples/kubernetes/deployment.yaml#L112-L113
this confuses the hell out of java and therefore it doesn't start
removing it will allow jvb to start.

dimm0:
You mean you didn't change the value?

Schnitzel:
yes, I didn't change it and so the jvb tried to start with 

`JAVA_SYS_PROPS="-Dnet.java.sip.communicator.SC_HOME_DIR_LOCATION=/ -Dnet.java.sip.communicator.SC_HOME_DIR_NAME=config -Djava.util.logging.config.file=/config/logging.properties -Dorg.ice4j.ice.harvest.NAT_HARVESTER_LOCAL_ADDRESS=10.1.16.155 -Dorg.ice4j.ice.harvest.NAT_HARVESTER_PUBLIC_ADDRESS=<Set the address for any node in the cluster here>" exec /usr/share/jitsi-videobridge/jvb.sh --apis=none`

which caused all kine of issues with Java, if I set `DOCKER_HOST_ADDRESS` to the IP of the Loadbalancer that is loadbalancing between all the Nodeports, it works perfectly.

dimm0:
Ah, I see. I'll add to docs.

dimm0:
I added it to my pending PR https://github.com/jitsi/docker-jitsi-meet/pull/349

saghul:
[...]

PR #349, opened 2020-03-31
Specify namespace for secret

dimm0:


obeleh:
I've tested this PR. It solves an issue I had fixes #310

PR #360, opened 2020-04-01
docs: fix grammar and typos

sapkra:
@saghul  I've used grammarly to find and fix all typos so that we don't need a PR for each one anymore.


saghul:
Thanks a lot!

Issue #304, opened 2020-03-26
org.jitsi.meet.ComponentMain.call().323 host-unknown, host:xmpp.meet.jitsi, port:5347


prodigy7:
Hi there!

I try to get jitsi working with docker but currently I have the behaviour, that when someone joined a open session, the sessions stops and I got the message "Unfortunately, something went wrong".

Now, looking into docker log I see following lines:
```
jicofo_1   | Mar 26, 2020 8:24:00 PM org.jitsi.util.LoggerImpl log
jicofo_1   | INFO: Owner has left the room !
jicofo_1   | Mar 26, 2020 8:24:00 PM org.jitsi.util.LoggerImpl log
jicofo_1   | INFO: Member holyshit@muc.meet.jitsi/1c8a6a53 is leaving
jicofo_1   | Mar 26, 2020 8:24:00 PM org.jitsi.util.LoggerImpl log
jicofo_1   | INFO: Removed participant: true, holyshit@muc.meet.jitsi/1c8a6a53
jicofo_1   | Mar 26, 2020 8:24:00 PM org.jitsi.util.LoggerImpl log
jicofo_1   | INFO: Disposed conference for room: holyshit@muc.meet.jitsi conference count: 0
prosody_1  | bosh00ec30a8-3853-4951-b2a9-bfbc6217e250  info     BOSH client disconnected
web_1      | 192.168.192.1 - - [26/Mar/2020:20:24:00 +0100] "POST /http-bind?room=holyshit HTTP/1.1" 200 309 "https://meet.my-domain.de/holyshit" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.149 Safari/537.36"
web_1      | 192.168.192.1 - - [26/Mar/2020:20:24:00 +0100] "POST /http-bind?room=holyshit HTTP/1.1" 200 116 "https://meet.my-domain.de/holyshit" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.149 Safari/537.36"
prosody_1  | jitsi-videobridge.meet.jitsi:component    warn     Component not connected, bouncing error for: <iq id='Ar2Um-199' type='get' to='jitsi-videobridge.meet.jitsi' from='focus@auth.meet.jitsi/focus2581782283962450'>
[...]

PR #362, opened 2020-04-02
Added related but missing part to quickstart

prodigy7:
As suggested in #304 documentation supplemented by missing part.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ follow these steps:
 * Create a ``.env`` file by copying and adjusting ``env.example``, and create required `CONFIG` directories
   * `cp env.example .env`
   * `mkdir -p ~/.jitsi-meet-cfg/{web/letsencrypt,transcripts,prosody,jicofo,jvb}`
+* Build with ``make clean && make``.
 * Run ``docker-compose up -d``.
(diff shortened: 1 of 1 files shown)


saghul:
That is not necessary. Only if you want to build youir own images, which not everyone needs to. I'd be happy to have a section with "Build your own images" as a title though.

prodigy7:
I followed first your quickstart, post my logs and everytime you said to me, that my versions are outdated. I pulled again and again but first when you told me, I'd to call "make clean && make" I got a current version. So, whats wrong? ;-)

sgumirov:
I agree with Sagul, no need for "make". The current quickstart is actually correct, works just fine.

prodigy7:
Okay. But then tell me, why I had a outdated Installation on my System where I had never fetched jitsi docker before? What was wrong? Only after execute make, all was on a current Version.

sgumirov:
@prodigy7 Most probably dockerhub repo image is not updated instantly as currently no version tag is used to fetch specific container image version. 
With running `make build` you effectively build image locally and local docker repo has higher priority over dockerhub, so the newer version is used instead seamlessly later on.

PR #374, opened 2020-04-04
Update Traefik v2 example with UDP

calvinbui:
Traefik supports UDP as of v2.2

Docs: https://docs.traefik.io/routing/providers/docker/#udp


geryepes:
+1

Issue #376, opened 2020-04-04
Set Docker restart policy


dbrgn:
Would it make sense to add a [restart policy](https://docs.docker.com/compose/compose-file/#restart_policy) to the docker-compose file, so that the containers are automatically restarted if they fail (and on system start)?

For example, something like this:

      restart_policy:
        condition: any
        delay: 5s
        max_attempts: 6

robvella:
Agree with this. I manually did docker update --restart=always on every container in the mean time.

sapkra:
See #377

saghul:
Let’s continue in 377

PR #377, opened 2020-04-04
Add restart option in docker-compose file

MathiasRenner:
After a server reboot or in case of an error, the Docker containers of Jitsi are not automatically restarted.

This PR starts all Jitsi service containers automatically in all conditions, i. e. for instance after a server reboot or after error 1 due to some failure. 
Thus, this makes the setup a bit more robust.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -37,7 +37,7 @@ services:
             - JIBRI_RECORDER_USER
             - JIBRI_RECORDER_PASSWORD
             - ENABLE_RECORDING
-        restart: always
+        restart: unless-stopped
         networks:
             meet.jitsi:
                 aliases:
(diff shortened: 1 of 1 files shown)


saghul:
@sapkra Thoughts?

sapkra:
I think this makes sense. But I'm thinking about if `unless-stopped` would be even better.

Docs:
https://docs.docker.com/config/containers/start-containers-automatically/

MathiasRenner:
@sapkra: Nice catch, I agree. Will update the PR soon.

dbrgn:
Fixes #376.

MathiasRenner:
@sapkra Updated now. Please merge.

saghul:
Thanks! Can you please also add it to the etherpad, jigasi and jibri yaml files?

saghul:
Superseded by https://github.com/jitsi/docker-jitsi-meet/pull/417 - thanks for getting the efort started though!

PR #391, opened 2020-04-06
Centralizing Etherpad Host Configuration

arthurd2:
Fixing some issues due the hard coded name of the EtherPad host written on the YAML.


saghul:
There are now 2 ways to configure it, with an external URL or the builtin one.

PR #407, opened 2020-04-08
Clean up gen-passwords.sh shell code

brother:
The shell script for generating the passwords uses old style backticks for executing the commands and expects the .env file to be placed in the current working directory.

Converting the backticks to $() style execution for better readability and also passing the script through dirname to enforce the placement of the .env to the CWD of the script.

### DIFF
--- a/gen-passwords.sh
+++ b/gen-passwords.sh
@@ -4,12 +4,12 @@ function generatePassword() {
     openssl rand -hex 16
 }
 
-JICOFO_COMPONENT_SECRET=`generatePassword`
-JICOFO_AUTH_PASSWORD=`generatePassword`
-JVB_AUTH_PASSWORD=`generatePassword`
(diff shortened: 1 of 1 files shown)


brother:
The .env file thing could possibly be better of as a flag to the script. if omitted use the `dirname "$0"` approach as default.

gen-passwords.sh
```diff
@@ -4,12 +4,12 @@ function generatePassword() {
     openssl rand -hex 16
```
sgumirov, line 4: It's possible to use base64 encoding to generate shorter passwords with the same entropy amount with -base64 instead of -hex.
```suggestion
    openssl rand -base64 16
```

sgumirov: Could be seen as offtopic, however and maybe I should open new PR.

PR #409, opened 2020-04-08
Fix WASM MIME type

JakubOnderka:
MIME type application/wasm is required by WebAssembly specification, but this MIME type is not part of of nginx default mime.types file.

This change should remove warning from logs at least for Firefox:

```
wasm streaming compile failed: TypeError: Response has unsupported MIME type index.js:8:9333
falling back to ArrayBuffer instantiation index.js:8:9375
```

I didn't check other browsers.

### DIFF
--- a/web/rootfs/defaults/nginx.conf
+++ b/web/rootfs/defaults/nginx.conf
@@ -26,7 +26,11 @@ http {
 
 	client_max_body_size 0;
 
-	include /etc/nginx/mime.types;
+ 	include /etc/nginx/mime.types;
+	types {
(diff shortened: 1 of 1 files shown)


saghul:
@aaronkvanmeerten WDYT?

JakubOnderka:
Same error also happen in latest Chrome (81.0.4044.113):
```
index.js:8 wasm streaming compile failed: TypeError: Failed to execute 'compile' on 'WebAssembly': Incorrect response MIME type. Expected 'application/wasm'.
index.js:8 falling back to ArrayBuffer instantiation
```

Issue #339, opened 2020-03-31
JIGASI SIP join connection fails with IndexOutOfBoundsException


albe:
```
jigasi_1   | Jigasi 2020-03-30 17:58:29.504 INFO: [104] org.jitsi.jigasi.SipGateway.incomingCallReceived().188 Incoming call received...
jigasi_1   | Jigasi 2020-03-30 17:58:30.509 INFO: [105] org.jitsi.jigasi.SipGatewaySession.run().1295 [ctx=15855839095041960505721]Using default JVB room name property inperson
jigasi_1   | Jigasi 2020-03-30 17:58:30.522 INFO: [105] org.jitsi.jigasi.JvbConference.setXmppProvider().539 [ctx=15855839095041960505721] Using ProtocolProviderServiceJabberImpl(Jabber:53ac09bc@meet.jitsi/53ac09bc)
jigasi_1   | Jigasi 2020-03-30 17:58:30.525 SEVERE: [21] impl.packetlogging.PacketLoggingServiceImpl.run().845 Error writing packet to file
jigasi_1   | java.lang.SecurityException: Insufficient rights to access this file in current user's home directory: /config/log/jitsi0.pcap
jigasi_1   |    at org.jitsi.impl.fileaccess.FileAccessServiceImpl.getPrivatePersistentFile(FileAccessServiceImpl.java:171)
jigasi_1   |    at net.java.sip.communicator.impl.packetlogging.PacketLoggingServiceImpl.getFileNames(PacketLoggingServiceImpl.java:197)
jigasi_1   |    at net.java.sip.communicator.impl.packetlogging.PacketLoggingServiceImpl.savePacket(PacketLoggingServiceImpl.java:582)
jigasi_1   |    at net.java.sip.communicator.impl.packetlogging.PacketLoggingServiceImpl.access$100(PacketLoggingServiceImpl.java:35)
jigasi_1   |    at net.java.sip.communicator.impl.packetlogging.PacketLoggingServiceImpl$SaverThread.run(PacketLoggingServiceImpl.java:834)
jigasi_1   | Jigasi 2020-03-30 17:58:30.526 SEVERE: [21] impl.packetlogging.PacketLoggingServiceImpl.run().845 Error writing packet to file
jigasi_1   | java.lang.SecurityException: Insufficient rights to access this file in current user's home directory: /config/log/jitsi0.pcap
[...]

PR #411, opened 2020-04-08
Jigasi: Add configuration option to skip reinvite on focus change (fi…

### LINKED ISSUE #393: Jigasi interoperability problems with FreeSWITCH
In order to be able to interface Jigasi with FreeSWITCH for dial-in, setting the option 

```
net.java.sip.communicator.impl.protocol.sip.SKIP_REINVITE_ON_FOCUS_CHANGE_PROP=true
```

is required in order to get dial-in working (cf. https://community.jitsi.org/t/recently-updated-jigasi-and-jvb-jitsi-meet-incoming-calls-no-longer-get-answered/21182/24).

This is currently not possible with this Docker image.

It would be great if you could add support for this config switch.

ginkel:
…xes a compatibility issue with inbound calls via FreeSWITCH)

Fixes #393


damencho:
This is already part of the code: https://github.com/jitsi/jigasi/commit/1658e409f977fde3321de18284340c15dfc71a82

ginkel:
Interesting. I couldn't get it to work without manually setting the property to `true` in Jigasi's `sip-communicator.properties`.

damencho:
I think maybe we havn't pushed jigasi to stable for a while and the docker is behind on versions.

albe:
See also https://github.com/jitsi/docker-jitsi-meet/issues/339#issuecomment-609403144

Is a push to stable and update docker in the pipeline or anything blocking that?
I guess you're probably overwhelmed with new issue reports all over the places, which challenges your well known development workflow.

saghul:
stable-4416 was just released. Please give it a test and holler if it doesn't work as expected.

ginkel:
ACK, stable-4416 works like a charm w/o supplying the config option. Thank you!

Issue #124, opened 2019-07-25
How to configure to your own turnservers?


whiranpat:
I have setup some coturn servers using instrumento/coturn but when I set the JVB_STUN_SERVERS to this address in the .env it still showing in the curl that it is using google stun servers

whiranpat:
how to do you test to see if it is connecting to the correct STUN servers? When I curl the jitsi site I still see this in the return message 
`  stunServers: [
            { urls: 'stun:stun.l.google.com:19302' },
            { urls: 'stun:stun1.l.google.com:19302' },
            { urls: 'stun:stun2.l.google.com:19302' }
        ],`

pansj2020:
Have you rm $config . You need rm $config after you modify .env  and redockercompse up. Otherwise it won't effect.

AntonDzyk:
Problem on this line
https://github.com/jitsi/docker-jitsi-meet/blob/master/web/rootfs/etc/cont-init.d/10-config#L78
{code}
cp /defaults/config.js /config/config.js

PR #414, opened 2020-04-08
Add option for custom P2P STUN servers

klonfish:
This adds an configuration option `STUN_SERVERS` to replace the predefined STUN servers for P2P connections with custom ones in `config.js`. The code is not really pretty but it seems to work with only one as well as with multiple entries.

The `JVB_STUN_SERVERS` variable maybe can be dropped and replaced by the new one but I'm not sure about that. Maybe there are setups where different servers for the clients and the videobridge are desirable.

The changes at least partly solve #124.


README.md
```diff
@@ -432,6 +432,7 @@ Variable | Description | Default value
 `JIGASI_TRANSCRIBER_ADVERTISE_URL` | Jigasi will post an URL to the chat with transcription file | true
 `DISABLE_HTTPS` | Handle TLS connections outside of this setup | 1
 `ENABLE_HTTP_REDIRECT` | Redirect HTTP traffic to HTTPS (necessary for Let's Encrypt) | 1
+`STUN_SERVERS` | STUN servers used by web clients for P2P connections | meet-jit-si-turnrelay.jitsi.net:443
```
albe, line 435: Wouldn't it make more sense to merge this with `JVB_STUN_SERVERS`?

klonfish: Indeed, I thought about that, too. However, as already stated in the initial comment, I was unsure if it would make sense to keep these two options separate for some setups and wanted to get some feedback on that.

luixxiul: This change need to be applied to https://github.com/jitsi/handbook/blob/master/docs/devops-guide/docker.md (cf: #601)

PR #415, opened 2020-04-08
feat: add helm chat and gcsfuse for jibri

rodrigopavezi:
I have created a helm chart for deploying all jitsi services including jibri into kubernetes.

I have tested it on GKE and works greatly.

I have also added gcsfuse to jibri docker image and configure it to sync recordings with GCS. I have also tested and works greatly.

Let me know if you have any question.

Cheers

### DIFF
--- a/examples/helm/README.md
+++ b/examples/helm/README.md
@@ -1,0 +1,65 @@
+# Install guide for kubernetes with helm
+
+This guide will deploy jitsi with jibri on kubernetes
+
+
+## For each node in the cluster do the following:
+
+### Install generic kernel image
+```
+sudo apt-get update
+sudo apt-get install linux-image-generic
+```
+#### Change grub to start node VM with generic kernel
+
+```
+# GKE Node
+sudo vim /etc/default/grub.d/50-cloudimg-settings.cfg
+
+# Other Setup
+sudo vim /etc/default/grub
+```
+
+Replace #GRUB_DEFAULT=0 with
+```
+# Make the latest generic kernel the default
+release=$(linux-version list | grep -e '-generic$' | sort -V | tail -n1)
+GRUB_DEFAULT="Advanced options for Ubuntu>Ubuntu, with Linux $release"
+```
+```
+sudo update-grub
+```
+#### Reboot VM
+```
+sudo reboot
+```
+
+### Setup virtual sound device in the node
+#### install the module
+```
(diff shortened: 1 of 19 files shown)


cjk:
This is cool, hope it gets merged!

ChrisTomAlx:
Hey @rodrigopavezi 

Couple of questions. Would be eternally great-full if you could share your thoughts. I don't see anyone else who has accomplished jibri in k8s functionality well, so I am asking this here.

1. One Jibri Pod can only run one recording at a time right? 
2. Can two or more jibri pods run on the same node simultaneously recording two or more different conferences that are happening at the same time

Cheers and thanks in advance !!

ChrisTomAlx:
If you don't add "/dev/shm" similar to "/dev/snd" as a volume when you do screen share the recording can crash. Just dropping this here for anyone else

kpeiruza:
> Hey @rodrigopavezi
> 
> Couple of questions. Would be eternally great-full if you could share your thoughts. I don't see anyone else who has accomplished jibri in k8s functionality well, so I am asking this here.
> 
> 1. One Jibri Pod can only run one recording at a time right?
> 2. Can two or more jibri pods run on the same node simultaneously recording two or more different conferences that are happening at the same time
> 
> How does one handle scaling in this situation ?
> 
> Cheers and thanks in advance !!

Hi, we have been running jibri on K8s for 2 months.

More than 1 jibri can be running on the same node if you're using Jibri with Pulseaudio. We have some local clusters with that strategy.

Then, for autoscaling what we do is use a simple Kubernetes HPA, thus, using 1 node per Jibri in any provider with autoscaling node pools.

Regards!

ChrisTomAlx:
@kpeiruza Jibri with PulseAudio seemed to have certain issues hence I never went that route. Jitsi if my understanding is correct used to use PulseAudio then shifted to ALSA because of the same reason. Do you have the container image that uses PulseAudio hosted somewhere publicly so I can take it out for a spin ?

vkruoso:
@kpeiruza This is exactly what I am doing right now. Can you share extra information on how we can have it running ourselves? Like the modifications needed on the jibri image and the k8s objects.

sapkra:
@rodrigopavezi Can you create a PR at our new helm repository? https://github.com/jitsi-contrib/helm

PR #417, opened 2020-04-09
Add service restart policy, doc, defaults

rcarmo:
I found it rather strange that my deployment failed upon rebooting the host... This fixes it, of course.

### DIFF
--- a/env.example
+++ b/env.example
@@ -320,4 +320,4 @@ JIBRI_LOGS_DIR=/config/logs
 
 # Container restart policy
 # Defaults to always-stopped
-RESTART_POLICY=always-stopped
+RESTART_POLICY=unless-stopped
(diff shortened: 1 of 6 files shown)


c-goes:
I would prefer `unless-stopped`, still it's a step in the right direction.

sapkra:
Duplicate of https://github.com/jitsi/docker-jitsi-meet/pull/377

rcarmo:
Here you go, all done. Kindly merge if you agree.

env.example
```diff
@@ -317,3 +317,7 @@ JIBRI_LOGS_DIR=/config/logs
 # Redirect HTTP traffic to HTTPS
 # Necessary for Let's Encrypt, relies on standard HTTPS port (443)
 #ENABLE_HTTP_REDIRECT=1
+
+# Container restart policy
+# Defaults to always-stopped
```
albe, line 322:
```suggestion
# Defaults to unless-stopped
```

PR #440, opened 2020-04-12
update to: debian:buster-slim and jre-11

snoopotic:
Hi,
with this PR I successfully updated the base image(s) to use buster - the new stable Debian as base. 
Only one pitfall for prosody had to be included: to build 'luajwtjitsi' it needs 'luacrypto' that does not build with buster's 'libssl-dev' (obviously some deprecated/obsolete libraries vanished). Thus you need to "downgrade" to libssl1.0 that is only available on stretch at least. luckily this libraries do not cross-fail other things and the build is a success. Well, in near future luacrypto at least needs an update (4 years without update... uuuh), or luajwtjitsi must become independent from luacrypto...

As well I updated lua-cjson to most actual minor-update. 

and after doing this I successfully could use openjdk-11-jre-headless, what was my initial goal - but stretch did not offer.
This should give the docker-users a little performance push ;)

oh, and don't forget:
```/etc/sysctl.d/jitsi.conf
net.core.rmem_max=10485760
net.core.netdev_max_backlog=100000
```
that will boost your experience as well, the installer-users just got it docker-users don't. (just little PR service. :P)


jkroepke:
@snoopotic 

Instead install openssl 1.0.0 you could use a fork which is compatible with openssl 1.1.x

Fork: https://github.com/evanlabs/luacrypto

See: https://github.com/jitsi/jitsi-meet/issues/2029

saghul:
Oh I wasn’t aware of that fork! That’s pretty awesome, yeah we should use it in the interim, and hopefully get rid of that dependency sooner rather than later.

snoopotic:
> @snoopotic
> 
> Instead install openssl 1.0.0 you could use a fork which is compatible with openssl 1.1.x
> 
> Fork: https://github.com/evanlabs/luacrypto
> 
> See: [jitsi/jitsi-meet#2029](https://github.com/jitsi/jitsi-meet/issues/2029)

Thanks. I tried a build locally. But it's quite harder, as that forked luacrypto only builds with lua51 :/
I'd still stick with my workaround. Hehe.

sapkra:
@snoopotic What about this fork https://github.com/Wassasin/luacrypto? It's a fork of the fork and has a change called `Remove hard dependency on Lua51 and instead use more general Lua`.

saghul:
Great point @sapkra im fact we need it to work with Lua 5.2.

snoopotic:
Thanks for your feedback guys. Sadly I had to do a fork as well... to fix a small gap. But I am waiting for their merges so we can possibly go back to the evanlabs one.
If you like to test, try this. out. Looks good ;)

saghul:
Folks, I'm sorry, but this will need to wait: https://github.com/jitsi/docker-jitsi-meet/pull/227#issuecomment-624486062

I'm closing this because I don't see it being merged in the sort term. We do keep the buster issue open. We'll reopen if need be (I hope we have gotten rid of luacrypto by then).

toby63:
@saghul Couldn't you use AdoptOpenJDK in the meantime, just like you already suggested for the handbook?
Because Debian 9 is now EOL, and it would be good to update your Docker repo.

I think that Docker should be able to download and unpack it, with a link like this:
`https://github.com/AdoptOpenJDK/openjdk${_java_ver}-binaries/releases/download/${_repo_ver}/OpenJDK${_java_ver}U-jdk_x64_linux_hotspot_${_targz_ver}.tar.gz`

Example: [Archlinux package in AUR](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=java8-adoptopenjdk)

Issue #433, opened 2020-04-11
Consider not creating dhparameters on the fly


hannob:
Using the docker-jitsi-meet installation I observed that a recurring issue is that startup takes quite a while, which I believe is in part due to the dhparam generation, also indicated e.g. in issue #123 .

With some background in TLS security I'd like to say that I believe this has almost no security value and is annoying to users and probably the source of quite a bit of confusion and installation difficulties.

I propose that you either use static dh parameters or just disable DHE support altogether.

Using per-host security parameters has a theoretical security advantage, because there are attacks that are faster if you attack several connections with the same parameter, but with long (>=2048) parameters it really doesn't matter and is merely a theoretical issue.

But more than that: Modern browsers don't use DHE, and Jitsi Meet doesn't work with ancient browsers anyway. While TLS 1.3 in theory has DHE support, it is not using the custom parameters (the parameters are fixed in the protocol). But anyway: Everyone's using ECDHE these days.

saghul:
Sounds reasonable. Do you know of any endpoint that would not work if DHE upport is removed? Maybe some old Android 5 devices? If not, we could ship a pre-generated dhparams, and only generate one if absent, thus allowing people to customize the images eaasily. What do you think?

hannob:
If you look at the handshake simulation
https://www.ssllabs.com/ssltest/analyze.html?d=meet.golem.de
barely anything will use DHE. The only ones that show up are Internet Explorer versions on unsupported Windows versions.

(Unrelated question, but as I'm seeing this is there a reason the docker installation doesn't do TLS 1.3?)

saghul:
Thanks for that! Any chance you could send a PR removing it then?



> (Unrelated question, but as I'm seeing this is there a reason the docker installation doesn't do TLS 1.3?)

Nope, I guess we forgot to update the nginx config, oops! (If you can add that to your PR, it would be great too!)

hannob:
The nginx image is using debian stretch as the base image, which does not support TLS 1.3. So I guess no TLS 1.3 for now.

saghul:
Thanks @hannob , hopefully we can update to Buster soon! I'm keeping this open until then.

saghul:
Added TLS 1.3 support in b0cb4a1

PR #450, opened 2020-04-14
Remove Diffie Hellman support and parameter generation, only needed b…

hannob:
…y very old browsers. This should increase startup speed, as parameter generation is slow.

See also issue #433


saghul:
Thank you!

PR #462, opened 2020-04-15
add Let's Encrypt Cloudflare DNS Support

muhviehstah:
PR will add Let's encrypt support for Cloudflare DNS


muhviehstah:
please note CF  has some service problems today https://www.cloudflarestatus.com/

saghul:
TBH I'm not sure we want to support this. This setup tries to be as lean as possible, and we encourage others with setups we don't support to use a companion container and proxy to the HTTP port. Just like those with pre-existing SSL certificates, which we also don't support.

README.md
```diff
@@ -487,3 +496,4 @@ You are now able to run `docker-compose up` as usual.
 [jwt.io]: https://jwt.io/#debugger-io
 [Etherpad]: https://github.com/ether/etherpad-lite
 [Jibri]: https://github.com/jitsi/jibri
+[latest release]: https://github.com/jitsi/docker-jitsi-meet/releases/latest
```
luixxiul, line 499: These change on README.md need to be applied to https://github.com/jitsi/handbook/blob/master/docs/devops-guide/docker.md (cf: #601)

saghul:
After thinking about it some more, this is not something I'd like to support here. It can be easily accomplished by another container fronting this setup.

PR #498, opened 2020-04-21
Set security headers also for non HTTPS

### LINKED ISSUE #493: Security headers are not set for HTTP version
For `jitsi/web` container, two security headers:
* `X-Content-Type-Options` and
* `X-XSS-Protection` 

are set just in for ssl version.

https://github.com/jitsi/docker-jitsi-meet/blob/d44230e2b121e72542a60b6dfbce9b07e84ed706/web/rootfs/defaults/ssl.conf#L23-L24

 So if somebody uses HTTPS reverse proxy in front of this container, these headers are not set. They should be moved to https://github.com/jitsi/docker-jitsi-meet/blob/master/web/rootfs/defaults/meet.conf file.

JakubOnderka:
Fixes #493

### DIFF
--- a/web/rootfs/defaults/meet.conf
+++ b/web/rootfs/defaults/meet.conf
@@ -11,6 +11,10 @@ ssi_types application/x-javascript application/javascript;
 index index.html index.htm;
 error_page 404 /static/404.html;
 
+# Security headers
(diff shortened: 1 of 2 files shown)


JakubOnderka:
It will be possible to merge this?

PR #501, opened 2020-04-21
Modify config.js and interface_config.js via environment

jkroepke:
Adjust config.js and interface_config.js via environment variable.

Useful to set some privacy settings if jitsi runs in an environment with ephemeral volumes (like kubernetes).

The users just needs to define the properties they want to change. The user provided settings will be merge via the provided default json.

```javascript
config = Object.assign({}, config, customConfig);
```

Example:
```properties
# See https://github.com/jitsi/jitsi-meet/blob/master/config.js for defaults.
#CONFIG_JSON={disableThirdPartyRequests: true}

# See https://github.com/jitsi/jitsi-meet/blob/master/interface_config.js for defaults.
#INTERFACE_CONFIG_JSON={RECENT_LIST_ENABLED: false}
```


env.example
```diff
@@ -315,6 +315,12 @@ JIBRI_LOGS_DIR=/config/logs
 # Disable HTTPS: handle TLS connections outside of this setup
 #DISABLE_HTTPS=1
 
+# See https://github.com/jitsi/jitsi-meet/blob/master/config.js for defaults.
+#CONFIG_JSON={disableThirdPartyRequests: true}
```
saghul, line 319: This is not JSON but actual JS.

saghul:
I like the simplicity of this, but it won't work beyond top level keys since Object.assign won't properly "merge" the objects.

env.example
jkroepke: Do you have a suggestion?

saghul: You could get the value quoted, as validd JSON and use JSON.parse on it when you assign it.

jkroepke:
Correct, but looking at config.js there aren't much non top-level keys. If a non top level key need to be modified by user, the user has to take the whole object. 

Adding a deep merge function results in more complex code that needs to maintain. 

I could use https://gist.github.com/ahtcx/0cd94e62691f539160b32ecda18af3d6 but I wont this logic to config.js but I can't find a better place.

Instead, I would document it that there is no deep merge + the workaround described above. It just good enough for the first approach. I would create a issue or 2nd pr for the discussion how to implement a deep merge function w/o overload the config.js.

env.example
jkroepke: Ah.. I though about just rename the variable like something without JSON. I don't see the benefit to use `JSON.parse`.

saghul:
Sorry, but I disagree. If are going to do this, we have to do it right. A very common cchange requested is the video resolution, and that's a nested object.

I have an idea we can try: create a small standalone config merger using lodash.merge. We can make it standalone by compiling it to a freestanding binary (minus libc) with QuickJS.

We could make it so it merges all config files given in the command line. This could give us the flexibility to do this: take the default config as is, then create one as part of container boot, then apply the user supplied one. I imagine ssomething like this: `cfgmerge -k config /config/default.js /config/settings.js /config/user.js > /config/config.js`

sapkra:
I've also worked on this topic a while ago. (see #89) Maybe it will be helpful somehow. :D

jkroepke:
I add a new proposal to deep merge the config objects. Depends on https://github.com/jitsi/jitsi-meet/pull/6296

I used the small library https://github.com/TehShrike/deepmerge to archive is. This should be cross browser proof.

I personally like to merge this at javascript / client site because this is way more light then tons of dependencies inside the docker image.

I'm still avoid to use JSON.parse here because of quoting issues. If no quotes are outside, nothing needs to be escaped.

saghul:
Thankss for taking another crack at it!

The problem is that your approach won't work on mobile, because mobile only loads config.js

> I personally like to merge this at javascript / client site because this is way more light then tons of dependencies inside the docker image.

I don't disagree that's why I said this can be done with a standalone binary with no dependencies. It could use the deepmerge library you mentioned.

> I'm still avoid to use JSON.parse here because of quoting issues. If no quotes are outside, nothing needs to be escaped.

Why not use a file instead? Or base64 encoded JSON to avoid quoting issues? Right now the comment says it mst be valid JSON, but it expects JS, which is very confusing.

jkroepke:
> Right now the comment says it mst be valid JSON, but it expects JS, which is very confusing.

I would like to change the name, remove the JSON and document that its must be valid Javascript. If js allowed here, users they want to change sublevel keys, you cloud do it like this:

```properties
CONFIG_JSON={p2p: Object.assign({}, config. p2p, {enabled: false})}
```

but it's maybe too complex. I'm using the JSON.parse now and escape just the `'`.

`CONFIG_JSON={"data": "I'm here"}`

result in

`var customInterfaceConfig = JSON.parse('{"data": "I\'m here"}')`.

It's valid syntax.

> Why not use a file instead?

Avoid more custom code. I don't know the supported browsers. I expect jitsi should work with IE11. So no template literals, no fetch API. Using XHR results in a lot of more code which needs to be maintained.

I always try to avoid own custom code.

> The problem is that your approach won't work on mobile, because mobile only loads config.js

I removed the custom.js code snippet. 

New commit - new review please.

jkroepke:
@saghul Can you review it?

francis-pouatcha:
+1 for this review.

README.md
```diff
@@ -446,6 +446,8 @@ Variable | Description | Default value
 `DISABLE_HTTPS` | Handle TLS connections outside of this setup | 1
 `ENABLE_HTTP_REDIRECT` | Redirect HTTP traffic to HTTPS (necessary for Let's Encrypt) | 1
 `LOG_LEVEL` | Controls which logs are output from prosody and associated modules | info
+`CONFIG_JSON` | Override config properties. Value must be a valid JSON. Will be merged with [defaults](https://github.com/jitsi/jitsi-meet/blob/master/config.js) | none
+`INTERFACE_CONFIG_JSON` | Override interfaceConfig properties. Value must be a valid JSON. Will be merged with [defaults](https://github.com/jitsi/jitsi-meet/blob/master/interface_config.js) | none
```
luixxiul, line 450: These change need to be applied to https://github.com/jitsi/handbook/blob/master/docs/devops-guide/docker.md (cf: #601)

jkroepke:
superseded by https://github.com/jitsi/docker-jitsi-meet/pull/619

PR #502, opened 2020-04-22
Add XMPP WebSocket / Stream Management support

jkroepke:
Alternative to: #181 

Works without manual adjustments for end users.

In config.js I have to set

```
websocket: document.location.protocol.replace('http', 'ws') + '//' + document.location.host + '/xmpp-websocket'
```

because unlike bosh, a relative path would not work because the protocol has to be define.

I also set `proxy_read_timeout` as documented: https://prosody.im/doc/websocket

### DIFF
--- a/env.example
+++ b/env.example
@@ -329,6 +329,9 @@ JIBRI_LOGS_DIR=/config/logs
 # Necessary for Let's Encrypt, relies on standard HTTPS port (443)
 #ENABLE_HTTP_REDIRECT=1
 
+# Enabled XMPP traffic over WebSocket (PUBLIC_URL must be defined!)
+#ENABLE_XMPP_WEBSOCKET=1
+
 # Container restart policy
 # Defaults to unless-stopped
 RESTART_POLICY=unless-stopped
--- a/prosody/Dockerfile
+++ b/prosody/Dockerfile
@@ -48,7 +48,6 @@ RUN \
     && rm -rf /tmp/pkg /var/cache/apt
 
 RUN patch -d /usr/lib/prosody/modules/muc -p0 < /prosody-plugins/muc_owner_allow_kick.patch
-RUN patch -d /usr/lib/prosody/modules/ -p0 < /prosody-plugins/mod_websocket_smacks.patch
 
 COPY rootfs/ /
 
--- a/prosody/Dockerfile
+++ b/prosody/Dockerfile
@@ -48,6 +48,7 @@ RUN \
     && rm -rf /tmp/pkg /var/cache/apt
 
 RUN patch -d /usr/lib/prosody/modules/muc -p0 < /prosody-plugins/muc_owner_allow_kick.patch
+RUN patch -d /usr/lib/prosody/modules/ -p0 < /prosody-plugins/mod_websocket_smacks.patch
 
 COPY rootfs/ /
 
--- a/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
+++ b/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
@@ -15,12 +15,22 @@ http_default_host = "{{ .Env.XMPP_DOMAIN }}"
 {{ $JWT_TOKEN_AUTH_MODULE := .Env.JWT_TOKEN_AUTH_MODULE | default "token_verification" }}
 {{ $ENABLE_LOBBY := .Env.ENABLE_LOBBY | default "0" | toBool }}
 
+{{ $ENABLE_XMPP_WEBSOCKET := .Env.ENABLE_XMPP_WEBSOCKET | default "0" | toBool }}
+{{ $PUBLIC_URL_DOMAIN := .Env.PUBLIC_URL | default "https://localhost:8443" -}}
+
 {{ if and $ENABLE_AUTH (eq $AUTH_TYPE "jwt") .Env.JWT_ACCEPTED_ISSUERS }}
 asap_accepted_issuers = { "{{ join "\",\"" (splitList "," .Env.JWT_ACCEPTED_ISSUERS) }}" }
 {{ end }}
 
 {{ if and $ENABLE_AUTH (eq $AUTH_TYPE "jwt") .Env.JWT_ACCEPTED_AUDIENCES }}
 asap_accepted_audiences = { "{{ join "\",\"" (splitList "," .Env.JWT_ACCEPTED_AUDIENCES) }}" }
+{{ end }}
+
+{{ if $ENABLE_XMPP_WEBSOCKET }}
+-- Deprecated in 0.12
+-- https://github.com/bjc/prosody/commit/26542811eafd9c708a130272d7b7de77b92712de
+cross_domain_websocket = { "{{ $PUBLIC_URL_DOMAIN }}" };
+consider_bosh_secure = true;
 {{ end }}
 
 VirtualHost "{{ .Env.XMPP_DOMAIN }}"
@@ -42,7 +52,15 @@ VirtualHost "{{ .Env.XMPP_DOMAIN }}"
     authentication = "internal_hashed"
   {{ end }}
 {{ else }}
+    -- https://github.com/jitsi/docker-jitsi-meet/pull/502#issuecomment-619146339
+    {{ if $ENABLE_XMPP_WEBSOCKET }}
+    authentication = "token"
+    {{ else }}
     authentication = "anonymous"
+    {{ end }}
+    app_id = ""
+    app_secret = ""
+    allow_empty_token = true
 {{ end }}
     ssl = {
         key = "/config/certs/{{ .Env.XMPP_DOMAIN }}.key";
@@ -50,6 +68,10 @@ VirtualHost "{{ .Env.XMPP_DOMAIN }}"
     }
     modules_enabled = {
         "bosh";
+        {{ if $ENABLE_XMPP_WEBSOCKET }}
+        "websocket";
+        "smacks"; -- XEP-0198: Stream Management
+        {{ end }}
         "pubsub";
         "ping";
         "speakerstats";
@@ -80,7 +102,16 @@ VirtualHost "{{ .Env.XMPP_DOMAIN }}"
 
 {{ if $ENABLE_GUEST_DOMAIN }}
 VirtualHost "{{ .Env.XMPP_GUEST_DOMAIN }}"
+    -- https://github.com/jitsi/docker-jitsi-meet/pull/502#issuecomment-619146339
+    {{ if $ENABLE_XMPP_WEBSOCKET }}
+    authentication = "token"
+    {{ else }}
     authentication = "anonymous"
+    {{ end }}
+    app_id = ""
+    app_secret = ""
+    allow_empty_token = true
+
     c2s_require_encryption = false
 
     {{ if $ENABLE_LOBBY }}
--- a/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
+++ b/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
@@ -16,7 +16,7 @@ http_default_host = "{{ .Env.XMPP_DOMAIN }}"
 {{ $ENABLE_LOBBY := .Env.ENABLE_LOBBY | default "0" | toBool }}
 
 {{ $ENABLE_XMPP_WEBSOCKET := .Env.ENABLE_XMPP_WEBSOCKET | default "0" | toBool }}
-{{ $PUBLIC_URL_DOMAIN := .Env.PUBLIC_URL | default "https://localhost:8443" -}}
+{{ $PUBLIC_URL := .Env.PUBLIC_URL | default "https://localhost:8443" -}}
 
 {{ if and $ENABLE_AUTH (eq $AUTH_TYPE "jwt") .Env.JWT_ACCEPTED_ISSUERS }}
 asap_accepted_issuers = { "{{ join "\",\"" (splitList "," .Env.JWT_ACCEPTED_ISSUERS) }}" }
@@ -29,7 +29,7 @@ asap_accepted_audiences = { "{{ join "\",\"" (splitList "," .Env.JWT_ACCEPTED_AU
 {{ if $ENABLE_XMPP_WEBSOCKET }}
 -- Deprecated in 0.12
 -- https://github.com/bjc/prosody/commit/26542811eafd9c708a130272d7b7de77b92712de
-cross_domain_websocket = { "{{ $PUBLIC_URL_DOMAIN }}" };
+cross_domain_websocket = { "{{ $PUBLIC_URL }}" };
 consider_bosh_secure = true;
 {{ end }}
 
--- a/web/rootfs/defaults/meet.conf
+++ b/web/rootfs/defaults/meet.conf
@@ -63,7 +63,6 @@ location = /xmpp-websocket {
 
     proxy_set_header Host {{ .Env.XMPP_DOMAIN }};
     proxy_set_header X-Forwarded-For $remote_addr;
-    proxy_read_timeout 60s;
     tcp_nodelay on;
 }
 {{ end }}
--- a/web/rootfs/defaults/meet.conf
+++ b/web/rootfs/defaults/meet.conf
@@ -1,3 +1,5 @@
+{{ $ENABLE_XMPP_WEBSOCKET := .Env.ENABLE_XMPP_WEBSOCKET | default "0" | toBool }}
+
 server_name _;
 
 client_max_body_size 0;
@@ -49,6 +51,22 @@ location = /http-bind {
     proxy_set_header X-Forwarded-For $remote_addr;
     proxy_set_header Host {{ .Env.XMPP_DOMAIN }};
 }
+
+{{ if $ENABLE_XMPP_WEBSOCKET }}
+# xmpp websockets
+location = /xmpp-websocket {
+    proxy_pass {{ .Env.XMPP_BOSH_URL_BASE }}/xmpp-websocket;
+    proxy_http_version 1.1;
+
+    proxy_set_header Connection "upgrade";
+    proxy_set_header Upgrade $http_upgrade;
+
+    proxy_set_header Host {{ .Env.XMPP_DOMAIN }};
+    proxy_set_header X-Forwarded-For $remote_addr;
+    proxy_read_timeout 60s;
+    tcp_nodelay on;
+}
+{{ end }}
 
 location ~ ^/([^/?&:'"]+)$ {
     try_files $uri @root_path;
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,6 +15,7 @@ services:
         environment:
             - ENABLE_LETSENCRYPT
             - ENABLE_HTTP_REDIRECT
+            - ENABLE_XMPP_WEBSOCKET
             - DISABLE_HTTPS
             - LETSENCRYPT_DOMAIN
             - LETSENCRYPT_EMAIL
@@ -110,6 +111,7 @@ services:
             - ENABLE_AUTH
             - ENABLE_GUESTS
             - ENABLE_LOBBY
+            - ENABLE_XMPP_WEBSOCKET
             - GLOBAL_MODULES
             - GLOBAL_CONFIG
             - LDAP_URL
@@ -154,6 +156,7 @@ services:
             - JWT_AUTH_TYPE
             - JWT_TOKEN_AUTH_MODULE
             - LOG_LEVEL
+            - PUBLIC_URL
             - TZ
         networks:
             meet.jitsi:
--- a/prosody/rootfs/defaults/prosody.cfg.lua
+++ b/prosody/rootfs/defaults/prosody.cfg.lua
@@ -43,7 +43,7 @@ modules_enabled = {
 	-- Not essential, but recommended
 		"private"; -- Private XML storage (for room bookmarks, etc.)
 		"vcard"; -- Allow users to set vCards
-	
+
 	-- These are commented by default as they have a performance impact
 		--"privacy"; -- Support privacy lists
 		--"compression"; -- Stream compression (Debian: requires lua-zlib module to work)
@@ -59,7 +59,7 @@ modules_enabled = {
 	-- Admin interfaces
 		"admin_adhoc"; -- Allows administration via an XMPP client that supports ad-hoc commands
 		--"admin_telnet"; -- Opens telnet console interface on localhost port 5582
-	
+
 	-- HTTP modules
 		--"bosh"; -- Enable BOSH clients, aka "Jabber over HTTP"
 		--"http_files"; -- Serve static files from a directory over HTTP
@@ -165,5 +165,10 @@ network_settings = {
 component_interface = { "*" }
 
 data_path = "/config/data"
+
+smacks_max_unacked_stanzas = 5;
+smacks_hibernation_time = 60;
+smacks_max_hibernated_sessions = 1;
+smacks_max_old_sessions = 1;
 
 Include "conf.d/*.cfg.lua"


web/rootfs/etc/cont-init.d/10-config
```diff
@@ -85,6 +85,12 @@ if [[ ! -f /config/config.js ]]; then
         /config/config.js
     fi
 
+    if [[ $ENABLE_WEBSOCKET -eq 1 ]]; then
+        sed -i \
+          -e "s#// websocket:.*#websocket: document.location.protocol.replace('http', 'ws') + '//' + document.location.host + '/xmpp-websocket',#" \
```
saghul: This won't work on mobile I think. Also, we have plans to move away from having an executable config file. So, how about using PUBLIC_URL for this?

README.md
```diff
@@ -445,6 +445,7 @@ Variable | Description | Default value
 `JIGASI_TRANSCRIBER_ADVERTISE_URL` | Jigasi will post an URL to the chat with transcription file | true
 `DISABLE_HTTPS` | Handle TLS connections outside of this setup | 1
 `ENABLE_HTTP_REDIRECT` | Redirect HTTP traffic to HTTPS (necessary for Let's Encrypt) | 1
+`ENABLE_WEBSOCKET` | Enabled xmpp traffic over websockets | 0
```
saghul, line 448: Leet's call this ENABLE_XMPP_WEBSOCKET, since JVB datachannels can also be configured to use WS.

web/rootfs/etc/cont-init.d/10-config
jkroepke: Added if PUBLIC_URL is set.

README.md
jkroepke: Done

jkroepke:
Thanks for the review! I hope I catch all points.

netaskd:
I beleive xmpp websocket do not make sense without enabling wrbsockets for jvb

jkroepke:
@netaskd Can you explain it? As I understand, the communication between browser and jvb is WebRTC, no http polling. Or do you mean the connection between prosody and jvb?

Eleminate http pooling between browser and prosody is already a great benefit and will reduce the amount of requests massively.

netaskd:
@jkroepke in my tests, when I just enable xmpp-wss and pass webrtc datachanels for jvb, simulcast is working in low resolution(180) and bps. But when I enable jvb-wss, it shows me simulcast is working properly. I didn't go deep into the code, but I beleive, that it needs to work together (xmpp-wss + jvb-wss)
@saghul can you confirm or disprove this "theory"?

paweldomas:
XMPP websocket and JVB websocket are totally different things and they work independent

saghul:
@netaskd As @paweldomas pointed out they are completely independent. Can you re-test?

jkroepke:
@saghul Was your quick test successful?

web/rootfs/defaults/meet.conf
```diff
@@ -37,6 +39,22 @@ location = /http-bind {
...
+location = /xmpp-websocket {
+    proxy_pass {{ .Env.XMPP_BOSH_URL_BASE }}/xmpp-websocket;
+    proxy_http_version 1.1;
+
+    proxy_set_header Connection "upgrade";
+    proxy_set_header Upgrade $http_upgrade;
+
+    proxy_set_header Host {{ .Env.XMPP_DOMAIN }};
+    proxy_set_header X-Forwarded-For $remote_addr;
+    proxy_read_timeout 3600s;
```
saghul, line 53: Thiss value is too high, WS sends pings, so I. thinkwe can. do with the sssuggested 900, right?

env.example
```diff
@@ -319,6 +319,9 @@ JIBRI_LOGS_DIR=/config/logs
 # Necessary for Let's Encrypt, relies on standard HTTPS port (443)
 #ENABLE_HTTP_REDIRECT=1
 
+# Enabled xmpp traffic over websocket
```
saghul, line 322: Let's mention that. PUBLIC_URL is needed too. See my other comments below.

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -17,6 +19,14 @@ asap_accepted_issuers = { "{{ join "\",\"" (splitList "," .Env.JWT_ACCEPTED_ISSU
 asap_accepted_audiences = { "{{ join "\",\"" (splitList "," .Env.JWT_ACCEPTED_AUDIENCES) }}" }
 {{ end }}
 
+{{ if $ENABLE_XMPP_WEBSOCKET }}
+  {{ if .Env.PUBLIC_URL }}
```
saghul, line 23: There should be no else here, this variable should be required for WS to work.

web/rootfs/etc/cont-init.d/10-config
```diff
@@ -85,6 +85,18 @@ if [[ ! -f /config/config.js ]]; then
         /config/config.js
     fi
 
+    if [[ $ENABLE_XMPP_WEBSOCKET -eq 1 ]]; then
+        if [[ ! -z "$PUBLIC_URL" ]]; then
+            sed -i \
+              -e "s#// websocket:.*#websocket: '${PUBLIC_URL/http/ws}/xmpp-websocket',#" \
+            /config/config.js
+        else
```
saghul, line 93: Exit with an error otherwise please. Thsi will only lead to trouble as it doesn't work on mobile (and never will, because config.js is evaluated on a separate JS VM for security reasons).

web/rootfs/defaults/meet.conf
jkroepke: I misunderstand the documentation. I thought the default value of 890 was bumped.

jkroepke:
```suggestion
    proxy_read_timeout 900s;
```

web/rootfs/etc/cont-init.d/10-config
jkroepke: nice to know, thanks.

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
jkroepke: Done + Add a check in 10-config

README.md
```diff
@@ -457,6 +459,17 @@ succeed. If your users are coming in over the Internet (and not over LAN), this
 The public IP address is discovered via [STUN]. STUN servers can be specified with the ``JVB_STUN_SERVERS``
 option.
 
+### Prosody Websocket Support
```
saghul, line 462:
```suggestion
### Prosody WebSocket Support
```

README.md
```diff
@@ -457,6 +459,17 @@ succeed. If your users are coming in over the Internet (and not over LAN), this
...
+
+By default the Jitsi client communicate via http to prosody.
```
saghul, line 464:
```suggestion
By default communication with Prosody is done using HTTP.
```

README.md
```diff
@@ -457,6 +459,17 @@ succeed. If your users are coming in over the Internet (and not over LAN), this
...
+
+If `ENABLE_XMPP_WEBSOCKET` is set to `1` the preferred communication method is WebSocket. http pooling as fallback is still
```
saghul, line 466:
```suggestion
If `ENABLE_XMPP_WEBSOCKET` is set to `1` the preferred communication method is WebSocket. HTTP is sstill available as a fallback.
```

README.md
```diff
@@ -457,6 +459,17 @@ succeed. If your users are coming in over the Internet (and not over LAN), this
...
+available as fallback.
```
saghul, line 467:
```suggestion

```

README.md
```diff
@@ -457,6 +459,17 @@ succeed. If your users are coming in over the Internet (and not over LAN), this
...
+
+If Jitsi runs behind a reverse proxy make should the reverse proxy configuration supports WebSocket.
```
saghul, line 469:
```suggestion
If Jitsi runs behind a reverse proxy make sure the reverse proxy configuration supports WebSocket.
```

README.md
jkroepke: There was a typo
```suggestion
If `ENABLE_XMPP_WEBSOCKET` is set to `1` the preferred communication method is WebSocket. HTTP is still available as a fallback.
```

README.md
jkroepke: is this wrong?

paweldomas:
I haven't followed the entire thread, but I would actually suggest to set proxy timeout to 45 seconds or at least try how will it behave. In general the client is supposed to send a ping every 10 seconds or so and with the stream resumption module on prosody it is able to reconnect gracefully even after timeout. That  is the only way to prevent from leaving XMPP Websocket ghost for a longtime when the connection is dropped without giving the OS chance to close the TCP socket.

paweldomas:
You also need to apply this patch on the websocket module: https://github.com/jitsi/jitsi-meet/blob/master/resources/prosody-plugins/mod_websocket_smacks.patch
And add this module to prosody and enable it in the global "modules_enabled" section:
https://github.com/jitsi/jitsi-meet/blob/master/resources/prosody-plugins/mod_smacks.lua
Set  the following options in the prosody global cfg section:
```
smacks_max_unacked_stanzas = 5;
smacks_hibernation_time = 60;
smacks_max_hibernated_sessions = 1;
smacks_max_old_sessions = 1;
```

saghul:
Thanks for chiming in @paweldomas !

jkroepke:
@paweldomas 

2 questions:
* If the patch located at https://github.com/jitsi/jitsi-meet/ is the patch already applied inside jitsi docker images?
* Any caveats to enable smack by default? (with the logic if websocket is enabled)

jkroepke:
Branch rebased and squashed into one commit.

paweldomas:
> Any caveats to enable smack by default?
Only jicofo and other components must be close to the latest, because we've had a bug where they would enable stream resumption for themselves and then get stuck in a deadlock.

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -17,6 +19,10 @@ asap_accepted_issuers = { "{{ join "\",\"" (splitList "," .Env.JWT_ACCEPTED_ISSU
 asap_accepted_audiences = { "{{ join "\",\"" (splitList "," .Env.JWT_ACCEPTED_AUDIENCES) }}" }
 {{ end }}
 
+{{ if $ENABLE_XMPP_WEBSOCKET }}
+cross_domain_websocket = { "{{ .Env.PUBLIC_URL }}" };
```
paweldomas, line 23: I think we set this to:
```
cross_domain_websocket =  true;
consider_websocket_secure = true;
```
But if it works without the `consider_websocket_secure` option then I guess it's fine.

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -44,6 +50,9 @@ VirtualHost "{{ .Env.XMPP_DOMAIN }}"
     }
     modules_enabled = {
         "bosh";
+        {{ if $ENABLE_XMPP_WEBSOCKET }}
+        "websocket";
```
paweldomas, line 54: I'd put `smacks` here. It doesn't really make sense to have it enabled when only BOSH is used.

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
paweldomas: Maybe you won't need the `PUBLIC_URL` with these  options?

Screenshot 2020-04-23 18 22 37

jkroepke: > Maybe you won't need the PUBLIC_URL with these options?

See above: The web component needs a `PUBLIC_URL`.

```
cross_domain_websocket = true;
```

I learned that allow any origin in CORS policies is an anti pattern.

paweldomas:
@jkroepke Can  you check in JS console if you see the following msg:
Screenshot 2020-04-24 07 46 09
Also can you try executing in the JS console this code and see if it reconnects? This will confirm that the XMPP Websocket and stream resumption is setup properly.
`APP.conference._room.xmpp.connection.closeWebsocket()`

jkroepke:
image

Thats working but

> APP.conference._room.xmpp.connection.closeWebsocket()

results in a disconnect (and disconnect screen). After a site refresh (automatically done by the refresh screen), I'm 2 times in a room. (Jabber ID: snirdjtgvrp5vcd9@meet.jitsi/jk11rlav, Jabber ID: yfapih46vxljfqt6@meet.jitsi/fLWW4IK5)

image

Logs from console:
```
2020-04-24T16:07:08.182Z [modules/xmpp/XmppConnection.js] Trying to resume the XMPP connection
Logger.js:154 2020-04-24T16:07:08.184Z [modules/xmpp/xmpp.js] <A.connectionHandler>:  (TIME) Strophe connecting:	 12032.805000024382
Logger.js:154 2020-04-24T16:07:08.414Z [modules/xmpp/xmpp.js] <A.connectionHandler>:  (TIME) Strophe bindrequired:	 12262.555000022985
Logger.js:154 2020-04-24T16:07:08.434Z [modules/xmpp/xmpp.js] <A.connectionHandler>:  (TIME) Strophe error[item-not-found]:	 12282.115000009071
Logger.js:154 2020-04-24T16:07:08.438Z [modules/xmpp/xmpp.js] <A.connectionHandler>:  (TIME) Strophe disconnected:	 12286.20500001125
Logger.js:154 2020-04-24T16:07:08.439Z [modules/xmpp/strophe.ping.js] <p.stopInterval>:  Ping interval cleared
Logger.js:154 2020-04-24T16:07:08.441Z [modules/xmpp/xmpp.js] <A.connectionHandler>:  XMPP connection dropped!
```

Some prosody logs: 
```
c2s55b7af9eb290                                              debug	Received[c2s]: <enable resume='true' xmlns='urn:xmpp:sm:3'>
meet.jitsi:smacks                                            debug	Enabling stream management
c2s55b7af9eb290                                              debug	Sending[c2s]: <enabled max='60' id='7d3cf90e-4393-4ab3-88bf-9ccb9ddffd86' resume='true' xmlns='urn:xmpp:sm:3'>
mod_websocket                                                debug	Websocket received frame: opcode=1, 205 bytes
c2s55b7af9eb290                                              debug	Handled 1 incoming stanzas

...


server_epoll                                                 debug	New connection FD 16 (192.168.112.3, 32802, 192.168.112.2, 5280)
server_epoll                                                 debug	Watching FD 16 (192.168.112.3, 32802, 192.168.112.2, 5280)
http.server                                                  debug	Firing event: GET /xmpp-websocket
c2s55b7afacf990                                              info	Client connected
c2s55b7afacf990                                              debug	Sending WebSocket handshake
mod_websocket                                                debug	Websocket received frame: opcode=1, 81 bytes
c2s55b7afacf990                                              debug	Client sent opening <stream:stream> to meet.jitsi
c2s55b7afacf990                                              debug	Sending[c2s_unauthed]: <open id='32d55f2b-4544-4b8e-8da4-3238d108f7b5' from='meet.jitsi' version='1.0' xml:lang='en' xmlns='urn:ietf:params:xml:ns:xmpp-framing'>
c2s55b7afacf990                                              debug	Sent reply <stream:stream> to client
c2s55b7afacf990                                              debug	Offering mechanism ANONYMOUS
c2s55b7afacf990                                              debug	Sending[c2s_unauthed]: <stream:features>
mod_websocket                                                debug	Websocket received frame: opcode=1, 70 bytes
runnerabA_rUbo                                               debug	creating new coroutine
c2s55b7afacf990                                              debug	Received[c2s_unauthed]: <auth mechanism='ANONYMOUS' xmlns='urn:ietf:params:xml:ns:xmpp-sasl'>
c2s55b7afacf990                                              info	Authenticated as 0dkcjlojrzjjztwf@meet.jitsi
meet.jitsi:saslauth                                          debug	sasl reply: <success xmlns='urn:ietf:params:xml:ns:xmpp-sasl'/>
c2s55b7afacf990                                              debug	Sending[c2s_unbound]: <success xmlns='urn:ietf:params:xml:ns:xmpp-sasl'>
mod_websocket                                                debug	Websocket received frame: opcode=1, 81 bytes
c2s55b7afacf990                                              debug	Client sent opening <stream:stream> to meet.jitsi
c2s55b7afacf990                                              debug	Sending[c2s_unbound]: <open id='136e06b1-dd9e-49bf-9e56-f2b609292a97' from='meet.jitsi' version='1.0' xml:lang='en' xmlns='urn:ietf:params:xml:ns:xmpp-framing'>
c2s55b7afacf990                                              debug	Sent reply <stream:stream> to client
c2s55b7afacf990                                              debug	Sending[c2s_unbound]: <stream:features>
mod_websocket                                                debug	Websocket received frame: opcode=1, 84 bytes
c2s55b7afacf990                                              debug	Received[c2s_unbound]: <resume xmlns='urn:xmpp:sm:3' h='10' previd='7d3cf90e-4393-4ab3-88bf-9ccb9ddffd86'>
c2s55b7afacf990                                              debug	Tried to resume non-existent session with id 7d3cf90e-4393-4ab3-88bf-9ccb9ddffd86
c2s55b7afacf990                                              debug	Sending[c2s_unbound]: <failed xmlns='urn:xmpp:sm:3'>
mod_websocket                                                debug	Websocket received frame: opcode=8, 0 bytes
server_epoll                                                 debug	Close FD 16 (192.168.112.3, 32802, 192.168.112.2, 5280) after writing
c2s55b7af3b1c30                                              debug	Received[c2s]: <presence id='CK8HY-403' to='jvbbrewery@internal-muc.meet.jitsi/0dc02ace27bb'>
internal-muc.meet.jitsi:muc                                  debug	presence update for jvbbrewery@internal-muc.meet.jitsi/0dc02ace27bb from session jvb@auth.meet.jitsi/H-XfcyH9
c2s55b7aefe50c0                                              debug	Sending[c2s]: <presence id='CK8HY-403' from='jvbbrewery@internal-muc.meet.jitsi/0dc02ace27bb' to='focus@auth.meet.jitsi/focus112461803424802'>
c2s55b7af3b1c30                                              debug	Sending[c2s]: <presence id='CK8HY-403' from='jvbbrewery@internal-muc.meet.jitsi/0dc02ace27bb' to='jvb@auth.meet.jitsi/H-XfcyH9'>
server_epoll                                                 debug	Close FD 16 (192.168.112.3, 32802, 192.168.112.2, 5280) now
c2s55b7afacf990                                              info	Client disconnected: connection closed
c2s55b7afacf990                                              debug	Destroying session for (unknown) (0dkcjlojrzjjztwf@meet.jitsi)
```

jkroepke:
The patch is correctly applied:

```
patch -d /usr/lib/prosody/modules/ -p0 < /prosody-plugins/mod_websocket_smacks.patch
patching file mod_websocket.lua
Hunk #3 succeeded at 229 with fuzz 1.
Hunk #4 succeeded at 277 (offset 1 line).
```

jkroepke:
@paweldomas 

Is the patch https://github.com/jitsi/jitsi-meet/blob/master/resources/prosody-plugins/mod_websocket_smacks.patch correct?

The patch adds a new parameter `conn` to two functions but no additional code added that handle the new parameter.

paweldomas:
I see that it  fails with `item-not-found` which is caused by Prosody not finding this session.
So I forgot that actually token authentication module is required for this to work:
https://github.com/jitsi/jitsi-meet/blob/master/resources/prosody-plugins/mod_auth_token.lua#L32
Probably you need to adjust from anonymous to 'token' and allow empty token which then will work like anonymous:
```
authentication = "token"
        allow_empty_token = true
```

paweldomas:
Hm, but also this part is weird:
```
c2s55b7afacf990                                              debug	Tried to resume non-existent session with id 7d3cf90e-4393-4ab3-88bf-9ccb9ddffd86
```

jkroepke:
Is this compatible with setups with authentication enabled, e.g.: `authentication = "cyrus"` ?

jkroepke:
@paweldomas Sure. To reduce the logs.. I just restart the docker setup, then open a room and just disconnect the WebSocket.

https://gist.github.com/jkroepke/813799f5d6b01266041212c411b40f27

1. Connect: Line 1403
2. Connect: Line 1785

paweldomas:
It can be compatible, but you'd need a separate global module which does that:
```
-- Extract 'token' param from URL when session is created
function init_session(event)
	local session, request = event.session, event.request;
	local query = request.url.query;

	if query ~= nil then
        local params = formdecode(query);
        -- previd is used together with https://modules.prosody.im/mod_smacks.html
        -- the param is used to find resumed session and re-use anonymous(random) user id
        -- (see get_username_from_token)
        session.previd = query and params.previd or nil;
    end
end

module:hook_global("bosh-session", init_session);
module:hook_global("websocket-session", init_session);
```
Probably the code I pasted above placed in any module would do.

paweldomas:
Ah no, actually the code that's missing for anonymous part to work is this:
https://github.com/jitsi/jitsi-meet/blob/master/resources/prosody-plugins/mod_auth_token.lua#L85

paweldomas:
So the problem with anonymous auth is that every time you connected you will get a random user ID. The code above is yet another hack to reassign the previous user id if the resume token is correct.

With other types of authentication, where user id remains the same between reconnects this may work without any hacks.

jkroepke:
Still not working even with token authentication. Same error.

I do a commit again. Please recheck.

Logs: https://gist.github.com/jkroepke/813799f5d6b01266041212c411b40f27#file-prosody2-log

1. Connect: 503
2. Connect:  704

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -63,7 +82,15 @@ VirtualHost "{{ .Env.XMPP_DOMAIN }}"
 
 {{ if and $ENABLE_AUTH (.Env.ENABLE_GUESTS | default "0" | toBool) }}
 VirtualHost "{{ .Env.XMPP_GUEST_DOMAIN }}"
-    authentication = "anonymous"
+    -- https://github.com/jitsi/docker-jitsi-meet/pull/502#issuecomment-619146339
+    modules_enabled = {
+        "auth_token";
```
paweldomas, line 87: this is not needed

jkroepke:
I debug a little bit, too.

With token auth, I also get always a new random token (with is my username).
https://github.com/jitsi/jitsi-meet/blob/851976ebdf5465bdf34827e9bd80d1c4fed74639/resources/prosody-plugins/mod_auth_token.lua#L104

smacks saves the sessions per user:
https://github.com/jitsi/jitsi-meet/blob/851976ebdf5465bdf34827e9bd80d1c4fed74639/resources/prosody-plugins/mod_smacks.lua#L85

Putting log statements inside the cache functions results into this:

```
c2s559db91ffa70                                              info	Authenticated as 9198c527-7ddf-427e-a007-a3dbdf6091df@meet.jitsi
meet.jitsi:smacks_jitsi                                      warn	Set User Key: 9198c527-7ddf-427e-a007-a3dbdf6091df
meet.jitsi:smacks_jitsi                                      warn	Set Cache Key: 418bbfbc-a295-4628-8550-3168cc5e422f
meet.jitsi:smacks_jitsi                                      warn	Set Cache Value: table: 0x559db91ffa70
...

c2s559db8ce0370                                              info	Authenticated as 47e1c263-4dfb-4df3-b326-7482fcd453e2@meet.jitsi
meet.jitsi:smacks_jitsi                                      warn	Get Cache User: 47e1c263-4dfb-4df3-b326-7482fcd453e2
meet.jitsi:smacks_jitsi                                      warn	Get Cache Key: 418bbfbc-a295-4628-8550-3168cc5e422f
meet.jitsi:smacks_jitsi                                      warn	Get Cache Value: <nil>
```

jkroepke:
As I mention [earlier](https://github.com/jitsi/docker-jitsi-meet/pull/502#issuecomment-619125139), the patch for mod_websocket isn't complete.

It looks like the method init_session from auth_token isn't call for websockets since there is no event like `websocket-session`. I can just find the `bosh-session` event.
https://github.com/bjc/prosody/blob/0.11/plugins/mod_bosh.lua#L313

I also can't find nothing about websocket-session inside the container:

```
root@e7594a3972dc:/# grep bosh-session -R /usr/lib/prosody/
/usr/lib/prosody/modules/mod_bosh.lua:          module:fire_event("bosh-session", { session = session, request = request });
root@e7594a3972dc:/# grep websocket-session -R /usr/lib/prosody/
```

paweldomas:
> As I mention earlier, the patch for mod_websocket isn't complete.

> It looks like the method init_session from auth_token isn't call for websockets since there is no event like websocket-session. I can just find the bosh-session event.
https://github.com/bjc/prosody/blob/0.11/plugins/mod_bosh.lua#L313

You're right about  that, that's the issue. I thought our patch made it to Prosody, maybe it's not released yet or something, I need to check with @aaronkvanmeerten

jkroepke:
@paweldomas 
I modified the patch: https://github.com/jitsi/jitsi-meet/pull/6365

paweldomas:
So I see this change in trunk:
https://hg.prosody.im/trunk/file/tip/plugins/mod_websocket.lua#l299
and also 0.11:
https://hg.prosody.im/0.11/file/tip/plugins/mod_websocket.lua#l308
What version is used here?

jkroepke:
0.11.5 - Looking here: https://prosody.im/downloads/source/prosody-0.11.5.tar.gz 

The commit is not included inside the lastest stable version

aaronkvanmeerten:
The changes will be in 0.11.6. They didn't make it in time for 0.11.5

saghul:
Since we are using Prosody from the official repos and will get the update as soon as it’s released, WDYT about waiting for a release before landing this, so we don need to undo the patch part later? @paweldomas @aaronkvanmeerten

paweldomas:
> Since we are using Prosody from the official repos and will get the update as soon as it’s released, WDYT about waiting for a release before landing this, so we don need to undo the patch part later?

Yes, I would vote to wait for it to land.

jkroepke:
Ok, I get rid of all hacks/tests now. Until then .. can you give a review? Then it it can be just merged until a new Prosody is out.

paweldomas:
> Ok, I get rid of all hacks/tests now. Until then .. can you give a review? Then it it can be just merged until a new Prosody is out.

Can we wait until it actually lands? Otherwise we'll get flooded by complaints  about not working feature?

jkroepke:
> Can we wait until it actually lands?
Thats not what I mean.

I'm just ask for a code review to be ready to merge if a new prosody version is out.

README.md
```diff
@@ -457,6 +459,16 @@ succeed. If your users are coming in over the Internet (and not over LAN), this
...
+### Prosody WebSocket Support
+
+By default communication with Prosody is done using HTTP.
+
+If `ENABLE_XMPP_WEBSOCKET` is set to `1` the preferred communication method is WebSocket. HTTP is still available as a fallback.
+
+If Jitsi runs behind a reverse proxy make sure the reverse proxy configuration supports WebSocket.
+
+Read more: http://nginx.org/en/docs/http/websocket.html
+
```
luixxiul, line 471: These changes need to be applied to https://github.com/jitsi/handbook/blob/master/docs/devops-guide/docker.md (cf: #601)

web/rootfs/defaults/meet.conf
```diff
@@ -37,6 +39,22 @@ location = /http-bind {
...
+location = /xmpp-websocket {
+    proxy_pass {{ .Env.XMPP_BOSH_URL_BASE }}/xmpp-websocket;
+    proxy_http_version 1.1;
+
+    proxy_set_header Connection "upgrade";
+    proxy_set_header Upgrade $http_upgrade;
+
+    proxy_set_header Host {{ .Env.XMPP_DOMAIN }};
+    proxy_set_header X-Forwarded-For $remote_addr;
+    proxy_read_timeout 60s;
```
bkempe, line 53: > A value higher than the socket framework network_settings.read_timeout setting is advisable. (default 840 seconds in trunk).
(https://prosody.im/doc/websocket#nginx)
Any thoughts on this?

paweldomas: We haven't play with this much yet, but I think it's fine. We want to get a timeout on inactivity and this one is specifically for the `/xmpp-websocket` path.

jkroepke:
@paweldomas prosody 0.11.5 https://prosody.im/doc/release/0.11.5 is released. This can be merged soon?

paweldomas:
> @paweldomas prosody 0.11.5 https://prosody.im/doc/release/0.11.5 is released. This can be merged soon?

Oh, that's cool. I will merge it after trying out

bkempe:
@paweldomas https://github.com/jitsi/docker-jitsi-meet/pull/502#issuecomment-619384019 : 0.11.5 does not contain the patch yet, but pulling it in manually would work:
```
RUN \
    curl -o /tmp/mod_websocket.patch https://hg.prosody.im/0.11/raw-rev/37936c72846d \
    && patch -d /usr/lib/prosody/modules/ -p2 < /tmp/mod_websocket.patch
```

jkroepke:
> Released: 2020-03-24

Ops, my fault. Sorry. +1 for manual patch

paweldomas:
@jkroepke have you ever ran this when docker was on localhost? what would you use as the `PUBLIC_URL`? is there the entry in `/etc/hosts` necessary?

jkroepke:
@paweldomas 

Yes. Looking at my browser history, I used http://localhost:8000/ . I can't remember what I had set on `PUBLIC_URL`. It's 2 months ago.

But I'm using Docker for Mac which means: Using `localhost` may have a different meaning compared to Docker on Linux, since Docker for Mac runs a virtual machine. 

From browser side its localhost, but from container side it's my network ip.

>  is there the entry in /etc/hosts necessary?

No.

pdarcos:
Hi,

What's the best way to enable sockets on debian 10 running jitsi-meet unstable?
Should I wait for this PR to be merged? Any ETA on when that will happen?

Thanks

DanielMcAssey:
Will this be merged anytime soon?

pdarcos:
I managed to get websockets enabled by following this thread, including the patch, enabling smacks and adding the suggested configuration above to nginx.

However I see tons of error messages related to e2ee ping failing in the console logs.

And when I ran APP.conference._room.xmpp.connection.closeWebsocket() in the console my connection was immediately dropped. When it reconnected after the timeout my ghost was still in the room.
Hope this report helps.

Going back to data channels for now until there's some update on this front.

Cheers

jkroepke:
I can't remeber 100% but I used the 0.11 nightly version of prosody - not the single patch.

https://github.com/jitsi/docker-jitsi-meet/pull/502#issuecomment-619378208

Replace https://github.com/jitsi/docker-jitsi-meet/blob/baed6055692f53b963ff66ca933ed0caa990aa7e/prosody/Dockerfile#L30

with `prosody-0.11 \` to get the [0.11 nightly builds](https://prosody.im/nightly/0.11/latest/).

pdarcos:
> I can't remeber 100% but I used the 0.11 nightly version of prosody - not the single patch.
> 
> [#502 (comment)](https://github.com/jitsi/docker-jitsi-meet/pull/502#issuecomment-619378208)
> 
> Replace
> 
> https://github.com/jitsi/docker-jitsi-meet/blob/baed6055692f53b963ff66ca933ed0caa990aa7e/prosody/Dockerfile#L30
> 
> with `prosody-0.11 \` to get the [0.11 nightly builds](https://prosody.im/nightly/0.11/latest/).

Thanks @jkroepke 

I'm not using docker. I'm running the debian unstable packages directly. But I'll try to install prosody 0.11 nightly build tomorrow and see if that works better than the current dirty hacks.

Cheers

DanielMcAssey:
Ive tried running this branch with JWT tokens. The tokens dont seem to be getting parsed. Using the http-bind service endpoint works fine with the token. However the websocket immediately returns a 403 unauthorised error.

jkroepke:
Does it work? @pdarcos

pdarcos:
> Does it work? @pdarcos

Sorry @jkroepke but I ended up having a work emergency and haven't had the time to try it yet.
Hopefully I can find some time to update prosody to latest nightly later today.

DanielMcAssey:
@pdarcos Just for reference updating to Prosody 0.11 nightly works for me. Got the websockets working with JWT auth

paweldomas:
> @jkroepke have you ever ran this when docker was on localhost? what would you use as the `PUBLIC_URL`? is there the entry in `/etc/hosts` necessary?

Just a note that I've managed to test this on localhost by setting `PUBLIC_URL=https://localhost:8443`

paweldomas:
@jkroepke can you please rebase? I wanted to push the manual mod_weboscket patch and merge

jkroepke:
@paweldomas as mention above patching mod_websocket is not enough.

The 0.11 nightly branch/build is required that this works.

saghul:
> The 0.11 nightly branch/build is required that this works.

I don't think we want to ship this with a nightly build requirement :-/

paweldomas:
> @paweldomas as mention above patching mod_websocket is not enough.

> The 0.11 nightly branch/build is required that this works.

It should be enough. I don't understand what is missing. Is it about JWT tokens? I think they should work with the patch

paweldomas:
@jkroepke can you please rebase and force push? I'll put the manual patch in and maybe some can check if JWT works for them

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -41,14 +50,22 @@ VirtualHost "{{ .Env.XMPP_DOMAIN }}"
     authentication = "internal_hashed"
   {{ end }}
 {{ else }}
-    authentication = "anonymous"
+    -- https://github.com/jitsi/docker-jitsi-meet/pull/502#issuecomment-619146339
+    authentication = "token"
```
saghul, line 54: IS this still the case @paweldomas ? Looks very counter-intuitive.

paweldomas: It is the case because that token module extracts the resume token from the URL. Maybe this should be configured this way only for the WebSockets. I don't feel like writing another auth module - the default anonymous module would have to be modified I think.

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -41,14 +50,22 @@ VirtualHost "{{ .Env.XMPP_DOMAIN }}"
...
     ssl = {
         key = "/config/certs/{{ .Env.XMPP_DOMAIN }}.key";
         certificate = "/config/certs/{{ .Env.XMPP_DOMAIN }}.crt";
     }
     modules_enabled = {
         "bosh";
+        {{ if $ENABLE_XMPP_WEBSOCKET }}
+        "websocket";
+        {{ end }}
+        "smacks"; -- XEP-0198: Stream Management
```
paweldomas, line 68: "smacks" should also be enabled only if `$ENABLE_XMPP_WEBSOCKET`

jmcrthrs:
Hey there, I tried testing this by building from the `jkroepke:add-websocket-support`branch, but the client was still using /http-bind? and in the browser console I got this message: 

`2020-07-24T15:20:32.689Z [modules/xmpp/XmppConnection.js] <_maybeEnableStreamResume>:  Stream resume enabled, but WebSockets are not enabled`

There are the values I set in my .env:
```
ENABLE_XMPP_WEBSOCKET=1
PUBLIC_URL=my.app.com
```

I also tried the following: 
```
PUBLIC_URL=https://my.app.com
PUBLIC_URL=//my.app.com
```

paweldomas:
Have you cleared the config (`~/.jitsi-meet-cfg`)?

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
jkroepke, line 68:
```suggestion
        "smacks"; -- XEP-0198: Stream Management
        {{ end }}
```

paweldomas:
@jkroepke thanks! I still plan to add 2 changes: 1. manually patch websocket for the websocket session event and 2. switch to token auth on the anonymous domain only when websockets are enabled. After that I'll merge the PR

jkroepke:
> 1. manually patch websocket for the websocket session event

I guess this should done here https://github.com/jitsi/jitsi-meet/blob/master/resources/prosody-plugins/mod_websocket_smacks.patch ?

Otherwise non docker deployment won't work with websockets.

> 2. switch to token auth on the anonymous domain only when websockets are enabled. 

Done

atripier:
Hey there! I'm just posting some information here in hope that this MR doesn't die :)

This websocket session event [patch](https://hg.prosody.im/0.11/raw-rev/37936c72846d) is included in [Prosody 0.11.6](https://prosody.im/doc/release/0.11.6), which was released earlier this month.

However docker-jitsi-meet is currently based on Debian Stretch: it seems that the 0.11.6 release of Prosody is packaged for Buster and above, not Stretch which is still on 0.11.5: https://packages.prosody.im/debian/pool/main/p/prosody/

I guess we have to wait until Buster/Java 11 is approved for use in production for Jitsi so that this MR can be merged? https://github.com/jitsi/docker-jitsi-meet/pull/227#issuecomment-624486062

saghul:
Worry not it won't die. We are a bit short on time these days, hopefully I can get back to this soon.

sapkra:
@jkroepke Can you rebase it please because there have been some major changes on the `config.js` stuff.

bukharin:
We are enable `mod_smacks` over WebSockets in in our jitsi meet installation, because websocket connection may be unstable. 

We receive a great result. See our client-side metrics for amount of conference fails events. This is must have feature by default.

image

jkroepke:
If the dev team have the time to review and merge this, I'll do a rebase. Otherwise the rebase is pointless.

bukharin:
Team already implements X-0198 on meet.jit.si, because stream resumption works there

manang:
HI,
I'm trying to use this version of docker-compose.
When I try to connect with the websocket I obtain:
WebSocket connection to 'wss://<my-ip>/xmpp-websocket?room=workshop014' failed: Error in connection establishment: net::ERR_CERT_INVALID

I'm using the self signed certificates.
Thanks

saghul:
> If the dev team have the time to review and merge this, I'll do a rebase. Otherwise the rebase is pointless.

We are committed to landing this, please do rebase. Sorry it has taken this long.

jkroepke:
I hope, I merged correctly.

I saw `ENABLE_WEBSOCKETS` already available inside new config template.

web/rootfs/defaults/meet.conf
```diff
@@ -50,6 +52,22 @@ location = /http-bind {
...
```
saghul, line 66: Can we drop this timeout?

prosody/rootfs/etc/cont-init.d/10-config
```diff
@@ -26,6 +26,11 @@ if [[ "$(stat -c %U /prosody-plugins-custom)" != "prosody" ]]; then
 fi
 
 if [[ ! -f $PROSODY_CFG ]]; then
+    if [[ $ENABLE_XMPP_WEBSOCKET -eq 1 && -z "$PUBLIC_URL" ]]; then
```
saghul, line 29: Let's drop this, PUBLIC_URL is now mandatory anyway. You can add something like this: https://github.com/jitsi/docker-jitsi-meet/blob/0c957941bc36795dccf241deaec18e23a004c279/jvb/rootfs/defaults/sip-communicator.properties#L29

saghul (changes requested):
A couple of minor changes, LGTM otherwise. @paweldomas / @damencho Do we know when / if the mod_smacks patch is applied to Prosody?

web/rootfs/defaults/meet.conf
jkroepke: We could drop the timeout, but the nginx default is still `proxy_read_timeout 60s;`

Do you want to drop the line or do you want to remove the timeout?

damencho: if you are talking about mod_websocket_smacks.patch, you can skip it. We recently discussed with Matt the patches and that one is not needed and we are currently running without it on meet.jit.si ... maybe we can also remove it from jitsi-meet repo.

saghul: > Do you want to drop the line or do you want to remove the timeout?

The line please.

saghul: Thanks @damencho !

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -23,6 +26,13 @@ asap_accepted_issuers = { "{{ join "\",\"" (splitList "," .Env.JWT_ACCEPTED_ISSU
 asap_accepted_audiences = { "{{ join "\",\"" (splitList "," .Env.JWT_ACCEPTED_AUDIENCES) }}" }
 {{ end }}
 
+{{ if $ENABLE_XMPP_WEBSOCKET }}
+-- Deprecated in 0.12
+-- https://github.com/bjc/prosody/commit/26542811eafd9c708a130272d7b7de77b92712de
+cross_domain_websocket = { "{{ $PUBLIC_URL_DOMAIN }}" };
```
saghul, line 32:
```suggestion
cross_domain_websocket = { "{{ $PUBLIC_URL }}" };
```

saghul (changes requested):

prosody/Dockerfile
```diff
@@ -48,6 +48,7 @@ RUN \
     && rm -rf /tmp/pkg /var/cache/apt
 
 RUN patch -d /usr/lib/prosody/modules/muc -p0 < /prosody-plugins/muc_owner_allow_kick.patch
+RUN patch -d /usr/lib/prosody/modules/ -p0 < /prosody-plugins/mod_websocket_smacks.patch
```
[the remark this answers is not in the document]
jkroepke:
```suggestion

```

web/rootfs/defaults/meet.conf
saghul, line 66:
```suggestion
```

saghul:
One last comment, LGTM otherwise.

saghul (approved):
LGTM. I'll give it a test and merge it shortly.

saghul:
Landed in d747bfb with a couple of changes to match the new / current config options. Great work @jkroepke ! Sorry it took this long to merge!

PR #512, opened 2020-04-23
jibri: chrome/driver 78 as a stopgap

netaskd:


jibri/Dockerfile
```diff
@@ -1,8 +1,10 @@
 ARG JITSI_REPO=jitsi
 FROM ${JITSI_REPO}/base-java
 
-ARG CHROME_RELEASE=latest
-ARG CHROMEDRIVER_MAJOR_RELEASE=latest
+#ARG CHROME_RELEASE=latest
+#ARG CHROMEDRIVER_MAJOR_RELEASE=latest
```
sapkra, line 5: Do we still need these comments? If you want to tell the user that he can change the version and why it is forced to be `78` it should be documented in the README or right here in the file.

saghul: They are not needed... but I think they serve the purpose of raising an eyebrow and git blame would do the rest. At any rate I hope we don't have to keep it for long...

tele-tom:
I used the current master branch and can see the changes in jibri dockerfile on my machine.
When I setup everything I get an error when starting recording.
In the logs still new version of chrome and chrome driver appear:

`2020-05-01 14:06:38.632 SEVERE: [52] org.jitsi.jibri.selenium.JibriSelenium.run() An error occurred while joining the call: org.openqa.selenium.WebDriverException: <unknown>: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.
  (Session info: chrome=81.0.4044.113)
  (Driver info: chromedriver=81.0.4044.69 (6813546031a4bc83f717a2ef7cd4ac6ec1199132-refs/branch-heads/4044@{#776}),platform=Linux 4.9.0-12-amd64 x86_64) (WARNING: The server did not provide any stacktrace information)`

I'm confused.... please advice.

saghul:
Did you rebuild the image?

tele-tom:
I did a docker system prune -af --volumes while containers were stopped and ran a normal setup according to the readme.

tele-tom:
With the new release I have the correct versions now. So don't know  what went wrong before.

But I get the same error with the older driver versions. Do I need to do anything extra or just a standard setup. Runs behind reverse proxy that does the TLS. jitsi is connected via thhp port 8000.

PR #519, opened 2020-04-24
Update Jibri Dockerfile to fix recoding problem

MaximMonin:
ignore-certificate-errors


saghul:
We have stopgapped this by using Chrome 78 for the time being.

PR #545, opened 2020-05-03
Add jq dep for upload integrations in Jibri #535

### LINKED ISSUE #535: Graceful_shutdown script can not be run, as curl, jq are missing in videobridge container
There is a graceful_shutdown.sh script in usr/share/jitsi-videobridge in the JVB-container which can't be run as the dependencies curl and jq are not installed. However, this script can probably be used for scaling the JVBs.

MatthewCroughan:
The [Jitsi Upload Integration](https://github.com/jitsi/jitsi-upload-integrations) scripts do not work since Jibri's container does not install the dependencies required for these scripts, specifically `jq`. When attempting to run dropbox_uploader.sh for example, `jq` will not be found. This fixes #535

### DIFF
--- a/jibri/Dockerfile
+++ b/jibri/Dockerfile
@@ -40,7 +40,7 @@ RUN \
 RUN \
         [ "$JITSI_RELEASE" = "unstable" ] \
         && apt-dpkg-wrap apt-get update \
-        && apt-dpkg-wrap apt-get install -y jitsi-upload-integrations \
+        && apt-dpkg-wrap apt-get install -y jitsi-upload-integrations jq \
         && apt-cleanup \
         || true
 


jibri/Dockerfile
```diff
@@ -40,7 +40,7 @@ RUN \
 RUN \
         [ "$JITSI_RELEASE" = "unstable" ] \
         && apt-dpkg-wrap apt-get update \
-        && apt-dpkg-wrap apt-get install -y jitsi-upload-integrations \
+        && apt-dpkg-wrap apt-get install -y jitsi-upload-integrations jq \
```
sapkra, line 43: Why just installing it for unstable releases?

PR #547, opened 2020-05-03
change CONFIG=

kopista:
better default directory for config, if upgrade installation (git clone, git pull) and leave old cfg dir, 
mistakes are made that difficult explained


saghul:
Nobody has complained about the default thus far, and you are free to change it, going through the values in the env file is something everyone should do before deployment, so I'm sorry but I won't take this change.

PR #552, opened 2020-05-04
Fixes #548

### LINKED ISSUE #548: jitsi-upload-integrations aren't installed in Jibri despite dockerfile definition?
I'm really confused, it seems that the jitsi-upload-integrations really aren't present in any version of Jibri and have not been for some time despite the definition being clear here:

https://github.com/jitsi/docker-jitsi-meet/blob/2d063ad162370b600aa5d986aaebbe2cd8bf435d/jibri/Dockerfile#L40-L45

Is this a consequence of the silly `|| true` which allows this to build without failure? Why has this been done? This seems to be a longstanding issue that people are really pulling their hair out over. 

https://community.jitsi.org/t/need-help-with-dropbox-integration/17896/19

When we do figure out why there's a `|| true`, can we please comment this into the dockerfile so people have some clue as to why this is defined but non functional?

mesouug:
Forces installation of jitsi-upload-integrations for all releases. Fixes #548

### DIFF
--- a/jibri/Dockerfile
+++ b/jibri/Dockerfile
@@ -38,11 +38,9 @@ RUN \
 	&& chromedriver --version
 
 RUN \
-        [ "$JITSI_RELEASE" = "unstable" ] \
-        && apt-dpkg-wrap apt-get update \
+        apt-dpkg-wrap apt-get update \
         && apt-dpkg-wrap apt-get install -y jitsi-upload-integrations jq \
-        && apt-cleanup \
-        || true
+        && apt-cleanup
 
 COPY rootfs/ /
 


jibri/Dockerfile
```diff
@@ -38,11 +38,9 @@ RUN \
 	&& chromedriver --version
 
 RUN \
-        [ "$JITSI_RELEASE" = "unstable" ] \
-        && apt-dpkg-wrap apt-get update \
+        apt-dpkg-wrap apt-get update \
         && apt-dpkg-wrap apt-get install -y jitsi-upload-integrations jq \
```
sapkra, line 42: Why not just moving it to line 11?

PR #555, opened 2020-05-06
Added environment var JICOFO_ENABLE_HEALTH_CHECKS

cedricroijakkers:
Since release 4548, health checks are disabled in Jicofo. These can be enabled with configuration key org.jitsi.jicofo.health.ENABLE_HEALTH_CHECKS. Since health checks can be very handy when running Jitsi in a container (i.e. when using the Kubernetes liveness probes) we need an environment variable in the container to switch it on.

This PR introduces environment variable JICOFO_ENABLE_HEALTH_CHECKS with default value false, matching the Jicofo defaults, but an option to set to true to re-enable the REST API health checks, which can then be used in the Kubernetes liveness probe.


saghul:
Can you please document this on the README too?

cedricroijakkers:
Done, I've added it to the documentation as well.

PR #559, opened 2020-05-07
Renew LetsEncrypt with webroot config

miguelpeixe:
To avoid having to stop the nginx server every renewal attempt, the certificate could be renewed using webroot configured with the nginx server.

### DIFF
--- a/web/rootfs/defaults/default
+++ b/web/rootfs/defaults/default
@@ -1,6 +1,16 @@
 server {
 	listen 80 default_server;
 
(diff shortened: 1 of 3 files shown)


saghul:
TIL! This looks very nice, I'll give it a try!

miguelpeixe:
Cool! Let me know if you need anything. I tested by running a `--dry-run`.

saghul:
We have since switched to acme.sh so this would need a do-over. Sorry about that.

PR #560, opened 2020-05-07
Fix: Dockerfile cleanup

Berbe:
- Minimise the number of `RUN` stages will help reducing the number of Docker layers, make the overall image history leanier, hence more readable/accountable
- Avoid unneeded APT update & clean
- Homogeneise spacing, previously mixing space & tabs indent
- Homogeneise directives linking using a suffix notation: helps readability

### DIFF
--- a/base-java/Dockerfile
+++ b/base-java/Dockerfile
@@ -1,9 +1,7 @@
 ARG JITSI_REPO=jitsi
 FROM ${JITSI_REPO}/base
 
-RUN	\
-	mkdir -p /usr/share/man/man1 && \
-	apt-dpkg-wrap apt-get update && \
-	apt-dpkg-wrap apt-get install -y openjdk-8-jre-headless && \
-	apt-cleanup
-
+RUN mkdir -p /usr/share/man/man1 && \
+    apt-dpkg-wrap apt-get update && \
+    apt-dpkg-wrap apt-get install -y openjdk-8-jre-headless && \
+    apt-cleanup
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -10,25 +10,20 @@ ADD https://github.com/subchen/frep/releases/download/v1.3.5/frep-1.3.5-linux-am
 
 COPY rootfs /
 
-RUN \
-	tar xfz /tmp/s6-overlay.tar.gz -C / && \
-	rm -f /tmp/*.tar.gz && \
-	apt-dpkg-wrap apt-get update && \
-	apt-dpkg-wrap apt-get install -y apt-transport-https apt-utils ca-certificates gnupg && \
-	apt-key add /tmp/jitsi.key && \
-	rm -f /tmp/jitsi.key && \
-	echo "deb https://download.jitsi.org $JITSI_RELEASE/" > /etc/apt/sources.list.d/jitsi.list && \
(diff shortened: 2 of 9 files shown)


base-java/Dockerfile
```diff
@@ -1,9 +1,8 @@
...
-	mkdir -p /usr/share/man/man1 && \
-	apt-dpkg-wrap apt-get update && \
-	apt-dpkg-wrap apt-get install -y openjdk-8-jre-headless && \
-	apt-cleanup
-
+RUN mkdir -p /usr/share/man/man1 && \
+    apt-dpkg-wrap apt-get update && \
+    apt-dpkg-wrap apt-get install -y openjdk-8-jre-headless && \
+    apt-cleanup && \
+    rm -rf /tmp/pkg /var/cache/apt
```
saghul, line 8: Nothing creates /tmp/pkg here, no need to remove it. Also, how about deleting /var/cache/apt in apt-cleanup?

Berbe: The actual useless path there is `/tmp/pkg`. Removing `/var/cache/apt` is just a more radical version of `apt-get clean` (too much of one?) and is already used in combination with it on the master branch:
https://github.com/jitsi/docker-jitsi-meet/blob/b039b29b7aeeee28f9f7395ec87e8aa69105ce9f/web/Dockerfile#L17
I was rather conservative in the approach I took, but you're right: anything related to APT cleanup shall be moved to the cleaning script.

vkruoso:
As a best-practice building Dockerfiles, people often recommend to pin the package versions the image needs to build. That way you can be sure that the build is deterministic. How do you feel about that?

Besides that point, the images use the pre-built packages in debian repository. I could not easily find were all debian packages are built. Can someone point me in the right direction?

Berbe:
@vkruoso I would definitely recommend pinning packages to specific versions, otherwise the release version of your Docker images does not mean anything: their consistency over time is not guaranteed as  their content will be associated with the state of the APT repositories being used on build time.
The way I would do it would be to provide specific packages version through `ARG` on build time and do on-the-fly APT pinning in the cooking recipe.

However, that feels a tad far-fetched for this PR, as right now, like you, I have no precise idea on how debian packages are provisioned.

vkruoso:
Yeah, that is not for the scope of this PR. I was trying to get a fell on how that would be seen by the maintainers. I really think the debian package building could be clarified and we may be help with improving docs on the process. I did look in the docs but cloud not find what actually creates the packages.

saghul:
> As a best-practice building Dockerfiles, people often recommend to pin the package versions the image needs to build. That way you can be sure that the build is deterministic. How do you feel about that?

We cannot do that right now, because images are built using our Debian repository.



> Besides that point, the images use the pre-built packages in debian repository. I could not easily find were all debian packages are built. Can someone point me in the right direction?

We build tghem on a Jenkins machine. It's a process we trigger manually (for the stable repository) when we want to promote a release to stable.

vkruoso:
Thanks for the response @saghul. Does the package build process (jenkins job) is in this repository? I've tried building the Docker image myself directly from the source, instead of the debian package, but without success. To make a custom image with interface changes, I had to use the stable image and overwrite the static generated files.

saghul:
Just as I initially thought, it's very hard to keep these things in check, and we're deep in conflict land now. I think at this point this PR would just add more churn than solve actual problems.

Some of the proposed changes were good, but we'd need them split away from space changes and the like.

PR #561, opened 2020-05-07
[readme] avoid confusion about gen-passwords.sh

lebreRafael:
I had a misunderstanding about gen-passwords.sh when runnig docker-jitsi-meet for the first time. Instead of running the bash file I edited it.

With this PR we respect the following pattern: description in the first asterisk and command to run in the nested asterisk.


saghul:
cheers!

PR #562, opened 2020-05-07
Doc: Add missing config directories to QuickStart

rkost:
This adds the two config directories introduced via f7796a160e8d621f307bdd36a60d3d1d96afe902 to the quick start section in the readme.md as the containers would not start without them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ follow these steps:
   * `cp env.example .env`
   * Set strong passwords in the security section options: `./gen-passwords.sh`
 * Create required `CONFIG` directories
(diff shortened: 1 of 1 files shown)


saghul:
I have never needed to create those by hand. What are you using to run the containers? Is it something other than Docker Compose?

rkost:
I also wondered why I need to create these manually. I am using docker compose 1.24.0 with docker 18.09.8 on a Synology NAS. Only checked out the stable tag stable-4548 and ran into this problem. 

Is docker-compose supposed to create these directories automatically? Maybe I have to check my setup then...

PR #584, opened 2020-05-12
moved COPY command after RUN

prayagsingh:
moved `COPY rootfs/` command after `RUN` commands to stop the overwriting of custom `config.js` and `interface_config.js` files.

### DIFF
--- a/web/Dockerfile
+++ b/web/Dockerfile
@@ -3,8 +3,6 @@ FROM ${JITSI_REPO}/base
 
 ADD https://dl.eff.org/certbot-auto /usr/local/bin/
 
-COPY rootfs/ /
-
 RUN \
 	apt-dpkg-wrap apt-get update && \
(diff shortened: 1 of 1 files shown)


prayagsingh:
I have moved `COPY rootfs/` command after `RUN` in web dockerfile because if we put custom `config.js` and `interface_config.js` in `rootfs` directory and `COPYING` it before `RUN` command then those custom config.js files got overwrite by config.js files installed by `RUN` commands. Another way(if we don't want to use `COPY rootfs/` after `RUN` command) to avoid this is to comment out below two lines in `Dockerfile` but this is not a good approach.
```
mv /tmp/pkg/usr/share/jitsi-meet-web-config/config.js /defaults && \
mv /usr/share/jitsi-meet/interface_config.js /defaults && \
```
@saghul Could you please review it?

saghul:
This is not something I want to support, sorry.

I'm still debating what is the best way to allow config customization, but this is not it.

PR #588, opened 2020-05-13
Fix: Build images directly with proper tag

Berbe:
The `Makefile` recipe previously added a special `tag` step to tag locally built images (with `latest` tag) with the defined `JITSI_BUILD` one.


saghul:
Hi there! Thanks for taking the time to send a PR!

Can you please explain a bit what the motivation behind the change? What problem does it try to fix?

Berbe:
If `JITSI_BUILD` is defined to something else than `latest`:
1. 2 steps are required to use the new tag (`make` & `make tag-all`)
2. images are uselessly being built with both tags

The whole purpose of tags is to allow building several versions of a product on the same system, allowing seamless switch between those.

saghul:
The workflow I’ve been using to build images is to build the latest image, then tag the release that needs to go out and push both.

Looks to me that with your proposal I’d still need to tag images with latest too, before pushing, right?

If that is the case I really see no advantage.

sapkra:
Just maybe if you change something and build the images by yourself while your changes are not based on the current master your containers might break because when you rebuild your working fork it might break because of breaking changes in the lastest debian packages (jitsi, prosody etc.).
If they are locked to a specific version it will never happen.

Berbe:
`latest`... is not a version. This tag has been used & abused, just because it is the default one when you do not specify any building Docker images.

That said, by default, the `Makefile` contains `latest` as `JITSI_BUILD` value, which was the case before my changes; hence integrating my changes won't change that. You can use whatever non-release-specific state of trunk, and build it with `latest` tag.
On annother hand, if you build a *specific* version, ie you want to make a release, then there is no use to build `latest` too and have all your images double-tagged: with my proposal, you will build all your images directly with the proper tag and you can them push them with it.

With this proposal, the `Makefile` recipe generates all images tagged with a single tag, the one specified in `JITSI_BUILD`.

saghul:
Sorry, but I still see no advantage here. I would need to adapt my current release workflow, for little benefit and there are many other ways I could spend that time.

PR #594, opened 2020-05-16
Provide a more consistent behavior on multi-homed docker hosts

hillct:
A simple change, adding DOCKER_HOST_ADDRESS to the ports declarations in docker-compose.yaml such that the host address selection as set in the .env file will be honored. Also uncommented DOCKER_HOST_ADDRESS in env.eample, with a sane default for single-homed docker hosts. Not sure whether this minor adjustment necessitates update of related documentation referenced in the comment above DOCKER_HOST_ADDRESS in env.example...

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,8 +6,8 @@ services:
         image: jitsi/web:latest
         restart: ${RESTART_POLICY}
         ports:
-            - '${HTTP_PORT}:80'
-            - '${HTTPS_PORT}:443'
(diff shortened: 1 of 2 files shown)


saghul:
That is not what that option is for. Please check the documentation on the README.

That (ill named) option relates to the JVB public interface.

hillct:
Having reviewed the documentation thoroughly, before submitting this fix, and not findinf an alternate preferred approach to addressing deployment on multi-homed infastructure, I invite to to point me to the specific documentation you referenced above (a line number might be helpful).

PR #602, opened 2020-05-20
Update TLS config to Mozilla security guidelines.

JonathanLennox:


web/rootfs/defaults/ssl.conf
```diff
@@ -13,10 +13,11 @@ ssl_certificate_key /config/keys/cert.key;
 {{ end }}
 
 # protocols
-ssl_protocols TLSv1.2;
-ssl_prefer_server_ciphers on;
-ssl_ecdh_curve secp384r1;
-ssl_ciphers ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES256-GCM-SHA384:ECDH+CHACHA20:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA:ECDHE-ECDSA-AES256-SHA:!aNULL:!eNULL:!EXPORT:!DES:!RC4:!3DES:!MD5:!PSK;
+# Mozilla Guideline v5.4, nginx 1.17.7, OpenSSL 1.1.1d, intermediate configuration, no OCSP
+# https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=intermediate&openssl=1.1.1d&ocsp=false&guideline=5.4
+ssl_protocols TLSv1.2 TLSv1.3;
```
saghul, line 18: we don't yet support 1.3 because the base image is squeeze, which won't get OpenSSL 1.1.1

web/rootfs/defaults/ssl.conf
```diff
@@ -13,10 +13,11 @@ ssl_certificate_key /config/keys/cert.key;
...
+ssl_ciphers ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384;
+ssl_prefer_server_ciphers off;
```
saghul, line 20: This one is somewhat surprising!

JonathanLennox: You mean `ssl_prefer_server_ciphers off`?  In the rationale at https://wiki.mozilla.org/Security/Server_Side_TLS they say:
> The cipher suites are all strong and so we allow the client to choose, as they will know best if they have support for hardware-accelerated AES

web/rootfs/defaults/ssl.conf
saghul: stretch, that is

web/rootfs/defaults/ssl.conf
saghul: 👍

web/rootfs/defaults/ssl.conf
```diff
@@ -13,10 +13,11 @@ ssl_certificate_key /config/keys/cert.key;
 {{ end }}
 
 # protocols
+# Mozilla Guideline v5.4, nginx 1.17.7, OpenSSL 1.1.1d, intermediate configuration, no OCSP
+# https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=intermediate&openssl=1.1.0l&ocsp=false&guideline=5.4
 ssl_protocols TLSv1.2;
-ssl_prefer_server_ciphers on;
-ssl_ecdh_curve secp384r1;
-ssl_ciphers ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES256-GCM-SHA384:ECDH+CHACHA20:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA:ECDHE-ECDSA-AES256-SHA:!aNULL:!eNULL:!EXPORT:!DES:!RC4:!3DES:!MD5:!PSK;
+ssl_ciphers ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384;
```
plokta, line 19: A note regarding `DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384`: Diffie-Hellman ciphers had been previously removed with https://github.com/jitsi/docker-jitsi-meet/pull/450 as discussed in https://github.com/jitsi/docker-jitsi-meet/issues/433
In case you add any DH ciphers again, it will also be necessary to add a `dhparams.pem` file again, which has also been removed with the PR mentioned above.

JonathanLennox: Following Mozilla's advice, I'd like to deploy ffdhe2048 as the DH params, rather than generating them randomly.

However, I'm not so familiar with Docker configuration - what's the best place to install this, both in the repo and in the docker?

saghul: Is that a file we'd like to include? In that case you can put it right on this directory, which will be available as /defaaults in the container.

saghul:
Landed an updated version on b0cb4a1 (generated today, with TLS 1.3 since we now use a recent enough Nginx)

Issue #529, opened 2020-04-27
[web] Configure Matomo settings using environment variables


ludovicm67:
Once https://github.com/jitsi/jitsi-meet/issues/6411 is done, it can be very useful to be able to configure Matomo settings using environment variables, like for example:

- `MATOMO_ENDPOINT`: full URL of Matomo instance,
- `MATOMO_SITE_ID`: the site ID given by the Matomo instance

c-prompt:
Thanks for the great functionality. As you improve it, please consider grabbing variables sent via the [JWT token authentication](https://github.com/jitsi/lib-jitsi-meet/blob/master/doc/tokens.md) to include as [Matomo custom variables](https://developer.matomo.org/guides/tracking-javascript-guide#custom-variables) (e.g., `_paq.push(['setCustomVariable','1','VisitorType','Member']);`) or other user-related data such as [UserID](https://developer.matomo.org/guides/tracking-javascript-guide#user-id) (e.g., `_paq.push(['setUserId', 'USER_ID_HERE']);`).

Since it's currently limited to what's defined in config.js, it'd be useful for this to be a bit more customizable. Additionally, there are use cases when you want to send multiple Matomo Site IDs (e.g., record visit on main site and also record same visit on different site) so it'd be great to designate a Matomo custom variable (e.g., setSiteId) in the JWT and, if the MatomoHandler class sees the JWT custom variable, it can add a _paq.push (e.g., `_paq.push(['setSiteId', {$JWT_CUSTOM_IDSITE}]);`)

Issue #527, opened 2020-04-27
[web] Make watermark customizable using environment variables


ludovicm67:
It could be really useful if the watermark can be customized using environment variables, for example:
 - `WEB_WATERMARK_IMAGE`: the watermark can be an other internal path or a complete url to a logo which is hosted outside of the container, like on a CDN for example.
 - `WEB_WATERMARK_LINK`: the link where the user will be redirected after clicking on the watermark.

dinigo:
hi. I opened a similar issue some time ago that got transfered to the jitsi web repo 

https://github.com/jitsi/adocker-jitsi-meet/issues/301

PR #619, opened 2020-05-28
FEATURE: web: build config.js on each boot

aaronkvanmeerten:
FEATURE: web: build system-config.js from tpl and env vars
FEATURE: web: include /config/custom-config.js if provided

### DIFF
--- a/web/rootfs/defaults/settings-config.js
+++ b/web/rootfs/defaults/settings-config.js
@@ -1,0 +1,270 @@
+{{ $DEPLOYMENTINFO_USERREGION := .Env.DEPLOYMENTINFO_USERREGION | default "" -}}
+{{ $BRIDGE_CHANNEL := .Env.BRIDGE_CHANNEL | default "datachannel" -}}
+{{ $ENABLE_AUDIO_PROCESSING := .Env.ENABLE_AUDIO_PROCESSING | default "true" | toBool -}}
+{{ $ENABLE_BRIDGE_CHANNEL := .Env.ENABLE_BRIDGE_CHANNEL | default "true" -}}
+{{ $ENABLE_CALENDAR := .Env.ENABLE_CALENDAR | default "false" | toBool -}}
+{{ $ENABLE_FILE_RECORDING_SERVICE := .Env.ENABLE_FILE_RECORDING_SERVICE | default "false" | toBool -}}
+{{ $ENABLE_FILE_RECORDING_SERVICE_SHARING := .Env.ENABLE_FILE_RECORDING_SERVICE_SHARING | default "false" | toBool -}}
+{{ $ENABLE_IPV6 := .Env.ENABLE_IPV6 | default "true" | toBool -}}
+{{ $ENABLE_LIPSYNC := .Env.ENABLE_LIPSYNC | default "false" | toBool -}}
+{{ $ENABLE_NO_AUDIO_DETECTION := .Env.ENABLE_NO_AUDIO_DETECTION | default "false" | toBool -}}
+{{ $ENABLE_P2P := .Env.ENABLE_P2P | default "true" | toBool -}}
+{{ $ENABLE_RECORDING := .Env.ENABLE_RECORDING | default "false" | toBool -}}
+{{ $ENABLE_REMB := .Env.ENABLE_REMB | default "true" | toBool -}}
+{{ $ENABLE_REQUIRE_DISPLAY_NAME := .Env.ENABLE_REQUIRE_DISPLAY_NAME | default "false" | toBool -}}
+{{ $ENABLE_SIMULCAST := .Env.ENABLE_SIMULCAST | default "true" | toBool -}}
+{{ $ENABLE_STATS_ID := .Env.ENABLE_STATS_ID | default "false" | toBool -}}
+{{ $ENABLE_STEREO := .Env.ENABLE_STEREO | default "false" | toBool -}}
+{{ $ENABLE_TALK_WHILE_MUTED := .Env.ENABLE_TALK_WHILE_MUTED | default "false" | toBool -}}
+{{ $ENABLE_TCC := .Env.ENABLE_TCC | default "true" | toBool -}}
+{{ $ENABLE_TRANSCRIPTIONS := .Env.ENABLE_TRANSCRIPTIONS | default "false" | toBool -}}
+{{ $P2P_USE_STUN_TURN := .Env.P2P_USE_STUN_TURN | default "true" | toBool -}}
+{{ $RESOLUTION := .Env.RESOLUTION | default "720" -}}
+{{ $RESOLUTION_MIN := .Env.RESOLUTION_MIN | default "180" -}}
+{{ $RESOLUTION_WIDTH := .Env.RESOLUTION_WIDTH | default "1280" -}}
+{{ $RESOLUTION_WIDTH_MIN := .Env.RESOLUTION_WIDTH_MIN | default "320" -}}
+{{ $START_AUDIO_MUTED := .Env.START_AUDIO_MUTED | default 10 -}}
+{{ $START_VIDEO_MUTED := .Env.START_VIDEO_MUTED | default 10 -}}
+{{ $TESTING_OCTO_PROBABILITY := .Env.TESTING_OCTO_PROBABILITY | default "0" -}}
+{{ $TESTING_CAP_SCREENSHARE_BITRATE := .Env.TESTING_CAP_SCREENSHARE_BITRATE | default "1" -}}
+{{ $USE_STUN_TURN := .Env.USE_STUN_TURN | default "true" | toBool -}}
+{{ $XMPP_DOMAIN := .Env.XMPP_DOMAIN -}}
+{{ $XMPP_RECORDER_DOMAIN := .Env.XMPP_RECORDER_DOMAIN -}}
+
+
+// Video configuration.
+//
+
+if (!config.hasOwnProperty('constraints')) config.constraints = {};
+if (!config.constraints.hasOwnProperty('video')) config.constraints.video = {};
+
+config.resolution = {{ $RESOLUTION }};
+config.constraints.video.height = { ideal: {{ $RESOLUTION }}, max: {{ $RESOLUTION }}, min: {{ $RESOLUTION_MIN }} };
+config.constraints.video.width = { ideal: {{ $RESOLUTION_WIDTH }}, max: {{ $RESOLUTION_WIDTH }}, min: {{ $RESOLUTION_WIDTH_MIN }}};
+config.disableSimulcast = {{ not $ENABLE_SIMULCAST }};
+config.startVideoMuted = {{ $START_VIDEO_MUTED }};
+
+{{ if .Env.START_BITRATE -}}
+config.startBitrate = '{{ .Env.START_BITRATE }}';
+{{ end -}}
+
+
+// Audio configuration.
+//
+
+config.enableNoAudioDetection = {{ $ENABLE_NO_AUDIO_DETECTION }};
+config.enableTalkWhileMuted = {{ $ENABLE_TALK_WHILE_MUTED }};
+config.disableAP = {{ not $ENABLE_AUDIO_PROCESSING }};
+config.stereo = {{ $ENABLE_STEREO }};
+config.startAudioMuted = {{ $START_AUDIO_MUTED }};
+
+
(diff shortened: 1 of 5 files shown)


aaronkvanmeerten:
Work in progress, do not merge

aaronkvanmeerten:
> Also I noticed you added the subdomain part. We need to fully bake that in the Docker setup, but that’s for another day :-)

I can also make an additional PR to the nginx config to enable it there, and a separate one to enable it in the prosody config.  This one simply addresses the needed pieces in config.js.  I'd like to keep that part in this PR as I'll need it to actual do any deployments using this method (or build a custom-config.js just for that until it does get merged, I suppose).

web/rootfs/defaults/settings-config.js
```diff
@@ -0,0 +1,259 @@
+{{ $XMPP_DOMAIN := .Env.XMPP_DOMAIN | default "jitsi-meet.example.com" -}}
```
saghul, line 1: doesn't the extra dash `-}}` break things?

web/rootfs/defaults/settings-config.js
```diff
@@ -0,0 +1,259 @@
...
+{{ $CONFIG_BOSH_HOST := .Env.CONFIG_BOSH_HOST | default "" }}
+{{ $CONFIG_RESOLUTION := .Env.CONFIG_RESOLUTION | default "720" -}}
+{{ $CONFIG_RESOLUTION_MIN := .Env.CONFIG_RESOLUTION_MIN | default "180" -}}
+{{ $CONFIG_RESOLUTION_WIDTH := .Env.CONFIG_RESOLUTION_WIDTH | default "1280" -}}
+{{ $CONFIG_RESOLUTION_WIDTH_MIN := .Env.CONFIG_RESOLUTION_WIDTH | default "320" -}}
```
saghul, line 6:
```suggestion
{{ $CONFIG_RESOLUTION_WIDTH_MIN := .Env.CONFIG_RESOLUTION_WIDTH_MIN | default "320" -}}
```

web/rootfs/defaults/settings-config.js
```diff
@@ -0,0 +1,259 @@
...
+{{ $CONFIG_DISABLE_SIMULCAST := .Env.CONFIG_DISABLE_SIMULCAST | default "false" -}}
+{{ $CONFIG_ENABLE_REMB := .Env.CONFIG_ENABLE_REMB | default "true" -}}
+{{ $CONFIG_ENABLE_TCC := .Env.CONFIG_ENABLE_TCC | default "true" -}}
+{{ $ENABLE_RECORDING := .Env.ENABLE_RECORDING | default "false" | toBool -}}
+{{ $CONFIG_FILE_RECORDING_SERVICE_ENABLED := .Env.CONFIG_FILE_RECORDING_SERVICE_ENABLED | default "false" | toBool -}}
+{{ $CONFIG_FILE_RECORDING_SERVICE_SHARING_ENABLED := .Env.CONFIG_FILE_RECORDING_SERVICE_SHARING_ENABLED | default "false" | toBool -}}
+{{ $CONFIG_EXTERNAL_CONNECT := .Env.CONFIG_EXTERNAL_CONNECT | default "false" | toBool -}}
+{{ $XMPP_RECORDER_DOMAIN := .Env.XMPP_RECORDER_DOMAIN | default "" -}}
+{{ $XMPP_RECORDER_DOMAIN_PREFIX := .Env.XMPP_RECORDER_DOMAIN_PREFIX | default "recorder" -}}
+{{ $CONFIG_CHROME_MIN_EXT_VERSION := .Env.CONFIG_CHROME_MIN_EXT_VERSION | default "0.1" -}}
```
saghul, line 16: Let's not add the screen-sharing extension related options, since they no longer apply... we just need to find the time to axe the code.

web/rootfs/defaults/settings-config.js
```diff
@@ -0,0 +1,259 @@
...
+config.constraints.video.width = {ideal: {{$CONFIG_RESOLUTION_WIDTH}}, max: {{$CONFIG_RESOLUTION_WIDTH}}, min: {{$CONFIG_RESOLUTION_WIDTH_MIN}}};
+
+config.enableTcc = {{ $CONFIG_ENABLE_TCC }};
+config.enableRemb = {{ $CONFIG_ENABLE_REMB }};
+
+// Enable / disable simulcast support.
+config.disableSimulcast = {{ $CONFIG_DISABLE_SIMULCAST }};
+
+// testing values, defaults should be sane
+if (!config.hasOwnProperty('testing')) config.testing={};
```
saghul, line 62:
```suggestion
Object.assign(config.testing, {});
```

I need to check if Duktape supports this though.

web/rootfs/defaults/settings-config.js
```diff
@@ -0,0 +1,259 @@
...
+{{ $CONFIG_REQUIRE_DISPLAY_NAME := .Env.CONFIG_REQUIRE_DISPLAY_NAME | default "false" | toBool -}}
+{{ $CONFIG_START_VIDEO_MUTED := .Env.CONFIG_START_VIDEO_MUTED | default 10 -}}
+{{ $CONFIG_START_AUDIO_MUTED := .Env.CONFIG_START_AUDIO_MUTED | default 10 -}}
+{{ $CONFIG_USE_STUN_TURN := .Env.CONFIG_USE_STUN_TURN | default "true" | toBool -}}
+{{ $CONFIG_USE_IPV6 := .Env.CONFIG_USE_IPV6 | default "true" | toBool -}}
+{{ $CONFIG_DEPLOYMENTINFO_USERREGION := .Env.CONFIG_DEPLOYMENTINFO_USERREGION | default "<!--# echo var=\"http_x_proxy_region\" default=\"\" -->" -}}
+
+{{ if $CONFIG_EXTERNAL_CONNECT -}}
+    {{ if $ENABLE_SUBDOMAINS -}}
+config.externalConnectUrl = '//{{ if .Env.CONFIG_BOSH_HOST }}{{ .Env.CONFIG_BOSH_HOST }}{{ end }}/<!--# echo var="subdir" default="" -->http-pre-bind';
```
saghul, line 40: shouldn't this go on the system ones?

web/rootfs/defaults/system-config.js
```diff
@@ -0,0 +1,55 @@
...
+{{ if $ENABLE_GUESTS -}}
+// When using authentication, domain for guest users.
+config.hosts.anonymousdomain = '{{ .Env.XMPP_GUEST_DOMAIN }}';
+{{ end -}}
+{{ if $ENABLE_AUTH -}}
+// Domain for authenticated users. Defaults to <domain>.
+config.hosts.authdomain = '{{ $XMPP_DOMAIN }}';
+{{ end -}}
+
+// BOSH URL. FIXME: use XEP-0156 to discover it.
```
saghul, line 51: Drop the fixme, it's never going to be fixed :-P

web/rootfs/defaults/settings-config.js
saghul, line 1: We need to sort these options somehow (alphabetically?) so it doesn't become a mess, WDYT?

saghul: also, I'm not sure about having the defaults here. It means that the generated config will always be set and we have to keep it in sync with the internals. I'd rather pass it blindly I think, except for specific cases.

web/rootfs/etc/cont-init.d/10-config
```diff
@@ -68,46 +68,11 @@ if [[ ! -f /config/nginx/site-confs/default ]]; then
...
-             /config/config.js
-    fi
+cp /defaults/config.js /config/config.js
+tpl /defaults/system-config.js >> /config/config.js
+tpl /defaults/settings-config.js >> /config/config.js
+if [[ -f /config/custom-config.js ]]; then
+    cat /config/custom-config.js >> /config/config.js
 fi
 
 if [[ ! -f /config/interface_config.js ]]; then
```
saghul, line 78: While we're here, same for interface_config?

web/rootfs/defaults/settings-config.js
aaronkvanmeerten: Negatory, the dash simply removes whitespace after this statement, in this case the newline.  This keeps us from having a large number of newlines at the top of the file before the content.

web/rootfs/defaults/settings-config.js
aaronkvanmeerten: Great, I've applied this everywhere

web/rootfs/defaults/settings-config.js
aaronkvanmeerten: It can, it's not required, but it is more of a system setting, so I'll move it.

web/rootfs/etc/cont-init.d/10-config
aaronkvanmeerten: I'm hesitant to tackle that part yet, since it's a whole separate bag of worms.  I'd prefer to make that a separate PR.

web/rootfs/defaults/settings-config.js
aaronkvanmeerten: done

web/rootfs/defaults/system-config.js
aaronkvanmeerten: Done

web/rootfs/defaults/settings-config.js
aaronkvanmeerten: Reverting this one:
Uncaught TypeError: Cannot convert undefined or null to object

web/rootfs/defaults/settings-config.js
aaronkvanmeerten: Removed defaults.  Also sorting the options alphabetically for the time being, until a better scheme comes along.

aaronkvanmeerten:
Big updates, including config option renames to avoid redundancy, and increase consistency.  Needs another review please.

ludovicm67:
Hi, thanks for your great work :+1: 

Can you also add support for some missing configuration keys?

It could be great if you could add support for Matomo (see here: https://github.com/jitsi/jitsi-meet/blob/6533071c4c91a5850c5545c81eea0df93ebf15aa/config.js#L404-L406) ; this will close #529.

Can you also add support for the `brandingDataUrl` key configuration (see: https://github.com/jitsi/jitsi-meet/blob/6533071c4c91a5850c5545c81eea0df93ebf15aa/config.js#L530)? This would help for #527.

saghul (approved):

saghul:
I know we didn't cover all options, please feel free to send PRs for each individual one! 🚀

nwvaras:
Is this working? It seems that if i add variables to the system-config or the settings-config it doesn't add to the config.js file. It seems like its been overrided by another file

web/rootfs/defaults/settings-config.js
```diff
@@ -0,0 +1,265 @@
...
+{{ $ENABLE_FILE_RECORDING_SERVICE_SHARING := .Env.ENABLE_FILE_RECORDING_SERVICE_SHARING | default "false" | toBool -}}
+{{ $ENABLE_IPV6 := .Env.ENABLE_IPV6 | default "true" | toBool -}}
+{{ $ENABLE_LIPSYNC := .Env.ENABLE_LIPSYNC | default "false" | toBool -}}
+{{ $ENABLE_NO_AUDIO_DETECTION := .Env.ENABLE_NO_AUDIO_DETECTION | default "false" | toBool -}}
+{{ $ENABLE_STUN_TURN := .Env.ENABLE_STUN_TURN | default "true" | toBool -}}
+{{ $ENABLE_RECORDING := .Env.ENABLE_RECORDING | default "false" | toBool -}}
+{{ $ENABLE_REMB := .Env.ENABLE_REMB | default "true" -}}
+{{ $ENABLE_REQUIRE_DISPLAY_NAME := .Env.ENABLE_REQUIRE_DISPLAY_NAME | default "false" | toBool -}}
+{{ $ENABLE_SIMULCAST := .Env.ENABLE_SIMULCAST | default "true" -}}
+{{ $ENABLE_STATS_ID := .Env.ENABLE_STATS_ID | default "false" | toBool -}}
```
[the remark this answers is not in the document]
chagai95: Is this working? I tried it and it didn't change the config.js @madmath03 @saghul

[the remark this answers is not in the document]
saghul: It should, yes.

PR #630, opened 2020-06-06
jvb: set LOCAL_ADDRESS to the correct local IP

debfx:
NAT_HARVESTER_LOCAL_ADDRESS needs to bet set to the local IP that is actually
used to communicate with clients.

In Docker containers "hostname -I" returns the IPs of all interfaces. The order
of the IPs depends on the interface create order.

Docker makes no guarantees about the interface order and in my tests it is
random.

As a result NAT_HARVESTER_LOCAL_ADDRESS is sometimes set to the IP of the
internal meet.jitsi interface.

Fix this by setting it to the source IP used for routing to $DOCKER_HOST_ADDRESS

Adds 2803 kB disk space for iproute2 and dependencies.

Partly addresses #338

### DIFF
--- a/jvb/Dockerfile
+++ b/jvb/Dockerfile
@@ -3,7 +3,7 @@ FROM ${JITSI_REPO}/base-java
 
 RUN \
 	apt-dpkg-wrap apt-get update && \
-	apt-dpkg-wrap apt-get install -y jitsi-videobridge2 jq curl && \
+	apt-dpkg-wrap apt-get install -y jitsi-videobridge2 jq curl iproute2 && \
 	apt-cleanup
 
 COPY rootfs/ /
--- a/jvb/rootfs/etc/services.d/jvb/run
+++ b/jvb/rootfs/etc/services.d/jvb/run
@@ -3,7 +3,7 @@
 JAVA_SYS_PROPS="-Dnet.java.sip.communicator.SC_HOME_DIR_LOCATION=/ -Dnet.java.sip.communicator.SC_HOME_DIR_NAME=config -Djava.util.logging.config.file=/config/logging.properties"
 
 if [[ ! -z "$DOCKER_HOST_ADDRESS" ]]; then
-    LOCAL_ADDRESS=$(hostname -I | cut -d " " -f1)
+    LOCAL_ADDRESS=$(ip route get "$DOCKER_HOST_ADDRESS" | head -n1 | cut -d " " -f7)
     JAVA_SYS_PROPS="$JAVA_SYS_PROPS -Dorg.ice4j.ice.harvest.NAT_HARVESTER_LOCAL_ADDRESS=$LOCAL_ADDRESS -Dorg.ice4j.ice.harvest.NAT_HARVESTER_PUBLIC_ADDRESS=$DOCKER_HOST_ADDRESS"
 fi
 


saghul:
What IPs do you egg in your container? You should get localist and the one network interface, which is the one we need here.

Note that DOCKED_HOST_ADDRESS can be undefined, so I don’t think your proposed approach would be correct there.

debfx:
> What IPs do you egg in your container? You should get localist and the one network interface, which is the one we need here.

You are right that this doesn't happen in the default configuration as there is only one interface.

I have configured jitsi with one interface that is connected to the internet and the meet.jitsi interface that is only used for communication between the jitsi services. A bit like the traefik example.

> Note that DOCKED_HOST_ADDRESS can be undefined, so I don’t think your proposed approach would be correct there.

This code is only run when DOCKED_HOST_ADDRESS isn't empty.

aeburriel:
> What IPs do you egg in your container? You should get localist and the one network interface, which is the one we need here.
> 
> Note that DOCKED_HOST_ADDRESS can be undefined, so I don’t think your proposed approach would be correct there.

In a swarm, using traefik's sample stack, the container has two network interfaces: internal jitsi network and docker_gwbridge.

For us, without @debfx patch, Jitsi is unusable, working randomly from time to time.
With this patch and setting DOCKER_HOST_ADDRESS, media is relayed correctly between all participants.

saghul:
Sorry for the delay, I re-read this and you're right, the PR makes setting the local IP more resilient, cheers!

jordisala1991:
> DOCKER_HOST_ADDRESS

To me, with this patch, jitsi is unusable. I have docker + traefik setup here with the two network interfaces you mention.

Before (stable-4857), LOCAL_ADDRESS: 10.0.4.231 ($(hostname -I | cut -d " " -f1))
After (stable-5076), LOCAL_ADDRESS: 172.18.0.6 $(ip route get "$DOCKER_HOST_ADDRESS" | head -n1 | cut -d " " -f7)

Before I could have more than 3 participants in a meet without problems, now I just get this error all the time:

```
INFO: verify if nominated pair answer again
May 30, 2020 10:40:45 AM org.ice4j.stack.NetAccessManager handleFatalError
SEVERE: Unexpected Error!
java.lang.NullPointerException
	at org.ice4j.socket.MergingDatagramSocket.initializeActive(MergingDatagramSocket.java:577)
	at org.ice4j.ice.ComponentSocket.propertyChange(ComponentSocket.java:174)
	at org.ice4j.ice.IceMediaStream.firePairPropertyChange(IceMediaStream.java:870)
	at org.ice4j.ice.CandidatePair.nominate(CandidatePair.java:629)
	at org.ice4j.ice.Agent.nominate(Agent.java:1847)
	at org.ice4j.ice.DefaultNominator.strategyNominateFirstValid(DefaultNominator.java:144)
	at org.ice4j.ice.DefaultNominator.propertyChange(DefaultNominator.java:120)
	at org.ice4j.ice.IceMediaStream.firePairPropertyChange(IceMediaStream.java:870)
	at org.ice4j.ice.CandidatePair.validate(CandidatePair.java:667)
	at org.ice4j.ice.IceMediaStream.addToValidList(IceMediaStream.java:668)
	at org.ice4j.ice.Agent.validatePair(Agent.java:1811)
	at org.ice4j.ice.ConnectivityCheckClient.processSuccessResponse(ConnectivityCheckClient.java:638)
	at org.ice4j.ice.ConnectivityCheckClient.processResponse(ConnectivityCheckClient.java:405)
	at org.ice4j.stack.StunClientTransaction.handleResponse(StunClientTransaction.java:314)
	at org.ice4j.stack.StunStack.handleMessageEvent(StunStack.java:1040)
	at org.ice4j.stack.MessageProcessingTask.run(MessageProcessingTask.java:196)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

To discard other possible errors, I have downgraded to `stable-4857` and applied this patch, and the error appears too.

And I couldn't find a way to restore the previous behavior on the new stable release. How can I fix this issue?

debfx:
I am assuming jvb gets incoming connections from traefik on one network interface, but the default gateway is on the other interface. Is that correct?

If it is you'd have to change the network setup so default gateway and traefik are on the same network interface.

jordisala1991:
I don't know how to really check what you mean, at this moment I have the container with jitsi with this config:

```
version: '3.8'

services:
    prosody:
        image: jitsi/prosody:stable-4857
        volumes:
            - ${CONFIG}/prosody/config:/config:Z
            - ${CONFIG}/prosody/prosody-plugins-custom:/prosody-plugins-custom:Z
        environment:
            - AUTH_TYPE
            - ENABLE_AUTH
            - ENABLE_GUESTS
            - ENABLE_LOBBY
            - XMPP_DOMAIN
            - XMPP_AUTH_DOMAIN
            - XMPP_GUEST_DOMAIN
            - XMPP_MUC_DOMAIN
            - XMPP_INTERNAL_MUC_DOMAIN
            - XMPP_MODULES
            - XMPP_MUC_MODULES
            - XMPP_INTERNAL_MUC_MODULES
            - XMPP_RECORDER_DOMAIN
            - JICOFO_COMPONENT_SECRET
            - JICOFO_AUTH_USER
            - JICOFO_AUTH_PASSWORD
            - JVB_AUTH_USER
            - JVB_AUTH_PASSWORD
            - JWT_APP_ID
            - JWT_APP_SECRET
            - TZ
        networks:
            net-public:
                aliases:
                    - ${XMPP_SERVER}
        deploy:
            mode: global
            placement:
                constraints: [node.role == manager]
            update_config:
                parallelism: 1
                delay: 10s
            restart_policy:
                condition: on-failure
    jicofo:
        image: jitsi/jicofo:stable-4857
        volumes:
            - ${CONFIG}/jicofo:/config:Z
        environment:
            - ENABLE_AUTH
            - XMPP_DOMAIN
            - XMPP_AUTH_DOMAIN
            - XMPP_INTERNAL_MUC_DOMAIN
            - XMPP_SERVER
            - JICOFO_COMPONENT_SECRET
            - JICOFO_AUTH_USER
            - JICOFO_AUTH_PASSWORD
            - JICOFO_ENABLE_HEALTH_CHECKS
            - JVB_BREWERY_MUC
            - TZ
        depends_on:
            - prosody
        networks:
            net-public:
        deploy:
            mode: global
            placement:
                constraints: [node.role == manager]
            update_config:
                parallelism: 1
                delay: 10s
            restart_policy:
                condition: on-failure
    jvb:
        image: jitsi/jvb:stable-4857
        ports:
            - '${JVB_PORT}:${JVB_PORT}/udp'
            - '${JVB_TCP_PORT}:${JVB_TCP_PORT}'
        volumes:
            - ${CONFIG}/jvb:/config:Z
        environment:
            - DOCKER_HOST_ADDRESS
            - XMPP_AUTH_DOMAIN
            - XMPP_INTERNAL_MUC_DOMAIN
            - XMPP_SERVER
            - JVB_AUTH_USER
            - JVB_AUTH_PASSWORD
            - JVB_BREWERY_MUC
            - JVB_PORT
            - JVB_TCP_HARVESTER_DISABLED
            - JVB_TCP_PORT
            - JVB_STUN_SERVERS
            - JVB_ENABLE_APIS
            - PUBLIC_URL
            - TZ
        depends_on:
            - prosody
        networks:
            net-public:
                aliases:
                    - jvb.meet.jitsi
        deploy:
            mode: global
            placement:
                constraints: [node.role == manager]
            update_config:
                parallelism: 1
                delay: 10s
            restart_policy:
                condition: on-failure

networks:
    net-public:
        external: true
```

Jvb is not going through traefik, I couldn't make it work with this stable version. Docker creates the docker_gwbridge when I init the swarm and the net-public is another network that I have created to comunicate all containers, I think I am missing something but I can't figure it out.

sapkra:
@debfx Wouldn't it be possible to get the local address without trying to send a package to the docker `DOCKER_HOST_ADDRESS`? I don't think it will solve your problem, it's more like another topic.

e.g.
```bash
/sbin/ip route|awk '/scope/ { print $9 }'
```
or
```bash
grep "$HOSTNAME" /etc/hosts|awk '{print $1}'
```

Additionally we wouldn't need the `iproute2` package anymore.
The reason I'm asking is, that I need the `LOCAL_ADDRESS` even when the `DOCKER_HOST_ADDRESS` is not set and I don't want to implement a second solution.

debfx:
In theory yes since LOCAL_ADDRESS should just be the source IP that's used for the default gateway.

I haven't found a good way to determine that though.

`ip route` just prints something like: `default via 172.17.0.1 dev eth0`

When you have multiple network interfaces `/etc/hosts` contains multiple entries for the hostname, but you don't know which one is which.

For example:

```
172.22.0.3	29ebafddd457
172.20.0.5	29ebafddd457
```

You could run something like this and parse the IP from there but it really isn't pretty:

`ip addr show dev $(ip route|awk '/^default/ { print $5 }')`

sapkra:
@debfx Interesting because in kubernetes `ip route` prints the following, while `10.128.0.136` is my container IP:

```
default via 10.128.0.1 dev eth0 
10.128.0.0/23 dev eth0 proto kernel scope link src 10.128.0.136 
10.128.0.0/14 dev eth0 
224.0.0.0/4 dev eth0
```

Your command outputs the following in my case:

```
3: eth0@if305: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1450 qdisc noqueue state UP group default 
    link/ether 0a:58:0a:80:00:88 brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 10.128.0.136/23 brd 10.128.1.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::4c6d:a6ff:fec0:7159/64 scope link 
       valid_lft forever preferred_lft forever
```

### New solution (a mix of both):
With this it is trying to send a package to itself so we can get the ip for the outgoing traffic. Does this work for you?
```bash
ip route get $(grep "$HOSTNAME" /etc/hosts|awk '{print $1}') | head -n1 | cut -d " " -f6
```

debfx:
Here is a full example of `ip route`:
```
default via 172.22.0.1 dev eth0 
172.20.0.0/16 dev eth1 proto kernel scope link src 172.20.0.5 
172.22.0.0/16 dev eth0 proto kernel scope link src 172.22.0.3 
```

and `/etc/hosts`:

```
127.0.0.1	localhost
::1	localhost ip6-localhost ip6-loopback
fe00::0	ip6-localnet
ff00::0	ip6-mcastprefix
ff02::1	ip6-allnodes
ff02::2	ip6-allrouters
172.22.0.3	29ebafddd457
172.20.0.5	29ebafddd457
```

In this case we need to make sure 172.22.0.3 is picked as LOCAL_ADDRESS (since the default gateway is on that interface), not 172.20.0.5.

I don't think relying on the order of IPs in `/etc/hosts` is a good idea. I doubt Docker makes any guarantees about that.

Something along the lines of this seems to be the most reliable way to me:
```
ip addr show dev $(ip route|awk '/^default/ { print $5 }') | grep -oP '(?<=inet\s)\d+(\.\d+){3}'
```

sapkra:
I'm fine with this solution. I will apply it to my PR. Thank you for testing it out.

slauth:
I'm experiencing simliar promblems like @jordisala1991 described [above](#issuecomment-704393876). `LOCAL_ADDRESS` doesn't resolve to the correct IP when in swarm mode / using an overlay network (tested with latest `stable-5765-1`). Some context:

```
# ip route
default via 172.18.0.1 dev eth1 
10.0.6.0/24 dev eth0 proto kernel scope link src 10.0.6.3 
172.18.0.0/16 dev eth1 proto kernel scope link src 172.18.0.4

# ip a
257: eth1@if258: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default 
    link/ether 02:42:ac:12:00:04 brd ff:ff:ff:ff:ff:ff link-netnsid 1
    inet 172.18.0.4/16 brd 172.18.255.255 scope global eth1
       valid_lft forever preferred_lft forever
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
255: eth0@if256: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1450 qdisc noqueue state UP group default 
    link/ether 02:42:0a:00:06:03 brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 10.0.6.3/24 brd 10.0.6.255 scope global eth0
       valid_lft forever preferred_lft forever

# cat /etc/hosts
127.0.0.1	localhost
::1	localhost ip6-localhost ip6-loopback
fe00::0	ip6-localnet
ff00::0	ip6-mcastprefix
ff02::1	ip6-allnodes
ff02::2	ip6-allrouters
10.0.6.3	34abe86e5cec

# ip addr show dev "$(ip route|awk '/^default/ { print $5 }')" | grep -oP '(?<=inet\s)\d+(\.\d+){3}'
172.18.0.4

# hostname -i
10.0.6.3
```

My current workaround is building a custom JVB Docker image with a modified `10-config` script that simply uses `hostname -i` as `LOCAL_ADDRESS`. Any ideas how to improve this?

prayagsingh:
@slauth try setting `JVB_WS_SERVER_ID: jvb1`  with jitsi image and 
```
      networks:
            jitsi:
              aliases:
                - jvb1
```

slauth:
@prayagsingh that wouldn't work with more then one JVB, would it?

prayagsingh:
> @prayagsingh that wouldn't work with more then one JVB, would it?

You have to change the Ws_server_id and the network part for every jvb.

slauth:
> You have to change the Ws_server_id and the network part for every jvb.

This approach would prevent me from easily scaling the JVBs however.

Right now I have deployed the JVBs as global service. Scaling is now as easy as simply adding more nodes to the swarm.

prayagsingh:
> > You have to change the Ws_server_id and the network part for every jvb.
> 
> This approach would prevent me from easily scaling the JVBs however.
> 
> Right now I have deployed the JVBs as global service. Scaling is now as easy as simply adding more nodes to the swarm.

How are you adding new nodes like on what parameters (CPU or Memory) and Is this process automated?

slauth:
> How are you adding new nodes like on what parameters (CPU or Memory) and Is this process automated?

There is little to no automation around this (yet). The idea is to start simple with having alerts on CPU and bandwidth usage, then manually adding or removing JVB nodes as required.

---

One thing I noticed while trying out your suggestion… my setup is currently only working when I **not** define `DOCKER_HOST_ADDRESS` (i.e. letting the JVB determine its public address via STUN). When I put the public IP into `DOCKER_HOST_ADDRESS` there is no more A/V for conferences with 3+ participants :confused:

Nginx access log of the `web` container reveals an HTTP 403 for `GET /colibri-ws/jvb1/…` in this case:

```
jitsi_web.1.kgnlyuafou07@meet.example.com    | 10.0.2.28 - - [28/Apr/2021:13:14:14 +0200] "GET /colibri-ws/jvb1/e5a759fb6a5c38c8/a36bb228?pwd=3v6knbj0ji7nshqvq4hf2d9bks HTTP/1.1" 101 5449 "-" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.93 Safari/537.36"
jitsi_web.1.kgnlyuafou07@meet.example.com    | 10.0.2.28 - - [28/Apr/2021:13:14:16 +0200] "GET /colibri-ws/jvb1/e5a759fb6a5c38c8/a36bb228?pwd=3v6knbj0ji7nshqvq4hf2d9bks HTTP/1.1" 403 304 "-" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.93 Safari/537.36"
```
I don't see those requests at all when using STUN. Any idea why is that?

prayagsingh:
@slauth for my setup I never enabled `DOCKER_HOST_ADDRESS` hence never faced this issue. AFAIK `DOCKER_HOST_ADDRESS` is required when the setup is running behind NAT. 

Is setting `hostname -i` as `LOCAL_ADDRESS`  in `10-config` working fine with/without `DOCKER_HOST_ADDRESS`? TBH I like the way you are managing it right now. I'm OK with creating my own docker image for jvb. 

Another way is to have an env variable like `ENABLE_DOCKER_SWARM_CONFIG` in `10-config` and by default set it to `false` and if someone is using `docker-swarm` for deployment then set it to `true`.

slauth:
> Is setting `hostname -i` as `LOCAL_ADDRESS` in `10-config` working fine with/without `DOCKER_HOST_ADDRESS`?

It is working fine without `DOCKER_HOST_ADDRESS` as far as I can tell. I tested with two JVBs (Docker swarm with 1 manager + 1 worker node) and saw both of them being used when hosting two conferences simultaneously.

However I'd like to have it also work with `DOCKER_HOST_ADDRESS` to support LAN-only installations as well as development environments (using Vagrant + Ansible over here). Think I'll have to dig a little deeper here… unless somebody has  a clue what could be the problem :pray: :smile:

nexuspecial:
`dig +short $HOSTNAME` command seems to be fine as `LOCAL_ADDRESS`

PR #648, opened 2020-06-16
✨ Etherpad public URL

### LINKED ISSUE #644: Support external Etherpad
Currently, we can set `ETHERPAD_URL_BASE` to enable use of Etherpad inside Jitsi, but we can only use a "_local_" Etherpad, not an external Etherpad which might already exist outside of Jitsi docker network.

I guess this could be fixed by adding a env var to define the value of `etherpad_base`:
https://github.com/jitsi/docker-jitsi-meet/blob/a7563d4f82d79cfaca6f5ff408299f7857f7cac7/web/rootfs/etc/cont-init.d/10-config#L100-L104

For instance:
* `ETHERPAD_PUBLIC_URL=https://etherpad.my.domain`

```shell
    if [[ -z "$(grep -om1 'etherpad_base:' /config/config.js)" ]]; then
        if [[ ! -z "${ETHERPAD_PUBLIC_URL}" ]]; then
            sed -i \
                -e "/enableWelcomePage/a\    etherpad_base: '${ETHERPAD_PUBLIC_URL}/p/'," \
                /config/config.js
        elif [[ ! -z "${ETHERPAD_URL_BASE}" ]]; then
            sed -i \
                -e "/enableWelcomePage/a\    etherpad_base: '${PUBLIC_URL}/etherpad/p/'," \
                /config/config.js
        fi
    fi
```

madmath03:
:sparkles: Etherpad public URL to fix #644

Signed-off-by: mathieu.brunot <mathieu.brunot@monogramm.io>

### DIFF
--- a/web/rootfs/etc/cont-init.d/10-config
+++ b/web/rootfs/etc/cont-init.d/10-config
@@ -97,10 +97,16 @@ if [[ ! -f /config/config.js ]]; then
             /config/config.js
     fi
 
-    if [[ ! -z "${ETHERPAD_URL_BASE}" && -z "$(grep -om1 'etherpad_base:' /config/config.js)" ]]; then
-        sed -i \
-            -e "/enableWelcomePage/a\    etherpad_base: '${PUBLIC_URL}/etherpad/p/'," \
-            /config/config.js
+    if [[ -z "$(grep -om1 'etherpad_base:' /config/config.js)" ]]; then
+        if [[ ! -z "${ETHERPAD_PUBLIC_URL}" ]]; then
+            sed -i \
+                -e "/enableWelcomePage/a\    etherpad_base: '${ETHERPAD_PUBLIC_URL}/p/'," \
+                /config/config.js
+        elif [[ ! -z "${ETHERPAD_URL_BASE}" ]]; then
+            sed -i \
+                -e "/enableWelcomePage/a\    etherpad_base: '${PUBLIC_URL}/etherpad/p/'," \
+                /config/config.js
+        fi
     fi
 
     if [[ $ENABLE_TRANSCRIPTIONS -eq 1 || "$ENABLE_TRANSCRIPTIONS" == "true" ]]; then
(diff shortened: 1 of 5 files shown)


web/rootfs/etc/cont-init.d/10-config
```diff
@@ -97,10 +97,16 @@ if [[ ! -f /config/config.js ]]; then
             /config/config.js
     fi
 
-    if [[ ! -z "${ETHERPAD_URL_BASE}" && -z "$(grep -om1 'etherpad_base:' /config/config.js)" ]]; then
-        sed -i \
-            -e "/enableWelcomePage/a\    etherpad_base: '${PUBLIC_URL}/etherpad/p/'," \
```
saghul, line 102: Can you please remove the /etherpad/ proxying in nginx in the case of ETHERPAD_PUBLIC_URL being specified?

madmath03: @saghul Not sure to understand... I mean, the code you pointed out is the one that's removed, not the new code 😅 

In case we use the `ETHERPAD_PUBLIC_URL`, we already do not use the nginx proxying:
```shell
        if [[ ! -z "${ETHERPAD_PUBLIC_URL}" ]]; then
            sed -i \
                -e "/enableWelcomePage/a\    etherpad_base: '${ETHERPAD_PUBLIC_URL}/p/'," \
                /config/config.js
        elif [[ ! -z "${ETHERPAD_URL_BASE}" ]]; then
```

If I misunderstood, can you make a code suggestion ?

saghul: I mean in the Nginx config sorry :-) Ah wait, it might not be needed after all! Nevermind...

madmath03: I confirm that it's not needed since it's already done.

The nginx proxying for Etherpad is only triggered for `ETHERPAD_URL_BASE`, not for `ETHERPAD_PUBLIC_URL`:
https://github.com/jitsi/docker-jitsi-meet/blob/a7563d4f82d79cfaca6f5ff408299f7857f7cac7/web/rootfs/defaults/meet.conf#L48-L52

So if Jitsi is properly configured (ie. only `ETHERPAD_PUBLIC_URL` to external Etherpad and `ETHERPAD_URL_BASE` left empty), the proxying will not be generated.

PR #660, opened 2020-06-25
WIP: add Lua script to allow Jibri to connect to rooms with password

### LINKED ISSUE #659: Live streaming/recording (jibri) does not work if room has a password
As reported in [the forum](https://community.jitsi.org/t/jibri-authenticating-if-a-room-has-a-password-added-by-moderator/18775/), Jibri cannot authenticate if the room has a password set by a moderator.

It would be great if we could include and enable the [Lua script](https://community.jitsi.org/t/jibri-authenticating-if-a-room-has-a-password-added-by-moderator/18775/9) given by @damencho in that discussion.

pmontepagano:
Fixes #659


saghul:
I think we use some allow-list plugin for this instead, right @damencho ? I think we may want to automatically configure that here.

me-vlad:
Hi, @pmontepagano, you can place your custom plugin into the .jitsi-meet-cfg/prosody/prosody-plugins-custom directory and set XMPP_MUC_MODULES variable in your .env file:
``` ini
XMPP_MUC_MODULES=jibripass
```
Dont forget to replace default domain 'recorder.meet.jitsi' to XMPP_RECORDER_DOMAIN value.

PR #665, opened 2020-06-30
Do not remove closed captions button if transcription is enabled

nshmyrev:
Even if transcriptions are enabled the button is removed currently in docker because of this simple logic mistake.


saghul:
Thanks!

PR #673, opened 2020-07-09
Use larger alphabet for generating secure passwords

dscottboggs:
This patch would generate passwords of the same length but with a larger alphabet (and hence greater entropy) than the suggested hex passwords.

### DIFF
--- a/gen-passwords.sh
+++ b/gen-passwords.sh
@@ -1,7 +1,8 @@
 #!/bin/bash
 
+# Generate a new 512-bit random key
 function generatePassword() {
-    openssl rand -hex 16
+    openssl rand -base64 64 | tr '+/' ':.' |tr -d '=\n'
 }
 
 JICOFO_COMPONENT_SECRET=$(generatePassword)


saghul:
I’d rather use the builtin OpenSSL generator than rolling our own.

dscottboggs:
The issue I have with the current script is the use of the `-hex` flag. Using the `-hex` flag limits you to 16 bits of entropy per character, meaning there's only 256 bits of entropy contained in these passwords. Granted, that's _just_ enough, but there's no reason not to gather more entropy when there's 0 overhead of doing it. Since OpenSSL is just using the system's entropy source to generate the value, I don't particularly see the advantage of using it over reading directly from that source.

Another option could be something like
```shell
openssl rand -base64 48 | tr -d '/+=' | head -c16
```
This would have roughly the same effect as my suggestion while retaining the use of only alphanumeric characters while increasing the entropy per character to 62 bits, leaving a 16-character passphrase with 992 bits of entropy. However, there are some edge cases where it could generate shorter passwords (for example, theoretically openssl could generate a 48 character base64 string of only `+`, `/`, and `=` chars, which would return a 0-length passphrase), and it is generally a bit less robust than the solution I suggested.

If you could guarantee that using `+` and `/` symbols in the passphrase won't screw up some script, then I don't see any harm in just using
```shell
openssl rand -base64 16
```
which would provide 1024 bits of entropy. I didn't suggest that initially though because I'd hate to introduce a bug into such a great platform :wink:

dscottboggs:
> If you could guarantee that using + and / symbols in the passphrase won't screw up some script

This could also be mitigated by running the `openssl rand -base64` output through a couple of `tr` commands, like

```sh
openssl rand -base64 16 | tr / - | tr + _
```

saghul:
The current proposal breaks on macOS.

dscottboggs:
My initial suggestion was based on a misunderstanding of the `openssl rand` command. I was under the impression that the numeric argument was the number of characters in the output, however,
> This command generates _num_ random bytes using a cryptographically secure pseudo random number generator (CSPRNG).

However, this still gives us 16\*8=128 bits of entropy, which is not enough. You could get away with 32\*8=256 bits, but I prefer to go a little higher to 512 bits/64 bytes. Despite my initial suggestion to use a larger alphabet being based on this incorrect assumption, a 64-byte hex password would be very long. My current suggestion uses the base64 alphabet to generate a 512-bit secure password, and strips any potentially problematic characters to be extra careful.

dscottboggs:
> The current proposal breaks on macOS.

Yes, this is the reason I didn't think of for not reading directly from the OS's source -- not all OS's are the same. Sticking with the OpenSSL tool makes sense IMO for this reason.

saghul:
> However, this still gives us 16*8=128 bits of entropy, which is not enough.

Can you elaborate on the threat model here? Note that since an attacker doesn't know if you used the provided script, wrote the password by hand or some other tool, they'd need to try all possible combinations. In addition, the attacker would need to establish a TLS connection and perform the XMPP authentication procedure to test if the password is correct, and will hit rate limits.

PR #676, opened 2020-07-12
 base/Dockerfile optimization

kaichao:
- Reduce 3 layers in the base image; 
- Free up unused space, from 93.4MB to 86.3MB.

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -9,11 +9,11 @@ COPY rootfs /
 RUN \
 	apt-dpkg-wrap apt-get update && \
 	apt-dpkg-wrap apt-get install -y apt-transport-https apt-utils ca-certificates gnupg && \
-	DEBIAN_FRONTEND=noninteractive apt-get install -y wget && \
+	apt-dpkg-wrap apt-get install -y wget && \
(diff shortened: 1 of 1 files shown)


base/Dockerfile
```diff
@@ -4,19 +4,16 @@ ARG JITSI_RELEASE=stable
...
 COPY rootfs /
 
 RUN \
-	tar xfz /tmp/s6-overlay.tar.gz -C / && \
-	rm -f /tmp/*.tar.gz && \
 	apt-dpkg-wrap apt-get update && \
 	apt-dpkg-wrap apt-get install -y apt-transport-https apt-utils ca-certificates gnupg && \
-	apt-key add /tmp/jitsi.key && \
-	rm -f /tmp/jitsi.key && \
+	DEBIAN_FRONTEND=noninteractive apt-get install -y wget && \
```
saghul, line 12: Can you please use apt-dpkg-wrap instead? https://github.com/jitsi/docker-jitsi-meet/blob/master/base/rootfs/usr/bin/apt-dpkg-wrap

kaichao: Have done. 
To be consistent, 'apt-get remove' has also been modified.

PR #699, opened 2020-07-29
Disable Traefik for containers that do not need it

### LINKED ISSUE #698: Unnecessary endpoints in traefik example
The `jicofo` and `prosody` containers do not need outside connectivity. So in `docker-compose.yml` should disable Traefik for those containers.

bluikko:
Jicofo and prosody do not need outside connectivity so disabled traefik on them.

Closes #698


viktor-c:
I think you don't have to disable traefik on that, because these containers don't run on traefik network, so they won't be picked up.

saghul:
Closing since Trefik support was moved to jitsi-contrib.

Issue #691, opened 2020-07-22
Jicofo health checks fail since upgrade to stable-4857


cedricroijakkers:
I've just upgraded to the latest container version stable-4857, and now the Jicofo container fails it's health checks with the following error:

```
Jicofo 2020-07-22 13:22:26.803 SEVERE: [32] org.jitsi.impl.protocol.xmpp.ChatRoomImpl.log() No presence packet obtained yet
Jicofo 2020-07-22 13:22:26.803 INFO: [32] org.jitsi.jicofo.FocusManager.log() Exception while trying to start the conference
net.java.sip.communicator.service.protocol.OperationFailedException: Failed to join the room
	at org.jitsi.impl.protocol.xmpp.ChatRoomImpl.joinAs(ChatRoomImpl.java:298)
	at org.jitsi.impl.protocol.xmpp.ChatRoomImpl.join(ChatRoomImpl.java:209)
	at org.jitsi.jicofo.JitsiMeetConferenceImpl.joinTheRoom(JitsiMeetConferenceImpl.java:581)
	at org.jitsi.jicofo.JitsiMeetConferenceImpl.start(JitsiMeetConferenceImpl.java:404)
	at org.jitsi.jicofo.FocusManager.conferenceRequest(FocusManager.java:477)
	at org.jitsi.jicofo.health.Health.check(Health.java:200)
	at org.jitsi.jicofo.health.Health.performCheck(Health.java:133)
	at org.jitsi.health.AbstractHealthCheckService.run(AbstractHealthCheckService.kt:155)
	at org.jitsi.utils.concurrent.RecurringRunnableExecutor.run(RecurringRunnableExecutor.java:216)
	at org.jitsi.utils.concurrent.RecurringRunnableExecutor.runInThread(RecurringRunnableExecutor.java:292)
	at org.jitsi.utils.concurrent.RecurringRunnableExecutor.access$000(RecurringRunnableExecutor.java:36)
	at org.jitsi.utils.concurrent.RecurringRunnableExecutor$1.run(RecurringRunnableExecutor.java:328)
Caused by: org.jivesoftware.smack.XMPPException$XMPPErrorException: XMPP error reply received from conference.meet.jitsi: XMPPError: not-allowed - cancel
	at org.jivesoftware.smack.XMPPException$XMPPErrorException.ifHasErrorThenThrow(XMPPException.java:132)
[...]

PR #706, opened 2020-07-30
Configuration fix for issue #691.

cedricroijakkers:
Set `org.jitsi.jicofo.XMPP_MUC_COMPONENT_PREFIX` to `muc` to resolve issues between jicofo and jvb.

### DIFF
--- a/jicofo/rootfs/defaults/sip-communicator.properties
+++ b/jicofo/rootfs/defaults/sip-communicator.properties
@@ -1,5 +1,6 @@
 org.jitsi.jicofo.ALWAYS_TRUST_MODE_ENABLED=true
 org.jitsi.jicofo.BRIDGE_MUC={{ .Env.JVB_BREWERY_MUC }}@{{ .Env.XMPP_INTERNAL_MUC_DOMAIN }}
+org.jitsi.jicofo.XMPP_MUC_COMPONENT_PREFIX=muc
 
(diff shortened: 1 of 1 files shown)


jicofo/rootfs/defaults/sip-communicator.properties
```diff
@@ -1,5 +1,6 @@
 org.jitsi.jicofo.ALWAYS_TRUST_MODE_ENABLED=true
 org.jitsi.jicofo.BRIDGE_MUC={{ .Env.JVB_BREWERY_MUC }}@{{ .Env.XMPP_INTERNAL_MUC_DOMAIN }}
+org.jitsi.jicofo.XMPP_MUC_COMPONENT_PREFIX=muc
```
saghul, line 3: Let's not hardcode it, since it can be changed by the user: https://github.com/jitsi/docker-jitsi-meet/blob/2a0120de8b4b274bcb22c98426705c54e570edd3/env.example#L190

You should use frep to get the first element in that domain.

saghul:
Thank you!

sapkra:
Since you have added the variable `XMPP_MUC_DOMAIN` to the jicofo container, you must also pass it to it in the `docker-compose.yml`.

saghul:
Yep, forgot to add that, oops.

Issue #700, opened 2020-07-29
lobby not available when using authentication?


wivaku:
When using authentication (in my case: internal) there's no option to activate the lobby. When clicking on the shield icon the only option is to add a password.

When **not** using authentication the lobby functionality works fine, although I have to enable it for each session by clicking the shield. 
Using ENABLE_AUTH=0, ENABLE_LOBBY=1, ENABLE_GUESTS=0 or =1

wivaku:
Update, it seems that to use authentication + lobby you have to disable guests.
So:
ENABLE_AUTH=1, ENABLE_LOBBY=1, ENABLE_GUESTS=0

ChrisPortman:
The issue is that the lobby config is always applied to the "main" virtualhost.  When auth is enabled, it must be applied to the guest virtual host.

Issue #697, opened 2020-07-28
Using lobby feature with ldap


prayagsingh:
I have couple of questions related to lobby.

1. Is it possible to use lobby with ldap? 
2. Is there any kind of relation between `lobby` and `enable_guest` like do I need to set `enable_guest` to true for lobby feature?

**My Observations**
I tried to use `lobby` feature with `ldap`. Let's say the meeting url is `meet.example.in/test`. Once I enabled the `lobby` and when other user tried to join the meeting then it lands to a new meeting on a same url i.e `meet.example.in/test`.

***Few points to note here***
1.  Every user is authenticated using ldap. 
2. When second user joins then he/she isn't getting options like `request to join` or `enter password`. 
3. `ENABLE_GUEST` is set to false because of bug #538

JacBra:
I want to enable Lobby too in combination with authentication (using ldap).

ENABLE_AUTH=0, ENABLE_LOBBY=1, ENABLE_GUESTS=1
--> this works

ENABLE_AUTH=1, ENABLE_LOBBY=1, ENABLE_GUESTS=0
--> this works, but then every user has to authenticate -> undesired for my use case

ENABLE_AUTH=1, ENABLE_LOBBY=1, ENABLE_GUESTS=1
--> this is the situation I want (only first to enter needs to authenticate, others are guests that get into the lobby). But this scenario does *not* work: the Lobby option is not visible in the gui when I click the Shield-icon.

Does anyone know how to get this working?

prayagsingh:
@JacBra In first case `ENABLE_AUTH=0, ENABLE_LOBBY=1, ENABLE_GUESTS=1`, If you are disabling `AUTH` then who is going to be the `moderator`?  

I tried this `ENABLE_AUTH=1, ENABLE_LOBBY=1, ENABLE_GUESTS=0` case but In my case second user is unable too see the first person on the sam url like `meet.example.in/test`.

`ENABLE_AUTH=1, ENABLE_LOBBY=1, ENABLE_GUESTS=1` : you don't need to enable guest mode for lobby.  

here the best case is seconds one. If a first person started the meeting and enables lobby then it shouldn't ask for password for second user but it is not happening as of now.

prayagsingh:
@JacBra Are you using `docker-swarm`?

JacBra:
@prayagsingh: no I don't use docker-swarm.

**first case**
[...]

PR #712, opened 2020-08-13
Add the lobby config to the guest.$XMPP_DOMAIN

ChrisPortman:
If the lobby is enabled and ENABLE_AUTH and ENABLE_GUESTS is true, and
therefore the guest.$XMPP_DOMAIN domain is to be defined, the lobby config
should be placed in the guest domain.

Addresses
 - #700 
 - #697

### DIFF
--- a/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
+++ b/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
@@ -7,6 +7,7 @@ plugin_paths = { "/prosody-plugins/", "/prosody-plugins-custom" }
 http_default_host = "{{ .Env.XMPP_DOMAIN }}"
 
 {{ $ENABLE_AUTH := .Env.ENABLE_AUTH | default "0" | toBool }}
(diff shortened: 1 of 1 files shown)


prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -22,6 +22,7 @@ asap_accepted_issuers = { "{{ join "\",\"" (splitList "," .Env.JWT_ACCEPTED_ISSU
 asap_accepted_audiences = { "{{ join "\",\"" (splitList "," .Env.JWT_ACCEPTED_AUDIENCES) }}" }
 {{ end }}
 
+{{ $enable_guest_domain := and $ENABLE_AUTH (.Env.ENABLE_GUESTS | default "0" | toBool)}}
```
saghul, line 25: can you please make the variable name uppercase and move it above with the other variable declarations? LGTM otherwise!

ChrisPortman: I did originally have it uppercase, but then thought lower case would differentiate it from ENV variables as an internal one?

But happy to change it if you can confirm it's what you want given the above.

Thanks

saghul:
Cheers!

ChrisPortman:
Thank you!

PR #718, opened 2020-08-20
Store secrets in separate .env file

0x326:
This allows a downstream-user to keep `.env` under some version control without worrying about password leakage.


saghul:
I like the idea!

0x326:
@saghul Ok, it should be ready now

PR #730, opened 2020-09-01
Update env.example

iambouali:
The other way it doesn't work.

### DIFF
--- a/env.example
+++ b/env.example
@@ -31,7 +31,7 @@ JIBRI_XMPP_PASSWORD=
 #
 
 # Directory where all configuration will be stored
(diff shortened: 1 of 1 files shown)


goacid:
Your change do not point to originally local home path, but now to the current path (which depend on each user)
No sur it's a good idea.
BTW this param is often updated by users before deploying Jitsi.

iambouali:
I think it is better, because it works in all cases. In the other way, only if you have it in the home.

goacid:
Well how it is done works also in all case.
Can you explain in which case ~/ doesn't exist ?

PR #748, opened 2020-09-24
Jvb new config

bbaldino:
This PR generates a new-style jvb config file (`jvb.conf`) and ports all supported properties from `sip-communicator.properties` over to it.  There is one property in ice4j which has not been ported yet which will have to stay, for now, in `sip-communicator.properties`--this is not ideal from a confusion perspective, but the config value itself is fine and it will still be observed.

### DIFF
--- a/jvb/rootfs/defaults/jvb.conf
+++ b/jvb/rootfs/defaults/jvb.conf
@@ -1,0 +1,61 @@
+{{ $JVB_TCP_PORT := .Env.JVB_TCP_PORT | default "4443" }}
+{{ $JVB_TCP_MAPPED_PORT := .Env.JVB_TCP_MAPPED_PORT | default $JVB_TCP_PORT }}
+{{ $PUBLIC_URL_DOMAIN := .Env.PUBLIC_URL | default "https://localhost:8443" | trimPrefix "https://" | trimSuffix "/" -}}
+{{ $WS_DOMAIN := .Env.JVB_WS_DOMAIN | default $PUBLIC_URL_DOMAIN -}}
+{{ $WS_SERVER_ID := .Env.JVB_WS_SERVER_ID | default .Env.LOCAL_ADDRESS -}}
+
+videobridge {
+    ice {
+        udp {
+            port = {{ .Env.JVB_PORT }}
+        {
+        tcp {
+            enabled = {{ not .Env.JVB_TCP_HARVESTER_DISABLED }}
+            port = {{ .Env.JVB_TCP_PORT }}
+{{ if not (eq $JVB_TCP_PORT $JVB_TCP_MAPPED_PORT) }}
+            mapped-port = {{ $JVB_TCP_MAPPED_PORT }}
+{{ end }}
+        }
+    }
+    apis {
+        xmpp-client {
+            configs {
+                shard {
+                    HOSTNAME = "{{ .Env.XMPP_SERVER }}"
+                    DOMAIN = "{{ .Env.XMPP_AUTH_DOMAIN }}"
+                    USERNAME = "{{ .Env.JVB_AUTH_USER }}"
+                    PASSWORD = "{{ .Env.JVB_AUTH_PASSWORD }}"
+                    MUC_JIDS = "{{ .Env.JVB_BREWERY_MUC }}@{{ .Env.XMPP_INTERNAL_MUC_DOMAIN }}"
+                    MUC_NICKNAME = "{{ .Env.HOSTNAME }}"
+                    DISABLE_CERTIFICATE_VERIFICATION = true
+                }
+            }
+        }
+    }
+    stats {
+        enabled = true
+    }
+    websockets {
+        enabled = true
+        domain = "{{ $WS_DOMAIN }}"
+        tls = true
+        server-id = "{{ $WS_SERVER_ID }}"
+    }
+    http-servers {
+        private {
+            host = 0.0.0.0
+        }
+    }
+}
+
+ice4j {
+    harvest {
+        mapping {
+            stun {
+{{ if .Env.JVB_STUN_SERVERS }}
+                addresses = [ "{{ join "\",\"" (splitList "," .Env.JVB_STUN_SERVERS) }}" ]
+{{ end }}
+            }
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


jvb/rootfs/etc/services.d/jvb/run
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/with-contenv bash
 
-JAVA_SYS_PROPS="-Dnet.java.sip.communicator.SC_HOME_DIR_LOCATION=/ -Dnet.java.sip.communicator.SC_HOME_DIR_NAME=config -Djava.util.logging.config.file=/config/logging.properties"
+JAVA_SYS_PROPS="-Dnet.java.sip.communicator.SC_HOME_DIR_LOCATION=/ -Dnet.java.sip.communicator.SC_HOME_DIR_NAME=config -Djava.util.logging.config.file=/config/logging.properties -Dconfig.file=/config/jvb.conf"
```
saghul, line 3: How about the logging properties?

jvb/rootfs/defaults/sip-communicator.properties
```diff
@@ -1,23 +1,4 @@
-org.jitsi.videobridge.SINGLE_PORT_HARVESTER_PORT={{ .Env.JVB_PORT }}
-org.jitsi.videobridge.DISABLE_TCP_HARVESTER={{ .Env.JVB_TCP_HARVESTER_DISABLED }}
-org.jitsi.videobridge.TCP_HARVESTER_PORT={{ .Env.JVB_TCP_PORT }}
+# TODO(brian): not ported in ice4j to new config yet
```
saghul, line 1: Ah I see, so this is a hybrid for a bit?

saghul (commented):
Made a quick pass!

jvb/rootfs/defaults/sip-communicator.properties
bgrozev: Looks like this is the last property in the file. I've ported it in jitsi/ice4j#215

saghul: Nice, thank you!

jvb/rootfs/etc/cont-init.d/10-config
```diff
@@ -13,6 +13,7 @@ fi
 
 if [[ ! -f /config/sip-communicator.properties ]]; then
```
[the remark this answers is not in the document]
bbaldino: I wasn't sure if we wanted to do the change to always regenerate the config file in this same PR or not.  I can if you think it'd be best?

jvb/rootfs/etc/services.d/jvb/run
bbaldino: Nothing has changed with the logging properties file.

jvb/rootfs/defaults/sip-communicator.properties
bbaldino: I can update this once that gets merged and we update JVB.  Any concern for the case where docker has to handle an older JVB version before this change?

dawidmalina:
Is this change gonna be compatible with websockets - currently enabled by default for new installs - https://github.com/jitsi/jitsi-meet/pull/7781

bbaldino:
> Is this change gonna be compatible with websockets - currently enabled by default for new installs - [jitsi/jitsi-meet#7781](https://github.com/jitsi/jitsi-meet/pull/7781)

I'll do this in a separate PR once this is in.

jvb/rootfs/etc/cont-init.d/10-config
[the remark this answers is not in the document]
bbaldino: I removed this so we always regenerate it.

jvb/rootfs/defaults/jvb.conf
```diff
@@ -0,0 +1,52 @@
...
+    }
+    apis {
+        xmpp-client {
+            configs {
+                shard {
+                    HOSTNAME = {{ .Env.XMPP_SERVER }}
+                    DOMAIN = {{ .Env.XMPP_AUTH_DOMAIN }}
+                    USERNAME = {{ .Env.JVB_AUTH_USER }}
+                    PASSWORD = {{ .Env.JVB_AUTH_PASSWORD }}
+                    MUC_JIDS = {{ .Env.JVB_BREWERY_MUC }}@{{ .Env.XMPP_INTERNAL_MUC_DOMAIN }}
```
saghul, line 24: looks like we need quotes here

jvb/rootfs/defaults/jvb.conf
```diff
@@ -0,0 +1,52 @@
...
+        ]
+    }
+}
+
+ice4j {
+    harvest {
+        mapping {
+            stun {
+{{ if .Env.JVB_STUN_SERVERS }}
+                [ "{{ join "\",\"" (splitList "," .Env.JVB_STUN_SERVERS) }}" ]
```
saghul, line 47: it complains about the syntax here.

```
Caused by: com.typesafe.config.ConfigException$Parse: /config/jvb.conf: 43: expecting a close parentheses ')' here, not: '['
```

Here is the generated section:

```
ice4j {
    harvest {
        mapping {
            stun {

                [ "meet-jit-si-turnrelay.jitsi.net:443" ]

            }   
        }   
    }   
}   
```

bbaldino: 🤦  Ah--the error is confusing but I see it now.  It's missing the field name before the array.

bbaldino:
This relies on an iec4j config param being ported, which is done in the bridge in 2.1-358-g3abd64a9, so we should wait until that goes stable before merging this.

bbaldino:
@saghul jitsi-videobridge2_2.1-376-g9f12bfe2 has now been made stable, so we can merge this (after I take care of the conflict which has arisen)

saghul:
@bbaldino I took this PR for a spin but the bridge seems to fail somehow and it doesn't register with jicofo. This might be the cause:

~~~
jvb_1      | Exception in thread "main" java.lang.IllegalArgumentException: Parameter specified as non-null is null: method org.jitsi.videobridge.rest.binders.ServiceBinder.<init>, parameter statsCollector
jvb_1      | 	at org.jitsi.videobridge.rest.binders.ServiceBinder.<init>(ServiceBinder.kt)
jvb_1      | 	at org.jitsi.videobridge.rest.root.Application.<init>(Application.java:37)
jvb_1      | 	at org.jitsi.videobridge.MainKt.main(Main.kt:129)
~~~

PR #751, opened 2020-09-29
add org.jitsi.jicofo.SHORT_ID

goacid:
https://github.com/jitsi/jicofo/pull/553


saghul:
Shouldn't this be part of the octo PR?

goacid:
I think it is a jicofo variable, which is use for Octo, but maybe not only.
In fact I don't really know.
I can we can merge it into the octo branch, just let me know what you prefer

saghul:
I'd rather keep it with the octo PR please.

PR #762, opened 2020-10-06
feat(config): configure startAudioOnly using environment variable

NullIsNot0:
Add an easy option to start conferences in audio only mode. Introduced new environment variable `START_AUDIO_ONLY` which can be set to `true` (default value is `false`).


NullIsNot0:
This is one of the option I'm missing and can't change after PR https://github.com/jitsi/docker-jitsi-meet/pull/619 was merged.

saghul (approved):

PR #782, opened 2020-10-12
Secondary IP: Allows running Jitsi around corporate firewalls

miggland:
A rather simple way to run Jitsi with less problems caused by corporate firewalls is described by @dammsugare here:
https://community.jitsi.org/t/corporate-firewalls-use-port-443-for-videobridge-with-jitsi-docker/28285/5

This allows easily running JVB over port 443, but requries a secondary IP address for the server to be defined.

Most settings @dammsugare describes can be implemented in the environment, but it requires changing a settings file in jvb. With this commit, an environment variable called DOCKER_HOST_ADDRESS2 can be used instead. If this is not defined, the behaviour returns to the default, using DOCKER_HOST_ADDRESS

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -210,6 +210,7 @@ services:
             - JVB_PORT
             - JVB_TCP_HARVESTER_DISABLED
             - JVB_TCP_PORT
+            - JVB_TCP_MAPPED_PORT
             - JVB_STUN_SERVERS
             - JVB_ENABLE_APIS
             - PUBLIC_URL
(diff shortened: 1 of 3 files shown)


saghul:
I don't understand how this helps at all. The only place where DOCKER_HOST_ADDRESS is used is right there, on the JVB. No other service uses it. So if you just set it to what you want you can use port 443 for the JVB.

miggland:
Sometimes the simple solutions are so obvious that they are easy to miss.. *facepalm*
Sorry about that, I'll go hide in a corner now.

saghul:
No worries mate!

PR #786, opened 2020-10-13
Don't listen API calls from all hosts when APIs are not enabled

NullIsNot0:
Don't listen API calls from all hosts when APIs are not enabled (which is default behaviour) within `.env` variable `JVB_ENABLE_APIS`


saghul:
What problem are you trying to solve? Here the host is the container itself so if no APIs are no enabled no harm is done. Am I wrong?

NullIsNot0:
Ok, I think we can safely ignore this PR.

PR #788, opened 2020-10-14
feat(jvb): make colibri websocket endpoints dynamic for multiple jvbs

### LINKED ISSUE #756: Improvements for WebSocket based bridge channels
This line will not work for other deployments because `jvb.meet.jitsi` will not be available in their network environments. We should consider to make this dynamic/configurable by an env var.
https://github.com/jitsi/docker-jitsi-meet/blob/b731c6033f7940bd2c5b526567ebc0f889754405/web/rootfs/defaults/meet.conf#L39

The second problem would be if someone wants to launch more that one JVB, then it is currently not possible because the `COLIBRI_WS_SERVER_ID` is not set and the endpoints will be missing because it's currently only configured for the `default-id`.

But it definitely would be possible to work around these problems by using an own proxy but for this the `COLIBRI_WS_SERVER_ID` must still be configurable.

sapkra:
Closes #756 
Relates to #630

### DIFF
--- a/jvb/rootfs/etc/cont-init.d/10-config
+++ b/jvb/rootfs/etc/cont-init.d/10-config
@@ -1,5 +1,7 @@
 #!/usr/bin/with-contenv bash
 
+export LOCAL_ADDRESS=$(ip addr show dev "$(ip route|awk '/^default/ { print $5 }')" | grep -oP '(?<=inet\s)\d+(\.\d+){3}')
+
 if [[ -z $JVB_AUTH_PASSWORD ]]; then
     echo 'FATAL ERROR: JVB auth password must be set'
     exit 1
--- a/web/rootfs/defaults/meet.conf
+++ b/web/rootfs/defaults/meet.conf
@@ -35,8 +35,8 @@ location ~ ^/(libs|css|static|images|fonts|lang|sounds|connection_optimization|.
 }
 
 # colibri (JVB) websockets
-location ~ ^/colibri-ws/default-id/(.*) {
-    proxy_pass http://jvb.meet.jitsi:9090/colibri-ws/default-id/$1$is_args$args;
+location ~ ^/colibri-ws/(.+)/(.*) {
+    proxy_pass http://$1:9090/colibri-ws/$1/$2$is_args$args;
     proxy_http_version 1.1;
     proxy_set_header Upgrade $http_upgrade;
     proxy_set_header Connection "upgrade";
--- a/jvb/rootfs/defaults/sip-communicator.properties
+++ b/jvb/rootfs/defaults/sip-communicator.properties
@@ -4,6 +4,10 @@ org.jitsi.videobridge.TCP_HARVESTER_PORT={{ .Env.JVB_TCP_PORT }}
 {{ if .Env.JVB_STUN_SERVERS }}
 org.ice4j.ice.harvest.STUN_MAPPING_HARVESTER_ADDRESSES={{ .Env.JVB_STUN_SERVERS }}
 {{ end }}
+{{ if .Env.DOCKER_HOST_ADDRESS }}
+org.ice4j.ice.harvest.NAT_HARVESTER_LOCAL_ADDRESS=.Env.LOCAL_ADDRESS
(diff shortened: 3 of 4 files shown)


web/rootfs/defaults/meet.conf
```diff
@@ -35,8 +35,8 @@ location ~ ^/(libs|css|static|images|fonts|lang|sounds|connection_optimization|.
 }
 
 # colibri (JVB) websockets
-location ~ ^/colibri-ws/default-id/(.*) {
-    proxy_pass http://jvb.meet.jitsi:9090/colibri-ws/default-id/$1$is_args$args;
+location ~ ^/colibri-ws/([0-9]+\.[0-9]+\.[0-9]+\.[0-9]+)/(.*) {
```
saghul, line 38: Since this container might be used standalone, let's have a catch-all here. It would be possible to use domains and connect over the public internet for example.

jvb/rootfs/etc/cont-init.d/10-config
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/with-contenv bash
 
+export LOCAL_ADDRESS=$(ip route get "$(grep "$HOSTNAME" /etc/hosts|awk '{print $1}')" | head -n1 | cut -d " " -f6)
```
saghul, line 3: This changes how we compute it. If DOCKER_HOST_ADDRESS is given, we should use the route for that IP instead, as we were doing before, I'd say.

saghul:
Left a comple of comments, thanks for getting this started!

jvb/rootfs/etc/cont-init.d/10-config
saghul: I just saw https://github.com/jitsi/docker-jitsi-meet/pull/630#issuecomment-708665793, so you can disregard my previous comment.

saghul (approved):
LGTM, is there anything missing here?

sapkra:
I still want to make sure that the new local IP solution is still working for @debfx

jvb/rootfs/etc/cont-init.d/10-config
sapkra: We have found a solution which works for both of us.

saghul:
Landed with a couple of fixups in https://github.com/jitsi/docker-jitsi-meet/commit/b2779263328fb1183b450af77d1e267c8b0c055a

debfx:
Hi @sapkra, I'm curious why you chose `LOCAL_ADDRESS` as the default for `JVB_WS_SERVER_ID`. Using `jvb.meet.jitsi` would seem the first choice to me. It has been used before and doesn't require any IP parsing magic.

This actually causes some issues on my admittedly not at all default setup :) 

In my setup there is the internal `meet.jitsi` network and a `bridge_internet` network that's exclusively used for connections to/from the internet. `LOCAL_ADDRESS` points to the IP on `bridge_internet` so now the web container tries to talk to jvb via  `bridge_internet` instead of `meet.jitsi` which is not allowed.

Now I'm not claiming my setup needs to be supported out of the box. Just a data point for why changing the default might be a good idea.

sapkra:
The reasons why we choose this as default:
1. If you have multiple JVBs you cannot set the same server id for all of them. The traffic has to be routed to the correct JVB instance.
2. Not all docker environments have a networking concept like docker-compose. For example I'm using kubernetes and for this environment it would be extremely hard to implement dynamic server IDs. The best options was to use something docker native, which are in this case the IPs of each container. They are unique and it will work in all docker environments.

Do you have an idea for changing the command behind `LOCAL_ADDRESS` again, so that everybody is happy? Maybe we have to introduce two variables to cover both usecases.

debfx:
I see, makes sense.

The only possibility I can come up with to set a `LOCAL_ADDRESS_INTERNAL`:

* Resolve `$XMPP_DOMAIN`
* If it is in the subnet of an interface, use the IP of that interface
* If not, fall back to `LOCAL_ADDRESS`

So not very pretty, especially if implemented in shell. I'm not sure if this is the way to go.

sapkra:
I think it doesn't really matter how pretty the code is in this case. If you could create a PR I would really appreciate it. I currently don't have a test environment ready nor time to do it.

PR #798, opened 2020-10-24
Updated link for running behind NAT

limitedAtonement:
There is no mention of running behind a NAT in README.md.

### DIFF
--- a/env.example
+++ b/env.example
@@ -48,7 +48,8 @@ TZ=UTC
 #PUBLIC_URL=https://meet.example.com
 
 # IP address of the Docker host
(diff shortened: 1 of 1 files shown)


limitedAtonement:
Oh yeah, it would be good to know if a hostname could be used instead of an IP address. If so, I can adjust the comment here in `env.example` to indicate that possibility.

saghul (approved):

saghul:
> Oh yeah, it would be good to know if a hostname could be used instead of an IP address. If so, I can adjust the comment here in `env.example` to indicate that possibility.

This is alas not possible.

PR #812, opened 2020-11-05
change Etherpad skin to "no-skin"

schnusch:
This commit changes the Etherpad skin back to the default skin before version
1.8.0. The old skin is better suited for embedding.

This commit effectively reverts commit 70bc71c "skins: make "colibris" the
default skin for new installations" to etherpad-lite.

### DIFF
--- a/etherpad/rootfs/defaults/settings.json
+++ b/etherpad/rootfs/defaults/settings.json
@@ -19,6 +19,19 @@
   "favicon": "favicon.ico",
 
   /*
+   * Skin name.
+   *
+   * Its value has to be an existing directory under src/static/skins.
+   * You can write your own, or use one of the included ones:
+   *
+   * - "no-skin":  an empty skin (default). This yields the unmodified,
+   *               traditional Etherpad theme.
+   * - "colibris": the new experimental skin (since Etherpad 1.8), candidate to
+   *               become the default in Etherpad 2.0
+   */
+  "skinName": "no-skin",
+
+  /*
    * IP and port which etherpad should bind at
    */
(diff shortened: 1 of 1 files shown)


saghul:
Can you please share a before / after screenshot?

schnusch:
before \
*before*

after \
*after*

saghul:
Why is the former one better for embedding? I find the new one more aesthetically pleasing to be honest!

schnusch:
The colibris skin wastes adds so much padding that almost half of the screen is wasted. The font of the colibris skin may be more pleasing, but it seems no-skin just uses the browser's defaults.

Especially when using Etherpad while in a jitsi-meet session an Etherpad will be opened in another window and the screen area will be shared between jitsi-meet and Etherpad and in therefore smaller window sizes a higher information density, due to less wasted space for padding, is preferrable.

sapkra:
If @saghul merges my PR https://github.com/jitsi/docker-jitsi-meet/pull/341 you could use the official etherpad env vars to configure your skin of choice.

@schnusch In the meanwhile you could just apply my changes to your deployment.

Here you can find the skin env vars:
https://github.com/ether/etherpad-lite/blob/develop/doc/docker.md#skins


Additionally we can use the new skin with full width:
> For the editor container, you can also make it full width by adding full-width-editor variant (by default editor is rendered as a page, with a max-width of 900px).

sapkra:
@schnusch Can you checkout the latest release?

schnusch:
This solution is more elegant and I'm happy to see `full-width-editor` as the default in commit 42a4346.

Issue #758, opened 2020-10-03
jvb colibri is running on private (127.0.0.1) server and is no longer accessible


ccoenen:
With the update to stable-5076, I can't use the colibri stats any more within my docker-setup. I believe the reason is [this change from jvb, where the `/colibri` endpoint is now only serverd on the loopback address](https://github.com/jitsi/jitsi-videobridge/commit/01bde07058fdb347b2169f92282be3b6158c3e87). This is of course a sensible choice for most setups, but within docker this means you can't access it any more at all.

from within that container, I can poll this:
```sh
docker-compose exec jvb curl http://localhost:8080/colibri/stats
  {"inactive_endpoints":0,"inactive_conferences":0,"total_ice_succeeded # shortened...
```
from the outside, not so much:
```sh
curl http://localhost:8080/colibri/stats
  curl: (56) Recv failure: Connection reset by peer
```

To make this 100% clear: this worked in the previous stable. I did set colibri and rest via `JVB_ENABLE_APIS=rest,colibri` in  my `.env`. And I also set up the additional port in my `docker-compose.yml`. This is a regression stemming from (probably) that mentioned jvb change.

According to [this config documentation](https://github.com/jitsi/jitsi-videobridge/blob/master/doc/rest.md#configuration) one should be able to set `org.jitsi.videobridge.rest.private.jetty.host` - if that was set to `0.0.0.0` that should bring back the previous behaviour, or so I thought. Putting this line into `jvb/rootfs/defaults/sip-communicator.properties` did _not_ fix the problem for me, sadly.

Can someone help me get the colibri stats back?

ccoenen:
So, my config explained as above did not apply. So I went looking where the configuration is read:

https://github.com/jitsi/jitsi-videobridge/blob/25c5f8db14a300b6e1ef83e7c7214f79dbf7b342/jvb/src/main/kotlin/org/jitsi/videobridge/Main.kt#L123-L140

combined with

https://github.com/jitsi/jicoco/blob/55b408182a329a1d8624f3384902e145aa3710d3/jicoco/src/main/kotlin/org/jitsi/rest/JettyBundleActivatorConfig.kt#L33-L45

gives us this string for legacy config:
`org.jitsi.videobridge.rest.private.jetty.host` as seen above. Odd. Should work.

[...]

PR #822, opened 2020-11-17
Fix websocket

pierreozoux:
We noticed that websocket were broken with this config.

We believe that the translation from old config to new config in [this commit](https://github.com/jitsi/docker-jitsi-meet/commit/130eb551a7666a9a143343e4ac7e707e3d6c6c02#diff-6f9b0c296a7ff789834c1414349f64f956f2b64df6e405fd45f62c6a11edad3fL34) didn't translate it well.

We are actually wondering what does private stands for in this context.

With this change, we can confirm that wss works again as expected.

### DIFF
--- a/jvb/rootfs/defaults/jvb.conf
+++ b/jvb/rootfs/defaults/jvb.conf
@@ -42,8 +42,9 @@ videobridge {
         server-id = "{{ $WS_SERVER_ID }}"
     }
     http-servers {
-        private {
+        public {
             host = 0.0.0.0
+            port = 9090
         }
     }
 }
--- a/jvb/rootfs/defaults/jvb.conf
+++ b/jvb/rootfs/defaults/jvb.conf
@@ -42,6 +42,9 @@ videobridge {
         server-id = "{{ $WS_SERVER_ID }}"
     }
     http-servers {
+        private { 
+          host = 0.0.0.0
+        }
         public {
             host = 0.0.0.0
             port = 9090


prayagsingh:
private config is used for health-checks and debugs. 
I think you can't check `colibri/stats` now. Can you please confirm it?

saghul:
@bbaldino can you PTAL?

pierreozoux:
@prayagsingh if you tell me how to check, I can do it :)

prayagsingh:
> 
> 
> @prayagsingh if you tell me how to check, I can do it :)

Please use `curl http://127.0.0.1:8080/colibri/stats` command. If it didn't work then try below config.
```
http-servers {
    # The HTTP server which hosts services intended for 'public' use
    # (e.g. websockets for the bridge channel connection)
    public {
      # See JettyBundleActivatorConfig in Jicoco for values
      port = 9090
      host = 0.0.0.0
    }
    # The HTTP server which hosts services intended for 'private' use
    # (e.g. health or debug stats)
    private {
      # See JettyBundleActivatorConfig in Jicoco for values
      host = 0.0.0.0
      port = 8080 
    }
  }
```
reference: https://github.com/jitsi/jitsi-videobridge/blob/master/jvb/src/main/resources/reference.conf

pierreozoux:
Looks like it is still active :)

```
docker-compose exec jvb bash
root@xxx:/# curl http://127.0.0.1:8080/colibri/stats
{"inactive_endpoints":0,"inactive_conferences...
```

Thanks for the link to the actual conf!

prayagsingh:
Actually In my case it's working with `private` config too. But there is a corner case when it failed. If using jibri to record the meeting and the other participant joined the meeting with conference setup(like Clearone Interact 8i) then I was getting websocket error in console log. I'll try your config.

sapkra:
@pierreozoux  Can you try accessing the API from outside the container?

bbaldino:
Right now I believe the private server is enabled by default and the public one isn't, but the private server binds to 127.0.0.1 by default, so not sure if that will work correctly in a docker env? In addition to the changes here, you may need to add:

```hocon
private {
    host = 0.0.0.0
}
```

sapkra:
We discussed this topic in the following issue and implemented this 0.0.0.0 workaround.

https://github.com/jitsi/docker-jitsi-meet/issues/758

unteem:
Hey there I have been using this fix with @pierreozoux I can confirm that the current PR will break access to the stats endpoint from an external container as mentioned in #758. I guess without it, it doesn't break jvb as health checks should still work but if you want to use a prom exporter for instance that won't work.

So for stats and websockets to work we need:
```
    http-servers {
        private { 
          host = 0.0.0.0
       }
        public {
            host = 0.0.0.0
            port = 9090
        }
    }
}
```

bbaldino (approved):

PR #830, opened 2020-11-20
prosody: add env var to configure cross-domain policy

sapkra:
Possible values:
- comma separated list of domains, but PUBLIC_URL will automatically be included
- `true` to enable cross domain policy for all domains

### DIFF
--- a/prosody/Dockerfile
+++ b/prosody/Dockerfile
@@ -18,6 +18,8 @@ RUN \
 
 FROM ${JITSI_REPO}/base
 
+ENV XMPP_CROSS_DOMAIN="false"
+
 RUN \
     wget -q https://prosody.im/files/prosody-debian-packages.key -O - | gpg --enarmor > /etc/apt/trusted.gpg.d/prosody.asc \
     && echo "deb http://packages.prosody.im/debian buster main" > /etc/apt/sources.list.d/prosody.list \
--- a/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
+++ b/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
@@ -26,11 +26,20 @@ asap_accepted_issuers = { "{{ join "\",\"" (splitList "," .Env.JWT_ACCEPTED_ISSU
 asap_accepted_audiences = { "{{ join "\",\"" (splitList "," .Env.JWT_ACCEPTED_AUDIENCES) }}" }
 {{ end }}
 
-{{ if $ENABLE_XMPP_WEBSOCKET }}
+consider_bosh_secure = true;
+
 -- Deprecated in 0.12
 -- https://github.com/bjc/prosody/commit/26542811eafd9c708a130272d7b7de77b92712de
-cross_domain_websocket = { "{{ $PUBLIC_URL }}" };
-consider_bosh_secure = true;
+{{ $XMPP_CROSS_DOMAINS := $PUBLIC_URL }}
+{{ if eq .Env.XMPP_CROSS_DOMAIN "true"}}
+cross_domain_websocket = true
+cross_domain_bosh = true
+{{ else }}
+{{ if and .Env.XMPP_CROSS_DOMAIN (not (eq .Env.XMPP_CROSS_DOMAIN "false" )) }}
+  {{ $XMPP_CROSS_DOMAINS = list $PUBLIC_URL .Env.XMPP_CROSS_DOMAIN | join "," }}
+{{ end }}
+cross_domain_websocket = { "{{ join "\",\"" (splitList "," $XMPP_CROSS_DOMAINS) }}" }
+cross_domain_bosh = { "{{ join "\",\"" (splitList "," $XMPP_CROSS_DOMAINS) }}" }
 {{ end }}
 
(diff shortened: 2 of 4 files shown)


prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -27,10 +27,25 @@ asap_accepted_audiences = { "{{ join "\",\"" (splitList "," .Env.JWT_ACCEPTED_AU
 {{ end }}
 
 {{ if $ENABLE_XMPP_WEBSOCKET }}
+consider_bosh_secure = true;
```
sapkra, line 30: Do we need this for websockets only? I think it would be great to remove the `if`.

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -27,10 +27,25 @@ asap_accepted_audiences = { "{{ join "\",\"" (splitList "," .Env.JWT_ACCEPTED_AU
...
+{{ $XMPP_CROSS_DOMAIN := $PUBLIC_URL }}
+{{ if eq (.Env.XMPP_CROSS_DOMAIN | default "false") "true"}}
+cross_domain_bosh = true
 -- Deprecated in 0.12
 -- https://github.com/bjc/prosody/commit/26542811eafd9c708a130272d7b7de77b92712de
-cross_domain_websocket = { "{{ $PUBLIC_URL }}" };
-consider_bosh_secure = true;
+{{ if $ENABLE_XMPP_WEBSOCKET }}
+cross_domain_websocket = true
+{{ end }}
```
sapkra, line 40: Same here! Do we need the if? Can we not just apply this setting even if websockets are disabled?

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
saghul: Yeah, let's drop it.

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
saghul: yep, we sure can

saghul (commented):
Left a couple of comments.

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -26,11 +26,20 @@ asap_accepted_issuers = { "{{ join "\",\"" (splitList "," .Env.JWT_ACCEPTED_ISSU
...
 {{ end }}
 
-{{ if $ENABLE_XMPP_WEBSOCKET }}
+consider_bosh_secure = true;
+
 -- Deprecated in 0.12
 -- https://github.com/bjc/prosody/commit/26542811eafd9c708a130272d7b7de77b92712de
-cross_domain_websocket = { "{{ $PUBLIC_URL }}" };
-consider_bosh_secure = true;
+{{ $XMPP_CROSS_DOMAIN := $PUBLIC_URL }}
```
saghul, line 33: Can we call this XMPP_CROSS_DOMAINS? Reading the variable being called the same as the env one but with a different meaning threw me off a bit.

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -26,11 +26,20 @@ asap_accepted_issuers = { "{{ join "\",\"" (splitList "," .Env.JWT_ACCEPTED_ISSU
...
+{{ if eq (.Env.XMPP_CROSS_DOMAIN | default "false") "true"}}
+cross_domain_websocket = true
+cross_domain_bosh = true
+{{ else }}
+{{ if .Env.XMPP_CROSS_DOMAIN }}
```
saghul, line 38: no default | to Bool ?

sapkra: We don't need a default or toBool because if a string is empty it's `false` by default.

sapkra: Ohh ok, you mean if someone wants to set `false` it breaks, right?

sapkra: Changed it.

prosody/Dockerfile
```diff
@@ -18,6 +18,8 @@ RUN \
 
 FROM ${JITSI_REPO}/base
 
+ENV XMPP_CROSS_DOMAIN="false"
```
saghul, line 21: We tend not to do this, why do it for this specific value?

saghul (commented):
One tiny nit, LGTM otherwise.

prosody/Dockerfile
sapkra: I think we should start defining the defaults in the dockerfile instead of defining defaults all over again in several files. There should be a single point of truth which should always be the env var IMO. I think this PR can be a start for this transition.

saghul: Fair enough!

saghul (approved):

PR #835, opened 2020-11-23
✨ Env var to setup STUN/TURN

madmath03:
:sparkles: Env var to setup STUN/TURN to close #820

Signed-off-by: mathieu.brunot <mathieu.brunot@monogramm.io>

PR #667, opened 2020-07-03
Coturn

goacid:
Based on PR163 from netaskd : https://github.com/jitsi/docker-jitsi-meet/pull/163/

-     Fix prosody configuration
-     Fix web configuration
-     Fix name : turn => coturn
-     Update README.md

PR #837, opened 2020-11-23
✨ Add mod_turncredentials to prosody

madmath03:
:sparkles: Add mod_turncredentials to prosody

There have been so many TURN Pull Requests (latest active being #163) and none being validated yet that I'm thinking it might be good to do it step by step.

- #835 Provides env var to configure Jitsi Web STUN support
- this one just adds mod_turncredentials.lua to prosody. Without any other developments, TURN can be setup simply with GLOBAL_* env var:

```
GLOBAL_MODULES=turncredentials
GLOBAL_CONFIG=turncredentials_secret = "mysupersecuresharedturnsecret";\nturncredentials_host = "turn.example.com";\nturncredentials_port = 5349;
```

@saghul what do you think ?


goacid:
This is already in the PR for adding Turn
https://github.com/jitsi/docker-jitsi-meet/pull/667

saghul:
#667 is indeed the way to go here.

PR #842, opened 2020-11-27
new jicofo config. 

prayagsingh:
Moving to jicofo.conf file. removed sip-communicator.properties entirely.


prayagsingh:
Entire discussion happened here: https://github.com/jitsi/docker-jitsi-meet/issues/825
docker images available for testing: `prayagsingh/jicofo:wsv9`.

env.example
```diff
@@ -306,10 +306,13 @@ JIBRI_FINALIZE_RECORDING_SCRIPT_PATH=/config/finalize.sh
 # XMPP user for Jibri client connections
 JIBRI_XMPP_USER=jibri
 
-# MUC name for the Jibri pool
+# MUC name for the Jibri pool. The JID of the MUC to be used as a brewery for jibri instances for streaming.
```
bgrozev, line 309: nit: It's just the name of the room ("JibriBrewery") as opposed to the JID ("JibriBrewery@something").

env.example
```diff
@@ -338,3 +341,40 @@ RESTART_POLICY=unless-stopped
...
+#ENABLE_AUTO_OWNER=true
+
+# How long to wait for the initial participant in a conference.
+#JICOFO_CONF_INITIAL_PARTICIPANT_WAIT_TIMEOUT="15 seconds"
+
+# How long a participant's media session will be kept alive once it remains the only participant in the room.
+#JICOFO_CONF_SINGLE_PARTICIPANT_TIMEOUT="20 seconds"
+            
+# jibri-sip  
+#ENABLE_JIBRI_SIP=0
```
bgrozev, line 377: Is this intentionally not implemented?

env.example
```diff
@@ -338,3 +341,40 @@ RESTART_POLICY=unless-stopped
 
 # Authenticate using external service or just focus external auth window if there is one already.
 # TOKEN_AUTH_URL=https://auth.meet.example.com/{room}
+
+# Jicofo config
+JICOFO_SHORT_ID=1
```
bgrozev, line 346: nit: I would name it JICOFO_OCTO_ID

jicofo/rootfs/etc/services.d/jicofo/run
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/with-contenv bash
 
-JAVA_SYS_PROPS="-Dnet.java.sip.communicator.SC_HOME_DIR_LOCATION=/ -Dnet.java.sip.communicator.SC_HOME_DIR_NAME=config -Djava.util.logging.config.file=/config/logging.properties"
```
bgrozev, line 3: Have you verified that logging works as expected after removing SC_HOME_DIR_LOCATION and SC_HOME_DIR_NAME? I see *some* code that uses them to decide where to put the log file.

bgrozev:
Excellent! I left a few minor comments. I hesitate to approve only because I don't know much about the docker setup. @saghul ?

env.example
prayagsingh: sure I will update it

env.example
prayagsingh: I thought of it but I don't want to break the earlier functionality like everyone have to change this variable too. Just upgrading to new image won't be enough. That's why I didn't change it.

env.example
prayagsingh: Yes. I have no idea about Jibri-sip(never used it). I think @saghul can help here.

env.example
```diff
@@ -351,3 +354,40 @@ RESTART_POLICY=unless-stopped
...
+# The bridge selection strategy. The built-in strategies are:
+# --> SingleBridgeSelectionStrategy: Use the least loaded bridge, do not split a conference between bridges (Octo).
+# --> SplitBridgeSelectionStrategy: Use a separate bridge for each participant (for testing).
+# --> RegionBasedBridgeSelectionStrategy: Attempt to put each participant in a bridge in their local region (i.e. use Octo for geo-location).
+# --> IntraRegionBridgeSelectionStrategy: Use additional bridges when a bridge becomes overloaded (i.e. use Octo for load balancing). 
+OCTO_BRIDGE_SELECTION_STRATEGY=SingleBridgeSelectionStrategy
+
+# Configure the codecs and RTP extensions to be used in the offer sent to clients.
+ENABLE_CODEC_VP8=true
+ENABLE_CODEC_VP9=false
```
prayagsingh, line 370: @bgrozev I have disabled VP9 for now. I'm seeing some issues when a meeting is started and one of the participant is using Mozilla firefox and other is using chrome. video started freezing on firefox and finall stopped. Chrome to Edge, Brave or vice-versa conferences are working fine as they all are built on top of chromium.

prayagsingh:
@bgrozev you have reverted back the jicofo reservation system. Any planning to add it in reservation.conf/jicofo.conf?
https://github.com/jitsi/jicofo/commit/be025fbde8e077784f21fe0fcf5b041716961b56

bgrozev:
> @bgrozev you have reverted back the jicofo reservation system. Any planning to add it in reservation.conf/jicofo.conf?

Yes, I have that on a local branch, I'll make a PR next week

env.example
prayagsingh: **Mozilla VP9 issue**
https://bugzilla.mozilla.org/show_bug.cgi?id=1633876

jicofo/rootfs/defaults/jicofo.conf
```diff
@@ -0,0 +1,126 @@
...
+      {{ if .Env.JIBRI_PENDING_TIMEOUT }}
+      pending-timeout = "{{ .Env.JIBRI_PENDING_TIMEOUT }}"
+      {{ end }}
+    }
+    {{ end }}
+  
+    {{ if $ENABLE_JIBRI_SIP }}
+    jibri-sip {
+      // The JID of the MUC to be used as a brewery for jibri instances for SIP.
+      # brewery-jid = "jibrisipbrewery@example.com"
```
prayagsingh, line 95: @saghul @sapkra  Can you please help here? What is the value that I need to put in here? Is there any env variable that I can refer to for `jibrisipbrewery` ?
`brewery-jid = "jibrisipbrewery@{{ .Env.XMPP_INTERNAL_MUC_DOMAIN }}"`

saghul: I'd drop the Jibri SIP config part. It's not yet ready for prime time.

env.example
```diff
@@ -271,7 +271,7 @@ JICOFO_AUTH_USER=focus
 #JICOFO_RESERVATION_REST_BASE_URL=http://reservation.example.com
 
 # Enable Jicofo's health check REST API (http://<jicofo_base_url>:8888/about/health)
-#JICOFO_ENABLE_HEALTH_CHECKS=true
+JICOFO_ENABLE_HEALTH_CHECKS=false
```
saghul, line 274: Please revert all changes to the env.example file, I want to move all of those to the handbook instead.

saghul (changes requested):

env.example
prayagsingh: Removed new env variables from env.example file.

prayagsingh:
Hello @bgrozev Any update on jicofo reservation system. Before holidays you mentioned that it is on your local. Is it merged ?
It is the only thing pending in this new jicofo config.

bgrozev:
Hey @prayagsingh the reservation system can now be configured via `jicofo.reservation.base-url` and `jicofo.reservation.enabled`. It is unclear what the longer term plans for it are (whether we'll continue to support it or now), so unless you have a use-case for it I would leave it out of the config file for now.

prayagsingh:
@saghul Please suggest whether to add it in jicofo.conf or not else I have to revert back sip-communicator file to support this functionality only. 

In case we add it in jicofo.conf.
```
jicofo {
  {{ if .Env.ENABLE_JICOFO_RESERVATION_SYS }}
  reservation {
    enable={{ .Env.ENABLE_JICOFO_RESERVATION_SYS }}
    base-url={{ .Env.JICOFO_RESERVATION_REST_BASE_URL }}
  }
 {{ end }}
}
```

saghul:
Let's add it in jicofo.conf, your proposal looks ok, but please let's call it JICOFO_RESERVATION_ENABLED

saghul:
Landed in 0f541c8, cheers!

PR #843, opened 2020-11-27
web: always rebuild nginx configs on start

debfx:
Recent changes to enable websockets by default require accompanying changes in the nginx config.

Unbreak upgrades by always rebuilding the nginx configs on start.

### DIFF
--- a/web/rootfs/etc/cont-init.d/10-config
+++ b/web/rootfs/etc/cont-init.d/10-config
@@ -54,17 +54,14 @@ fi
 # copy config files
 tpl /defaults/nginx.conf > /config/nginx/nginx.conf
 
-if [[ ! -f /config/nginx/meet.conf ]]; then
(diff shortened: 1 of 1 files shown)


sapkra:
@saghul In my opinion it doesn't really make sense to not generate the configs on each start. I mean each config, not just those in the PR. What was the reason you have designed it like this? The startup time? I don't think it will be improved by skipping the templating that much (it will be in the milliseconds I believe).

saghul:
The reason it started that way was my own ignorance :-)

We have gradually moved things to be created at startup now, LoL the new web and JVB configs.

The plan is to have everything created at runtime. Any help towards this goal is more than welcome.

saghul (approved):
Thank you!

sapkra:
@saghul Ahh ok, so this will simplify a lot.

PR #852, opened 2020-12-04
Add container_name attributes to docker-compose.yml

matthinc:
The container_name attribute makes it easier to reference a specific container when working with docker-commands.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,6 +4,7 @@ services:
     # Frontend
     web:
         image: jitsi/web:latest
+        container_name: 'jitsi_web'
         restart: ${RESTART_POLICY}
         ports:
             - '${HTTP_PORT}:80'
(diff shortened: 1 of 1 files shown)


saghul:
Why do you need this? When using `docker-compose` you can already refer to them by their short name: `docker-compose logs web` for example.

matthinc:
In my specific case: The monitoring system requires a container_name to check if all docker containers are running properly.
But I like the idea of container_name in general. It's very handy if you use `docker ps` or `docker stats` to get an overview of all docker containers on the current machine.

debfx:
> It's very handy if you use docker ps or docker stats to get an overview of all docker containers on the current machine.

docker-compose automatically assigns readable container names. If you name the folder where docker-compose.yml is located `jitsi` it's `jitsi_web_1` for example.

> I just think there is no reason not to have it.

The downside is that you can't run multiple Jitsi instances on the same dockerd (think of test and prod instances).

saghul:
> The downside is that you can't run multiple Jitsi instances on the same dockerd (think of test and prod instances).

Big enough of a downside indeed.

PR #853, opened 2020-12-05
More adaptable setting for worker_processes in nginx

MarkErik:
re: worker_processes: Rather than a fixed number, use auto so that based on the machine that will be running it, Nginx can select the appropriate amount of processes.

From the nginx docs:
"The optimal value depends on many factors including (but not limited to) the number of CPU cores, the number of hard disk drives that store data, and load pattern. When one is in doubt, setting it to the number of available CPU cores would be a good start (the value “auto” will try to autodetect it)."

### DIFF
--- a/web/rootfs/defaults/nginx.conf
+++ b/web/rootfs/defaults/nginx.conf
@@ -1,5 +1,5 @@
 user www-data;
-worker_processes 4;
+worker_processes auto;
 pid /run/nginx.pid;
 include /etc/nginx/modules-enabled/*.conf;
 


debfx:
This seems problematic to me in a container context. Jitsi might run on a big machine among many containers.

nginx would at least need to detect the container CPU limit. `auto` however always uses the host CPU count.

debfx:
See https://github.com/nginxinc/docker-nginx/issues/472 for a similar discussion on the nginx image.

MarkErik:
@debfx Thank you for highlighting the implications that 'auto' would have in a container context - and linking to the ongoing discussion at the nginx docker project.

Good point about auto perhaps assigning more worker_processes than desired when Jitsi is run on a big machine.

In a similar vein, having 4 as the default can also be less than optimal for more modest deployments e.g. 1 or 2 CPUs, due to context switching.

From the nginx documentation: [https://www.nginx.com/blog/inside-nginx-how-we-designed-for-performance-scale/](url)
- "Processes and threads consume resources. They each use memory and other OS resources, and they need to be swapped on and off the cores (an operation called a context switch)."
- "The NGINX configuration recommended in most cases – running one worker process per CPU core – makes the most efficient use of hardware resources."

The question then is: what is the most sensible default value for Jitsi's docker nginx config?

I might argue that those who are running it on large machines / multiple JVBs etc. are likely to customise the config defaults themselves to tune the performance.

So perhaps a better default for Jitsi might be:
worker_processes: 2
worker_connections: 1024 (a reasonable default)
-> providing 2048 connections, compared to the current 3072, and with less of a need for context switching.

I wonder how many connections Jitsi needs open at one time?

Would even 1 worker_process and 1024 worker_connections be sufficient for deployments?

Given that there seems to be a limit to how many people can feasibly be in one room at a time (~10-15 with video, 30+ audio only) on basic servers, and how many concurrent conferences can be run perhaps 1*1024 is a very good default.
And again, if people are planning to deploy a setup that can handle significant load, they are likely customising these defaults.

saghul:
Given this is the first time this has been brought up, I'd say the current defaults are good enough. Generally speaking we try to make upgrades as painless as possible, so I wouldn't want to force people to manually tune what they previously didn't have to tune.

MarkErik:
@saghul Agreed that using defaults which need to be tuned is not a good idea - that was my original motivation for suggesting we use 'auto'.

I am curious how did the current defaults get picked - as 4 worker_processes, and 768 worker_connections isn't the usual "out of the box" nginx defaults.

Note: I want to say I do have a bias - in setting up private Jitsi instances for friends and family it has been on small VPSs with less than 4 cores. (I am trying to track down what is causing stutters in the video conferences - on different servers (always with very little CPU usage), different regions, P2P, through JVB, Fibre connections, DSL, Chrome, Windows, Mac, 480p to 1080p cams etc.) that's why I came across the nginx.conf and thought that perhaps its not optimal for smaller servers.

saghul:
> I am curious how did the current defaults get picked - as 4 worker_processes, and 768 worker_connections isn't the usual "out of the box" nginx defaults.

As the one who picked them... I don't remember :-P



> Note: I want to say I do have a bias - in setting up private Jitsi instances for friends and family it has been on small VPSs with less than 4 cores. (I am trying to track down what is causing stutters in the video conferences - on different servers (always with very little CPU usage), different regions, P2P, through JVB, Fibre connections, DSL, Chrome, Windows, Mac, 480p to 1080p cams etc.) that's why I came across the nginx.conf and thought that perhaps its not optimal for smaller servers.

Appreciate the feedback! Note that I'm not opposing we change this! Happy to make it configurable via env variables, defaulting to the current values. Since picking a good value automagically seems complex, we can just expose it as such, so it's easy to test: changing the value and restarting the container would do.

MarkErik:
@saghul "Appreciate the feedback! Note that I'm not opposing we change this! Happy to make it configurable via env variables, defaulting to the current values. Since picking a good value automagically seems complex, we can just expose it as such, so it's easy to test: changing the value and restarting the container would do."

That would be fantastic!

saghul:
Sweet. Are you up for submitting a PR?

MarkErik:
Yes, happy to investigate how to do it and implement it.

MarkErik:
PR request at: https://github.com/jitsi/docker-jitsi-meet/pull/864

PR #860, opened 2020-12-07
Add reloadcmd to acme.sh install-cert script

konsti:
The `README` on https://github.com/acmesh-official/acme.sh states that

> The reloadcmd is very important. The cert can be automatically renewed, but, without a correct 'reloadcmd' the cert may not be flushed to your server [...]

This PR adds a `reloadcmd` to the `--install-cert` routine and sends a SIGHUP to restart nginx gracefully if the service exists.

@saghul


saghul:
I don't think that is necessary because of the pre and post install hooks. nginx will be stopped and then started.

konsti:
You are right. I didn't find this page in their wiki: https://github.com/acmesh-official/acme.sh/wiki/Using-'--pre-hook',--'--post-hook',-'---renew-hook'-and-'---reloadcmd'

PR #861, opened 2020-12-07
Add support for multi-domain certificates

konsti:
We found the need to support multiple domains with LetsEncrypt as we want to serve the video bridge under a separate domain.

This PR enhances the Nginx config in the `web` project to allow multiple domains with the same certificate.

### DIFF
--- a/env.example
+++ b/env.example
@@ -65,7 +65,7 @@ TZ=UTC
 # Enable Let's Encrypt certificate generation
 #ENABLE_LETSENCRYPT=1
 
-# Domain for which to generate the certificate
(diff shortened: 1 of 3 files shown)


saghul:
Thanks! In general I don't want to add options which are not really used by this setup. If we need multiple domains I'd rather have extra env variables for each one we use. Maybe you can give us some more details on how exactly you're using the certs?

konsti:
Totally understood. We have configured nginx in `web` to route the colibri traffic to the videobridge with an alternative `server_name`. I thought this config here might be useful for others.

Feel free to close the PR.

saghul:
Sorry, I'm considering this out of scope for the time being.

PR #866, opened 2020-12-09
Add config for presence_identity module of prosody.

zamarawka:
Add optional env param for parse user identifiers from jwt token.
relates https://github.com/jitsi/lib-jitsi-meet/blob/master/doc/tokens.md#access-token-identifiers--context


sapkra:
You can also just use the XMPP_MODULES variable:

```
XMPP_MODULES=presence_identity
```

Issue #805, opened 2020-10-31
Jisti-meet stable-5142 and stable-5076 doesn't work on system without ipv6.


michaeloosh:
Jisti-meet doesn't work on system without ipv6 since ipv6 support has been added in docker-jitsi-meet. Tested on two systems the latest Arch Linux and Centos 7. Doesn't work when ipv6 is disabled in OS. Tested versions: Jisti-meet stable-5142 and stable-5076. Both versons work after enabling ipv6.

saghul:
Hum, does reverting this work? https://github.com/jitsi/docker-jitsi-meet/commit/5ceaf5fd025b005c6ddfbc87b18c77fe402a8893#diff-4b5e57f6eb2f42b9039b3d1e13929295f231749c510cbe341cd68036d9af97e2

Not sure why that would be problematic since it should listen also on IPv4.

saghul:
Can you try to edit the file and set it as such: `listen [::]:80 ipv6only=off;`

michaeloosh:
Adding `ipv6only=off` didn't help. I still got an errors in nginx error logs:
`[emerg] 369#369: socket() [::]:80 failed (97: Address family not supported by protocol)`
or
`[emerg] 596#596: socket() [::]:443 failed (97: Address family not supported by protocol)` when disabling `listen [::]:80`

Only complete  removal of ipv6 from nginx configuration helped to start nginx and run jitsi on host with ipv4 only, so yes. reverting to https://github.com/jitsi/docker-jitsi-meet/commit/5ceaf5fd025b005c6ddfbc87b18c77fe402a8893#diff-4b5e57f6eb2f42b9039b3d1e13929295f231749c510cbe341cd68036d9af97e2 helped.

https://admin-ahead.com/forum/general-linux/address-family-not-supported-by-protocol-nginx-error/

Another option is to enable ipv6 on the host OS.

saghul:
Hum. If you can find a way to make it work on both cases I'd take a PR (an env var defaulting to true would suffice), but it's 2020, I don't think reverting is a good idea.

ebricca:
this might be related .. when starting via docker-compose up on the latest version (5142)
.. nginx inside the web container does not start .. error shown .. 
nginx: [emerg] bind() to 0.0.0.0:80 failed (98: Address already in use) 
.. host is ubuntu 20.04 - with ipv4 adresses configured via netplan

PR #868, opened 2020-12-09
Feat/can disable ipv6

karniemi:
In some environments it's required to disable ipv6 due to security constraints. It was reported in #805, that docker-jitsi-meet no longer started properly in such environments. The problem was shortly discussed in the related commit: https://github.com/jitsi/docker-jitsi-meet/commit/5ceaf5fd025b005c6ddfbc87b18c77fe402a8893

This PR introduces a new environment parameter, which makes it possible to disable the problem causing ipv6 nginx-listeners. By default, ipv6 is enabled.

### DIFF
--- a/env.example
+++ b/env.example
@@ -353,7 +353,7 @@ JIBRI_LOGS_DIR=/config/logs
 #ENABLE_HTTP_REDIRECT=1
 
 # Enable IPv6
-# Provides means to disable ipv6 in environments that don't support it.
+# Provides means to disable IPv6 in environments that don't support it (get with the times, people!)
 #ENABLE_IPV6=1
 
 # Container restart policy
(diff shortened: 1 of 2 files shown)


saghul:
Can you add the variable to docker-compose.yml as well?

karniemi:
@saghul ...luckily....it seems I picked the same name that is already used for this purpose :-): https://github.com/jitsi/docker-jitsi-meet/search?q=ENABLE_IPV6&type=

I'm embarrassed to admit that I did not even check if there already was some variable for ipv6. Shame on me.

saghul:
Oh, that's nice :-)

env.example
```diff
@@ -352,6 +352,10 @@ JIBRI_LOGS_DIR=/config/logs
 # Necessary for Let's Encrypt, relies on standard HTTPS port (443)
 #ENABLE_HTTP_REDIRECT=1
 
+# Enable ipv6
```
saghul, line 355:
```suggestion
# Enable IPv6
```

saghul (approved):

Issue #616, opened 2020-05-27
jigasi default room not defined or customizable


r-eason:
From https://github.com/jitsi/jigasi/blob/master/jigasi-home/sip-communicator.properties org.jitsi.jigasi.DEFAULT_JVB_ROOM_NAME=siptest  does not exist in docker deployment unless manually added.

soletan:
I second this feature request as it is the only piece missing in current docker-based setup right now.

PR #873, opened 2020-12-14
control SIP default room for incoming calls

soletan:
This merge request is adding support for controlling default room to use on incoming SIP calls lacking of proper Invite header to prevent this setting from requiring manual patching of internal configuration file. It is regarding #616.

### DIFF
--- a/env.example
+++ b/env.example
@@ -129,9 +129,6 @@ ETHERPAD_SKIN_VARIANTS="super-light-toolbar super-light-editor light-background 
 # SIP server transport
 #JIGASI_SIP_TRANSPORT=UDP
 
-# SIP default room name for incoming calls lacking of Invite header
-#JIGASI_SIP_DEFAULT_ROOM=conference
-
 #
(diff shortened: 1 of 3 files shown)


saghul:
Sorry for the delay, looks like there is a conflict now, any chance you could take a look?

soletan:
That's obviously due to my other pull request ...

env.example
```diff
@@ -129,6 +129,9 @@ ETHERPAD_SKIN_VARIANTS="super-light-toolbar super-light-editor light-background
 # SIP server transport
 #JIGASI_SIP_TRANSPORT=UDP
 
+# SIP default room name for incoming calls lacking of Invite header
+#JIGASI_SIP_DEFAULT_ROOM=conference
```
saghul, line 133: Please drop this part, I want to move all the examples to the handbook instead. Otherwise LGTM.

soletan: I see. It's done.

saghul (approved):

Issue #500, opened 2020-04-21
jigasi + ldap auth not working out of the box


creativeguitar:
Jigasi works when auth is turned off.
If I turn on ldap auth, then I get the following error

```
jigasi_1   | Jigasi 2020-04-21 13:50:36.475 INFO: [75] org.jitsi.jigasi.SipGateway.incomingCallReceived().188 Incoming call received...
jigasi_1   | Jigasi 2020-04-21 13:50:37.480 INFO: [76] org.jitsi.jigasi.SipGatewaySession.run().1295 [ctx=1587495036476430832605]Using default JVB room name property siptest
jigasi_1   | Jigasi 2020-04-21 13:50:37.486 INFO: [76] org.jitsi.jigasi.JvbConference.setXmppProvider().539 [ctx=1587495036476430832605] Using ProtocolProviderServiceJabberImpl(Jabber:3979586d@meet.jitsi/3979586d)
jigasi_1   | Jigasi 2020-04-21 13:50:37.558 INFO: [78] impl.protocol.jabber.OperationSetBasicTelephonyJabberImpl.registrationStateChanged().125 Jingle : ON
jigasi_1   | Jigasi 2020-04-21 13:50:37.558 INFO: [78] org.jitsi.jigasi.JvbConference.registrationStateChanged().577 [ctx=1587495036476430832605] Registering XMPP.
jigasi_1   | Jigasi 2020-04-21 13:50:37.577 INFO: [78] impl.protocol.jabber.ProtocolProviderServiceJabberImpl.authenticated().2535 Authenticated: false
jigasi_1   | Jigasi 2020-04-21 13:50:37.579 INFO: [78] org.jitsi.jigasi.JvbConference.joinConferenceRoom().648 [ctx=1587495036476430832605] Joining JVB conference room: siptest
jigasi_1   | Jigasi 2020-04-21 13:50:37.584 SEVERE: [78] org.jitsi.jigasi.JvbConference.joinConferenceRoom().760 [ctx=1587495036476430832605]java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
jigasi_1   | java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
jigasi_1   |    at java.util.ArrayList.rangeCheck(ArrayList.java:657)
jigasi_1   |    at java.util.ArrayList.get(ArrayList.java:433)
[...]

PR #874, opened 2020-12-14
fix LDAP jigasi auth

soletan:
This merge request is conditionally adding configuration suggested in #500 to jigasi SIP configuration file when LDAP authentication has been selected in .env file.

### DIFF
--- a/jigasi/rootfs/defaults/sip-communicator.properties
+++ b/jigasi/rootfs/defaults/sip-communicator.properties
@@ -143,4 +143,4 @@ org.jitsi.jigasi.transcription.RECORD_AUDIO={{ .Env.JIGASI_TRANSCRIBER_RECORD_AU
 org.jitsi.jigasi.transcription.RECORD_AUDIO_FORMAT=wav
 {{end}}
 
-org.jitsi.jigasi.MUC_SERVICE_ADDRESS={{ .Env.XMPP_INTERNAL_MUC_DOMAIN }}
+org.jitsi.jigasi.MUC_SERVICE_ADDRESS={{ .Env.XMPP_MUC_DOMAIN }}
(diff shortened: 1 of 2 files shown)


saghul:
I don't see how this is specific to LDAP based authentication.

@damencho In general, does it make sense to have this option set at all times?

damencho:
Yep, this can be set all the time. It also optimizes some room discovery happening in some cases before joining if I recall correctly...

soletan:
Well, even better ... I just don't know the underlying system and so I tried to satisfy #500 at most.

jigasi/rootfs/defaults/sip-communicator.properties
```diff
@@ -142,3 +142,5 @@ org.jitsi.jigasi.transcription.SEND_TXT={{ .Env.JIGASI_TRANSCRIBER_SEND_TXT | de
 org.jitsi.jigasi.transcription.RECORD_AUDIO={{ .Env.JIGASI_TRANSCRIBER_RECORD_AUDIO | default "false"}}
 org.jitsi.jigasi.transcription.RECORD_AUDIO_FORMAT=wav
 {{end}}
+
+org.jitsi.jigasi.MUC_SERVICE_ADDRESS=muc.meet.jitsi
```
saghul, line 146: Please use the corresponding env variable here instead of hardcoding the default.

soletan: I don't know MUC, XMPP or anything related to the technical background of Jitsi service. All I know is the way docker files are created and scripts are used to establish configuration files here. Due to that I don't know if the variable picked in latest commit is the proper one for there are two variables in env.example using this value both looking like eligible candidates here.

saghul: No worries. This should be XMPP_MUC_DOMAIN, which defaults to `muc.meet.jitsi`

soletan: So it is.

saghul:
Cheers!

PR #899, opened 2021-01-04
add CONFCODE_URL and DIALIN_NUMBERS_URL

loelkes:
I've added placeholders for the environment variables `CONFCODE_URL` and `DIALIN_NUMBERS_URL` to the `env.example` file. I believe it should be available in the examples.


saghul:
I'd like to move away from there and put them all in the handbook. I'm hoping to get that started this week. Thanks for the effort though!

loelkes:
OK, good. I already prepared that PR for you as well, see jitsi/handbook#129

saghul:
Excellent!

PR #900, opened 2021-01-06
configure multiple jwt specific modules with comma separated values

prayagsingh:


saghul:
You should be able to accomplish this with XMPP_MUC_MODULES, right?

prayagsingh:
I missed it completely. yes we can use this module.

PR #906, opened 2021-01-11
Be able to add custom parameters at sip-communicator.properties

kpeiruza:
With this tiny patch you can define a multiline variable containing any missing config entries.

In our case, we need it to ease OCTO's configuration, tweaking and testing.


kpeiruza:
Not needed if /config/custom-sip-communicator.properties added.

Issue #908, opened 2021-01-15
Publish unstable tag to docker hub as well


pierreozoux:
Hi,

would it be possible to also publish unstable tags?

And I think that you use Jenkins. I'd love to open a Pulll Request to create a nightly unstable build, but it doesn't look like possible at the moment :) Or? if you give me indication, I'd love to contribute that feature, I think we are quite some people to need the unstable, so this would share the effort for everybody, instead of all rebuilding on our side.

I could also try a github action if you are open to.

saghul:
I have toyed with the idea, thanks for voicing it out! I think we could use GH actions and run it daily at some point in time as a start.

pierreozoux:
Thanks, I'll put it in my todo :)

PR #918, opened 2021-01-25
Publish nightly the unstable tag 

hrenard:
Hi,
This solve #908. 
[Workflow demo](https://github.com/hrenard/docker-jitsi-meet/actions/runs/530524485)
Need to add 3 secrets to the repository : 
- DOCKERHUB_USERNAME (jitsi)
- DOCKERHUB_TOKEN
- JITSI_REPO (jitsi)


saghul:
Looks great, thank you! I need to take a second look, since I think your change will break release.sh at the moment.

saghul:
Also, I wonder if we should tag this with unstable-YYYY-MM-DD ?

hrenard:
Thanks. 

I tried to minimize changes on the existing stuff. If I understood correctly, images are always built on top of the latest base (or base-java) image. The BASE_TAG that I introduce has latest value by default, so it shouldn’t break anything. But tell me if I’m mistaken.

It’s a good question, I’m not sure there is real a benefice of keeping such history. I think people should use the unstable build to try out the latest features but if they want to stick to a specific version, it should be a release.

saghul:
> It’s a good question, I’m not sure there is real a benefice of keeping such history. I think people should use the unstable build to try out the latest features but if they want to stick to a specific version, it should be a release.

I think there is definitely some value in being able to tell "something works in unstable-XXX but not in unstable-YYY".

hrenard:
Okay, I'll add this change.

hrenard:
I added the dated tag. Tell me if I need to change something else.

pierreozoux:
@saghul do you need help to get this moving? it would be an amazing addition :)

Thanks for your time!

saghul:
It's close to the top of my list, I just haven't gotten around it yet, sorry!

saghul:
Sorry it took some time. Landed now, let's see if I set it up correctly :-)

saghul:
I ran the job manually for the first time and we now have the first batch of unstable tags. Thanks a lot @hrenard !

PR #969, opened 2021-03-03
fix(k8s)

faust64:
Get node address from Pod metadata.
Requires k8s >= 1.7.

### DIFF
--- a/examples/kubernetes/README.md
+++ b/examples/kubernetes/README.md
@@ -14,7 +14,7 @@ Deploy the service to listen for JVB UDP traffic on all cluster nodes port 30300
 
 `kubectl create -f jvb-service.yaml`
 
-Now we can deploy the rest of the application. First modify the `DOCKER_HOST_ADDRESS` env value in deployment.yaml to point to one of nodes in your cluster (or load-balancer for all nodes if you have one), and then deploy it:
+Now we can deploy the rest of the application:
 
 `kubectl create -f deployment.yaml`
 
--- a/examples/kubernetes/deployment.yaml
+++ b/examples/kubernetes/deployment.yaml
@@ -112,7 +112,9 @@ spec:
             - name: XMPP_SERVER
               value: localhost
             - name: DOCKER_HOST_ADDRESS
-              value: <Set the address for any node in the cluster here>
+              valueFrom:
+                fieldRef:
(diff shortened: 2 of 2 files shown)


sapkra:
I used this solution for a while but after switching to a larger cluster with some more complex networking it stopped working. I had to switch to a LoadBalancer setup. I added a sidecar container which gets the LoadBalancer IP and adds it to the env var. I found no better way in my setup. But maybe the solution in your PR will work for most setups.

faust64:
Maybe some firewalling issue - when you tell about complex networking: something like network zones hosting kubernetes nodes, maybe only some of those being reachable to public? Or NetworkPolicies?

Just deployed my first copy 5 minutes ago, I definitely don't have any perspective on this.
Though editing a yaml setting some value that is otherwise known by the API is probably not necessary.

sapkra:
I tested this solution on a setup were I only had 1 master and 1 node on the same baremetal server...everything worked fine. After moving to Google Cloud VMs with 3 masters and 3 nodes it stopped working. If I remember correctly I got an internal networking address from `status.hostIP`, which obviously doesn't work.

faust64:
Right, with could integrated Kubernetes clusters, I think there's additional steps to open ports on your node addresses. Creating a LoadBalancer service would, in that case, request a specific IP from your cloud provider, and ensure traffic would flow.

I can reproduce with a GKE cluster. The have a service such as the following:

```
kube-system       default-http-backend          NodePort       10.31.246.181   <none>         80:31882/TCP                 3y96d
```

Say I curl one of my nodes public IP, on that port 31882, connection would timeout:

```
curl -vvv http://34.78.209.88:31882/ --connect-timeout 5
* Expire in 0 ms for 6 (transfer 0x563067e1be20)
* Expire in 10000 ms for 8 (transfer 0x563067e1be20)
* Expire in 5000 ms for 2 (transfer 0x563067e1be20)
*   Trying 34.78.20X.Y...
* TCP_NODELAY set
* Expire in 200 ms for 4 (transfer 0x563067e1be20)
* Connection timed out after 5001 milliseconds
* Closing connection 0
curl: (28) Connection timed out after 5001 milliseconds
```

saghul:
Closing. The k8s work will be moved to the jitsi-contrib org.

PR #970, opened 2021-03-03
fix(k8s): PodSecurityPolicy

faust64:
Dealing with clusters having PodSecurityPolicies: allow for root containers.

### DIFF
--- a/examples/kubernetes/README.md
+++ b/examples/kubernetes/README.md
@@ -14,6 +14,10 @@ Deploy the service to listen for JVB UDP traffic on all cluster nodes port 30300
 
 `kubectl create -f jvb-service.yaml`
 
+If PodSecurityPolicies were enabled, we would then install a PSP and Role for jitsi:
+
+`kubectl create -f rbac.yaml`
+
 Now we can deploy the rest of the application. First modify the `DOCKER_HOST_ADDRESS` env value in deployment.yaml to point to one of nodes in your cluster (or load-balancer for all nodes if you have one), and then deploy it:
 
 `kubectl create -f deployment.yaml`
--- a/examples/kubernetes/deployment.yaml
+++ b/examples/kubernetes/deployment.yaml
@@ -1,3 +1,10 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: jitsi
+  namespace: jitsi
+---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -143,3 +150,4 @@ spec:
               value: jvbbrewery
             - name: TZ
               value: America/Los_Angeles
+      serviceAccountName: jitsi
--- a/examples/kubernetes/rbac.yaml
+++ b/examples/kubernetes/rbac.yaml
@@ -1,0 +1,57 @@
+---
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: jitsi-privileged
+spec:
+  allowPrivilegeEscalation: true
+  fsGroup:
+    rule: RunAsAny
+  hostIPC: false
+  hostNetwork: true
+  hostPID: true
+  hostPorts:
+  - max: 65535
+    min: 0
+  privileged: true
+  runAsUser:
+    rule: RunAsAny
+  seLinux:
+    rule: RunAsAny
+  supplementalGroups:
+    rule: RunAsAny
+  volumes:
+  - configMap
+  - downwardAPI
+  - emptyDir
+  - persistentVolumeClaim
+  - projected
+  - secret
+---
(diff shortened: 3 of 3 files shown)


saghul:
@sapkra Thoughts on this?

faust64:
We could argue the PodSecurityPolicy doesn't require `hostNetwork: true`.

I left it in, assuming one may want to expose the JVB directly on nodes -- which would then allow you to start multiple instances of JVB, all exposing the same port, on different nodes.
Also, there's no guarantee the NodePort ports range would allow for exposing port 30300, while hostNetwork access would let you expose ports based on range limitations defined in the PodSecurityPolicy -- can be changed without reconfiguring your cluster.

Otherwise, that's pretty much Kubernetes default "privileged" PSP. Removing hostPath access, allowedCapabilities, allowedUnsafeSysctls and hostIPC access.

sapkra:
@faust64 Wouldn't it be better to make the containers run without root? I'm using them for years with the changes of my PR #192 in production. But I still have to test the latest changes after the most recent rebase.

faust64:
This PR is about getting the YAML in this repository to actually work out of the box, in Kubernetes clusters that use PodSecurityPolicy.
I don't pretend I would know how to patch those containers getting them to work unprivileged.

If your concern is OpenShift, I could provide some SecurityContextConstraints equivalent as well.
Though I wouldn't be much concerned: OKD4 is now a bloatware, still heavily bugged despite over 1 year of active developments, the CSI spec has been fixed recently coping with nodes losses (and still, there is a need for human intervention deleting VolumeAttachments https://github.com/ceph/ceph-csi/issues/740#issuecomment-770750182), not to mention FCOS lack of stability/random crashs, OperatorHub SNAFUs, ...
While the previous major branch latest release, OKD 3.11, based on Kubernetes 1.11, hasn't been updated in over a year or two: if you still have one, and are concerned about security, I would first recommend you to migrate to something that is still somewhat maintained.


Of course, regardless of OpenShift, it would be better for those containers to start without root privileges.
And sure, it's probably doable. Then again, not everything can, and I'm no Jitsi expert.

Given that the PR you mention has been there for over a year, something tells me we're not there yet.
Checking the diffs, nginx would bind on `8080`, then you need to put your JVB in another Pod, don't you? The Kubernetes deployment sample probably won't work, either nginx or JVB would fail to bind.

The JAVA_SYS_PROPS mentioning `/root` looks wrong (can you read/write over there? Wouldn't this be an issue?)

Accessing /dev/snd, and /dev/shm: you do setup hostPath mounts.
Regarding /dev/shm, you could actually use a Memory-backed emptyDir volume.
For /dev/snd, this would be an issue: root or non-root, you would still need a SecurityContextConstraint (OKD) and a PodSecurityPolicy (k8s), allowing container to access your host filesystem. Not sure if that helps, but pulseaudio should be able to provide sound input/output to unprivileged users, without such device 
( https://github.com/aau-zid/BigBlueButton-liveStreaming/pull/106/files )

No tini/dumb-init, rewriting files in /etc/ and /bin during startup, mentions of s6, ... let's be honest, those Dockerfiles won't comply with the best practices for building images any time soon.

And it's now been over two ours I'm reading diff, last comments suggest there are still some missing file in the jicofo container, ...

Still that's already a great start, kudos. But it's not just as easy as wishing it would work. And I won't fix this.

> All containers are running with USER 1001 (all users are always in the root group)

This is true in OpenShift, not necessarily in Kubernetes and derivatives. Depends of your PodSecurityPolicy. Something OpenShift admins tend to dismiss.

> The containers have to run with random user IDs which is the case in OpenShift/OKD for example

That's the case for OpenShift. Not in Kubernetes. The USER from your Dockerfile is actually observed, in regular Kubernetes, and most derivatives. Imagine, you could just grant permissions to uid 1001, instead of that whole recursive chmod g=u that is, in the end, not portable.
Aren't you complicating things, to get it working on a platform that has no future?
Are we sure this would work in Kubernetes, where hardening is a concern?

Either way, my PR is about getting the yamls already provided, to work in Kubernetes clusters that do have PSP. Nothing more.
If there's a way to get them to run without root privileges, that would be even better.
Ping me when you those containers to start.
In the meantime, despite the incomplete configuration provided in that repository, I managed to get my copy running. Thought I could share. If you don't care about this, then let's just close and leave it there.

sapkra:
@faust64 I really like the overall concept of an all-in-one system around Kubernetes, but I have to agree that the development speed is very slow at RedHat. At least it feels like that. I will migrate to a vanilla Kubernetes architecture in the next weeks.

Thank you for all the feedback - I closed my PR and will start completely from scratch some time in the future.

@saghul  I'm fine with this PR.

saghul:
Cheers folks!

PR #972, opened 2021-03-04
wrong skin name for etherpad

BlackWolfDEsign:
[2021-03-04 12:42:10.788] [ERROR] console - Skin path /opt/etherpad-lite/src/static/skins/"colibris" does not exist. Falling back to the default "colibris".
[2021-03-04 12:42:10.788] [INFO] console - Using skin "colibris" in dir: /opt/etherpad-lite/src/static/skins/colibris

### DIFF
--- a/env.example
+++ b/env.example
@@ -104,7 +104,7 @@ ETHERPAD_TITLE="Video Chat"
 ETHERPAD_DEFAULT_PAD_TEXT="Welcome to Web Chat!\n\n"
 
 # Name of the skin for etherpad
-ETHERPAD_SKIN_NAME="colibris"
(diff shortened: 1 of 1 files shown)


sapkra:
I think this comes from an already fixed bug within docker-compose. Can you update to docker-compose 1.26 or higher and try again?

saghul:
I think this is the only place where we quote env vars. How  about we get rid of them?

sapkra:
Ok, let's remove them everywhere. I will create a PR with all changes.

sapkra:
@saghul #973

saghul:
Thank you!

PR #974, opened 2021-03-05
add a env variable to enable/disable deep linking in config.js

prayagsingh:


web/rootfs/defaults/settings-config.js
```diff
@@ -304,3 +304,8 @@ if (!config.testing.hasOwnProperty('octo')) config.testing.octo = {};
 
 config.testing.capScreenshareBitrate = {{ $TESTING_CAP_SCREENSHARE_BITRATE }};
 config.testing.octo.probability = {{ $TESTING_OCTO_PROBABILITY }};
+
+{{ if $DISABLE_DEEP_LINKING -}}
```
saghul, line 308: Assign it unconditionally please, since we are using a default value on the template.

saghul (changes requested):

prayagsingh:
@saghul Please review

saghul (approved):

PR #980, opened 2021-03-10
Make a custom interface config possible

DTrierweiler:
For the `config.js` you can already add a `custom-config.js` in order to override config variables.
I would like to have this for the interface-config as well, since I need to override values like `DISPLAY_WELCOME_PAGE_CONTENT` that are not currently supported via ENV-variables.


saghul (approved):

saghul:
Cheers!

PR #982, opened 2021-03-11
Refresh prosody dockerfile

benbz:
This updates https://github.com/jitsi/docker-jitsi-meet/compare/master...matrix-org:jaywink/prosody-image-http-module against a newer upstream base and minimises how much we need to build

### DIFF
--- a/prosody/Dockerfile
+++ b/prosody/Dockerfile
@@ -29,6 +29,7 @@ RUN \
       libsasl2-modules-ldap \
       lua-basexx \
       lua-ldap \
+      lua-http \
       patch \
     && apt-dpkg-wrap apt-get install -t buster-backports -y \
       lua-cyrussasl \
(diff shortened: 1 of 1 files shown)


benbz:
Sorry, I incorrectly targetted this repo from a fork

saghul:
Hey, we'll be glad to take improvements Matrix needs :-)

benbz:
Heh, that's always good to hear, thanks!

We've a [pair of modules](https://github.com/matrix-org/prosody-mod-auth-matrix-user-verification) that currently use the lua `http` library, which is what we have forked this repo for. As per https://github.com/matrix-org/docker-jitsi-meet/pull/1 it may be that we remove the need for that dependency and so we can use these images. If the dependency doesn't go away, I'm happy to submit a PR to include the `http` library here, if you think it has value for yourself / others.

saghul:
I think it would be good if Matrix users can use our Docker images unchanged, just by configuring the needed modules with ENV variables (already possible) proven we have the necessary dependencies. This would help users get the latest and greatest versions and you folks not needing to maintain a fork.

Issue #989, opened 2021-03-19
Kubernetes example is broken on updated version of Jitsi


marner2:
A few things seem to be missing with the Kubernetes example, especially since the recent jitsi upgrade that turns on websockets by default (and doesn't seem to be able to turn them off).

As it stands, if I try to run the kubernetes scripts, I end up running into the problem where the websockets are trying to connect to "wss://localhost:8443." This is fixed by adding `PUBLIC_URL` to the web environment. However, even with that fix, something is still wrong with the forwarding because I'm getting errors connecting to those sockets (this seems to be the problem with #952 )

Can we either get the kubernetes example to use a compatible version of the docker containers or get it updated to actually have the right environment variables for the latest defaults from the jitsi docker containers? Thanks.

saghul:
As mentioned in https://github.com/jitsi/docker-jitsi-meet/blob/master/examples/README.md the k8s setup is community maintained. PRs are welcome.

rsoika:
I still did not manage to get it running on kubernetes. 
Can you provide an example of you tweeked deployments.yaml file?

saghul:
PSA: https://github.com/jitsi/docker-jitsi-meet/issues/991

rsoika:
My question was for @marner2 
I am still trying to get it working in my kubernetes cluster and if I succeed I would of course provide my results. But we also need a place to discuss....

saghul:
Sorry, my message was not directed at you but rather at those who have an interest in the k8s setup in general, who might be following this thread.

rsoika:
@saghul  I have no created a pull request containing a new concept of how to setup jitsi in kuberenetes with [Kustomize](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/).

https://github.com/jitsi/docker-jitsi-meet/pull/993

marner2:
@rsoika where were you getting errors? The purpose of these instructions to allow for simplified community transformation of the `docker-compose.yml` file into something that works with kubernetes (since the jitsi organization is only managing the `docker-compose.yml` file)

Have you checked out [KOMPOSE.md](https://github.com/jitsi/docker-jitsi-meet/blob/83e06562e56f66d4986143422a0f6da378ea25e6/examples/kubernetes/KOMPOSE.md) in #990? It's more complete.

[...]

PR #990, opened 2021-03-19
Add instructions for kompose w/ docker-compose.yml

marner2:
Addresses #989 by adding instructions for using [kompose](https://kompose.io/) to convert the latest `docker-compose.yml` file into a set of k8s deployments and services. This could be useful as a starting point for an actual set of k8s files that works with the latest published containers. Another suggestion would be to use `envFrom: - configMapRef: name: jitsi-config` in all of the deployments (as referenced [in this blog about postgres](https://humanitec.com/blog/handling-environment-variables-with-kubernetes)) which might be more robust to changes coming from the docker maintainers. Then we would only have to update a single configMap.yml file.

This also addresses a few points of confusion that I haven't seen addressed elsewhere.
1. If you set `DOCKER_HOST_ADDRESS` to your external ip address, you won't be able to attend meetings from the same LAN that your k8s nodes are on. There's also no reason to do that if you are using a STUN server. So just set a STUN server to figure out your external ip address, and use one of your k8s node's LAN address for this setting (or use that valueFrom: fieldRef: fieldPath: status.hostIP as mentioned in #969 )
2. You need to expose port 9090 on the jvb pods, or else the websocket connection will break.


saghul:
Closing as this will be published in https://github.com/jitsi-contrib

Issue #991, opened 2021-03-25
PSA: New GH org for community maintained 3rd party Jitsi projects


saghul:
Hey all!

We just announced a thing: https://community.jitsi.org/t/introducing-the-jitsi-contrib-github-organization/96409

I think there are a number of people here that would he happy to contribute to a k8s setup, helm chart, etc. This new org could be a home for that.

If there is enough interest we could move the projects currently in examples/ to standalone projects in this new org, fully managed by the members of the community who choose to steer them.

Any feedback is welcome!

Thanks @sapkra for working with us in shaping this.

saghul:
Perhaps a mention to this on the README here would be nice. Up for a PR?

rsoika:
See: https://github.com/jitsi/docker-jitsi-meet/pull/998

rsoika:
Note that I just renamed the repo form 'kubernetes' to 'jitsi-kubernetes' to avoid name conflicts

https://github.com/jitsi-contrib/jitsi-kubernetes

saghul:
Closing since this issue has run its course. Further activity continues over at jitsi-contrib!

PR #993, opened 2021-03-25
Kubernetes Setup using kustomize

rsoika:
Hi,

this is my suggestion of how we could improve the configuration for Kuberentes based on kustomize.

The core configuraition is placed on docker-jitsi-meet. And users can customize the setup with [Kustomize](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/) declaring new objects or providing custom environment variables by overwriting the deployment.yaml file. 

I have explained the setup in the README.md file

@saghul  : Can you take a look at this and give you your feedback if it is worse to be merged?


rsoika:
If you are unhappy with this pull request I can create a new one with less commits?

prayagsingh:
@saghul @sapkra  I have a suggestion here. Instead of changing `examples/kubernetes` files, we can have sub-directories under `example/kubernetes` like `kubernetes/simple-deployment`, `kubernetes/deployment-with-kustomize` and finally `kubernetes/deployment-with-helm`. WDYT?

rsoika:
@saghul @sapkra may I provide a new pull request using the sub-directory layout as suggested by @prayagsingh 

    /examples/kubernetes/deployment-with-kustomize

I would like to share my setup and I also would like to discuss the websocket problem I am still facing

saghul:
@rsoika That's not the direction we'd like to follow. See #991

rsoika:
I will provide a new pull request explaining how to fix the websocket problem

PR #997, opened 2021-03-30
provide a deployment example for Kubernetes Kustomize

rsoika:
This is an example how to setup jitsi with Kubernetes Kustomize. 
It includes the base configuration and a description how custom settings and an Ingress can be applied

### DIFF
--- a/examples/kubernetes/kustomize/010-jvb-service.yaml
+++ b/examples/kubernetes/kustomize/010-jvb-service.yaml
@@ -1,0 +1,17 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    service: jvb
+  name: jvb-udp
+  namespace: jitsi
+spec:
+  type: NodePort
+  externalTrafficPolicy: Cluster
+  ports:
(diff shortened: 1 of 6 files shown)


rsoika:
Note that I created a new sub directory for kubernetes as mentioned by @prayagsingh 

https://github.com/jitsi/docker-jitsi-meet/pull/993#issuecomment-808181258

saghul:
I think this should go into https://github.com/jitsi-contrib as a standalone repo. What do you think? I can add you to the org so you can create the repo and work on it there.

rsoika:
of course. If you agree we can name the repo jitsi-contrib/kubernetes and bundle all kubernetes related stuff there

saghul:
Looks like there are many ways to skin the cat :-) https://github.com/jitsi-contrib/meta/issues/2

saghul:
Just invited you to the org @rsoika . Feel free to self-organize, try stuff out and see what's the best option going forward. IMHO starting with separate projects is a good start, there is always time to consolidate later.

Issue #568, opened 2020-05-08
Etherpad integration not working


mathieubrun:
In the file https://github.com/jitsi/docker-jitsi-meet/blob/master/web/rootfs/defaults/meet.conf
The "Host" header is added twice :
````
{{ if .Env.ETHERPAD_URL_BASE }}
# Etherpad-lite
location /etherpad/ {
    proxy_http_version 1.1;
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection 'upgrade';
    proxy_set_header Host $host;
    proxy_cache_bypass $http_upgrade;

    proxy_pass {{ .Env.ETHERPAD_URL_BASE }}/;
    proxy_set_header X-Forwarded-For $remote_addr;
    proxy_buffering off;
    proxy_set_header Host {{ .Env.XMPP_DOMAIN }};
}
{{ end }}
````

Resulting in the error "400 Bad Request: too many Host headers" when opening the etherpad document.

Removing one of the headers results in a 404 either on the meet server, or the etherpad server.
Removing both Host headers solves the issue. 

I can make a PR if requested :)

cboylan:
I can confirm that this is a problem as well. We're trying to work around it by bind mounting in a modified copy of meet.conf so that the generated config is correct here: https://review.opendev.org/c/opendev/system-config/+/781154/1/playbooks/roles/jitsi-meet/files/meet.conf#84 Ideally we wouldn't need to do that as the vendoring is likely to become stale over time.

saghul:
Oops. A PR would be most welcome.

PR #999, opened 2021-03-31
fix: removed duplicate host headers

mathieubrun:
See : https://github.com/jitsi/docker-jitsi-meet/issues/568

### DIFF
--- a/web/rootfs/defaults/meet.conf
+++ b/web/rootfs/defaults/meet.conf
@@ -81,12 +81,10 @@ location /etherpad/ {
     proxy_http_version 1.1;
     proxy_set_header Upgrade $http_upgrade;
     proxy_set_header Connection 'upgrade';
-    proxy_set_header Host $host;
     proxy_cache_bypass $http_upgrade;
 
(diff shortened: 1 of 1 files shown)


web/rootfs/defaults/meet.conf
```diff
@@ -81,12 +81,10 @@ location /etherpad/ {
     proxy_http_version 1.1;
     proxy_set_header Upgrade $http_upgrade;
     proxy_set_header Connection 'upgrade';
-    proxy_set_header Host $host;
     proxy_cache_bypass $http_upgrade;
 
     proxy_pass {{ .Env.ETHERPAD_URL_BASE }}/;
     proxy_set_header X-Forwarded-For $remote_addr;
     proxy_buffering off;
-    proxy_set_header Host {{ .Env.XMPP_DOMAIN }};
```
saghul, line 90: Is removing this necessary?

mathieubrun: Yes, Removing one of the headers results in a 404 either on the meet server, or the etherpad server.

saghul (approved):

PR #1000, opened 2021-04-01
Update docker-compose.yml

Saalbuerger:
Whitout this, jicofi will Not start

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -189,6 +189,7 @@ services:
             - ENABLE_CODEC_H264
             - ENABLE_RECORDING
             - ENABLE_SCTP
+            - JICOFO_COMPONENT_SECRET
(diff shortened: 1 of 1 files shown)


saghul:
This is no longer needed, master is now using client proxy connections instead of a component connection.

Saalbuerger:
Yes, i now it but new install will not start. Only an error: FATAL ERROR: Jicofo auth password must be set'

- fresh install
- gen. Password 
- Docker-compose up -d
- docker restart whit error

saghul:
You should not use master just like that. Use a release tag, which pins images to config.

Current master works with the unstable images.

Saalbuerger:
Hi @saghul - i get it, thanks.

but the quick start Guide is now broken (the way i used it befor), only whit the last Part (Build Instruction) it will ne work.

PR #1001, opened 2021-04-02
disable octo by default

prayagsingh:
In the jicofo reference.conf file, Octo is disabled by default hence added an environment variable to enable/disable octo in jicofo.conf.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -206,6 +206,7 @@ services:
             - JVB_BREWERY_MUC
             - MAX_BRIDGE_PARTICIPANTS
             - OCTO_BRIDGE_SELECTION_STRATEGY
+            - TESTING_OCTO_PROBABILITY
             - TZ
             - XMPP_DOMAIN
             - XMPP_AUTH_DOMAIN
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -187,6 +187,7 @@ services:
             - ENABLE_CODEC_VP8
             - ENABLE_CODEC_VP9
             - ENABLE_CODEC_H264
+            - ENABLE_OCTO
             - ENABLE_RECORDING
             - ENABLE_SCTP
             - JICOFO_AUTH_USER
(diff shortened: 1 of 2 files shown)


prayagsingh:
@saghul 

I have not added conditional check like `{{ if $ENABLE_OCTO }}` because even if the value of `octo.enabled=false` then it is still looking for `octo.id`. 
**Snippet from jicofo logs**
```
Apr 02, 2021 2:55:25 PM org.jitsi.utils.logging2.LoggerImpl log
INFO: Initialized octoId=1
```

**snippet jicofo.conf**
```
 octo {
      // Whether or not to use Octo. Note that when enabled, its use will be determined by
      // $jicofo.bridge.selection-strategy. There's a corresponding flag in the JVB and these
      // two MUST be in sync (otherwise bridges will crash because they won't know how to
      // deal with octo channels).
      enabled = false

      id = "1"
    }
```

L3o-pold:
IMO it's better to add a new configuration variable than using `TESTING_OCTO_PROBABILITY` that could be a float from 0 to 1.

prayagsingh:
@L3o-pold  I'm just trying to avoid a new env variable. `TESTING_OCTO_PROBABILITY` is already defined for service `web` to enable/disable Octo hence using the same here.

But Yes, I'm also in favor of changing `TESTING_OCTO_PROBABILITY` to `ENABLE_OCTO`.

@saghul WDYT?

L3o-pold:
@prayagsingh enabling Octo is not the same thing that the probability it will be used.

prayagsingh:
> 
> 
> @prayagsingh enabling Octo is not the same thing that the probability it will be used.

Totally agree with you. Let's wait for @saghul response.

saghul:
Agreed, let's have 2.

prayagsingh:
sure @saghul. I'm going to name it `ENABLE_OCTO`.

saghul:
Sounds good!

saghul (approved):

PR #1005, opened 2021-04-05
Provide component secret to jicofo - fix fatal jicofo startup error

fntc:
Connection to a session fails after setting up a fresh docker installation.

Jicofo failed with:
FATAL ERROR: Jicofo component secret and auth password must be set

so this commit adds 
- JICOFO_COMPONENT_SECRET
in jicofo.environment


prayagsingh:
Are you trying to use main repo with stable tag? If yes then it won't work. To use the main branch, you have to build your own docker images.

OR 

Clone the repo using stable tag.

PR #1008, opened 2021-04-07
added octo for jvb

prayagsingh:


jvb/rootfs/defaults/jvb.conf
```diff
@@ -51,6 +52,16 @@ videobridge {
             port = 9090
         }
     }
+
+    {{ if $ENABLE_OCTO -}}
+    octo {
+        enabled = true
```
prayagsingh, line 58: @saghul I'm not using `enabled = {{{ $ENABLE_OCTO }}` here because in the [reference.conf](https://github.com/jitsi/jitsi-videobridge/blob/3fa0b7c2e3e69b783ad12e89cc94bfd0175f384b/jvb/src/main/resources/reference.conf#L115) its value is `false`  hence no need to use env variable here.

docker-compose.yml
```diff
@@ -244,6 +245,10 @@ services:
             - JVB_WS_DOMAIN
             - JVB_WS_SERVER_ID
             - PUBLIC_URL
+            - JVB_NAT_HARVESTER_LOCAL_ADDRESS
```
saghul, line 248: Prefix these 2 with octo, since the names suggest they could be used for regular communication.

saghul:
@bgrozev @bbaldino can you take a quick peek here?

docker-compose.yml
prayagsingh: I'm thinking of using `JVB_OCTO_BIND_ADDRESS` and `JVB_OCTO_PUBLIC_ADDRESS` instead. WDYT?

saghul: Sounds good!

saghul (approved):

PR #1012, opened 2021-04-10
allow to configure shard name using env variable

prayagsingh:
Configure Octo shard name and Region using env vars.


saghul (approved):

prayagsingh:
@saghul If we set `config.deploymentInfo.shard = shard-0` then we have to add the `X-Jitsi-Shard ` header in `meet.conf` else it would result in **Detected that shard changed from undefined to null** error and meeting crashed. 

Below is the snippet of how I'm using it. 
```
{{ if .Env.DEPLOYMENTINFO_SHARD }}
add_header X-Jitsi-Shard {{ .Env.DEPLOYMENTINFO_SHARD }};
{{ end }}
```
Adding above config works fine for me. Please let me know if I create a new PR for this or shall I add it in this PR only.

saghul:
Sounds good, add it to this PR please.

saghul (approved):

PR #1013, opened 2021-04-12
JVB HTTP servers should be configurable

DTrierweiler:
Those 4 new envs would enable the http server interface to be configured.
This would be especially helpful in a kubernetes environment, where I use `hostNetwork: true` but I don't want the private http server to be exposed to the public.

Default values would stay the same.

Cheers.

### DIFF
--- a/jvb/rootfs/defaults/jvb.conf
+++ b/jvb/rootfs/defaults/jvb.conf
@@ -45,12 +45,12 @@ videobridge {
     }
     http-servers {
         private { 
-          host = "{{ .Env.JVB_HTTP_SERVER_PRIVATE_ADDRESS | default "0.0.0.0" }}"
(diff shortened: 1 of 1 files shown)


jvb/rootfs/defaults/jvb.conf
```diff
@@ -45,11 +45,12 @@ videobridge {
     }
     http-servers {
         private { 
-          host = 0.0.0.0
+          host = "{{ .Env.JVB_HTTP_SERVER_PRIVATE_ADDRESS | default "0.0.0.0" }}"
```
saghul, line 48: I don't think this needs quotation marks.

DTrierweiler: I copied the style from the other config options:
https://github.com/jitsi/docker-jitsi-meet/blob/master/jvb/rootfs/defaults/jvb.conf#L59

But I'll remove the quotation marks.

PR #1014, opened 2021-04-12
Add codecs to config, fixing issue with simulcast

### LINKED ISSUE #981: Codec issues: H264 vs VP8
I've noticed some interesting behaviour. Lately, I've been experimenting with the codec settings in the deployment, notably the following three environment variables:

```
ENABLE_CODEC_VP8
ENABLE_CODEC_VP9
ENABLE_CODEC_H264
```

Right now, I have it set to the following:

```
ENABLE_CODEC_VP8="false"
ENABLE_CODEC_VP9="false"
ENABLE_CODEC_H264="true"
```

The reason here being that we use Jitsi mostly for voice and screen sharing, and the occasional video. I've noticed that the H264 coding yields the sharpest screen share image without flickering.

Using VP8 only:
- Screen sharing is pixelated. It gets sharp after about 10 seconds, but returns to pixels when the screen is scrolled.
- Camera video is very sharp.

Using H264 only:
- Screen sharing is very sharp, and even when scrolling the screen share remains very sharp.
- Camera video is very pixelated. Even though my Jitsi says I'm uploading 1280x720 at 30 frames per seconds, everybody else sees video that looks like 640x360 or even less sharp.

Using VP9 only:
For some reason, this falls back to VP8.

Enabling all three codecs:
This also falls back to VP8.

Since we need to have sharp screen sharing and camera video is just for the occasional lulz, the codec is set to H264, and VP8 and VP9 are off.

Now, this is where it gets interesting: this is all using the Jitsi Meet Electron desktop app. Also when using a Chrome-based browser, the behaviour is like this. But when I switch to Firefox, even when using H264 the video is crystal clear in 720p and also the screen sharing is crisp. So it seems like Firefox is compressing the video differently? But using the Electron app is much more convenient, so Firefox rarely gets used.

So I have some questions:
- Why is H264 so much better than VP8 for screen sharing and horrible for video (in Chrome and Electron)?
- Why is VP8 so much better for video and horrible for screen sharing (in Chrome and Electron)?
- Is it possible to have a different codec when using the camera as opposed to when sharing the computer screen (to use the best of both worlds)?
- Why is H264 video in Firefox so much better than H264 video in Chrome (and Electron)?

cedricroijakkers:
Added environment variables to the default Docker configuration.
Added a little note to disable simulcast when H264 is used.

This fixes #981

### DIFF
--- a/env.example
+++ b/env.example
@@ -376,3 +376,11 @@ RESTART_POLICY=unless-stopped
 
 # Authenticate using external service or just focus external auth window if there is one already.
 # TOKEN_AUTH_URL=https://auth.meet.example.com/{room}
+
+# Codec configuration
+ENABLE_CODEC_VP8: "true"
+ENABLE_CODEC_VP9: "true"
+ENABLE_CODEC_H264: "true"
+
+# If you wish to use h.264 as codec, then you must disable simulcast below since JVB does not support it with h.264
+ENABLE_SIMULCAST: "true"


saghul:
I don't want to add every single option to env.example. In fact I plan to reduce it to the bare minimum, eventually. Can you please add these to the handbook instead? https://jitsi.github.io/handbook/docs/devops-guide/devops-guide-docker#configuration

Issue #1018, opened 2021-04-16
Cannot start conferences anymore if  LDAP Auth is active with latest stable 5765


Twilek-de:
Everything seems to work, but then the login window gets stuck on "Getting session ID" (or something similar, as I see the german translation). In the logs it says that LDAP Auth was successfull. When I turn off Auth alltogether things work fine again. Same setup worked with the stable version before that.

Twilek-de:
Tried JWT Auth. Shows a successfull login in the logs und then a login prompt on the page. I think authorisation is seriously broken in this release....

tob123:
i have the same issue using internal authentication.
for me it is fixed by adjusting /config/jicofo.conf by adding 'client-proxy = "focus.meet.jitsi"' to the xmpp client configuration.
this is also referenced here for the debian package: https://github.com/jitsi/jitsi-meet/pull/8439/files

saghul:
Care to make a PR?

aminvakil:
Same problem here using jwt auth and rocketchat 3.11.3

aminvakil:
Adding `client-proxy = "focus.meet.jitsi"` in
https://github.com/jitsi/docker-jitsi-meet/blob/8d2a29f2864cbff5faeeb73ae8e06fbb3bed08b8/jicofo/rootfs/defaults/jicofo.conf#L132-L140
did not fix the problem for me neither.

TheSleepySlee:
> i have the same issue using internal authentication.
> for me it is fixed by adjusting /config/jicofo.conf by adding 'client-proxy = "focus.meet.jitsi"' to the xmpp client configuration.
> this is also referenced here for the debian package: https://github.com/jitsi/jitsi-meet/pull/8439/files

hi, can you details how you are fixing the issue?
I tried to edit .jitsi-meet-cfg/jicofo/jicofo.conf but it seems it gets replaced when I start the container again
Editing docker-jitsi-meet/jicofo/rootfs/defaults/jicofo.conf and doing docker-compose up don't work either

thanks

daradib:
Editing `.jitsi-meet-cfg/jicofo/jicofo.conf` as described fixed internal auth for me in stable-5765. Thanks!

@aminvakil  @TheSleepySlee try restarting jicofo without restarting the container to avoid regenerating the file:

    docker-compose exec jicofo s6-svc -r /var/run/s6/services/jicofo

[...]

PR #1019, opened 2021-04-17
Update jicofo

tob123:
this is a pull request to fix jicofo in case internal authentication is used.
see also https://github.com/jitsi/docker-jitsi-meet/issues/1018


saghul:
Thanks! I'll run some tests and make a -1 release.

Javinator9889:
As a patch before the new release is available, I have just posted an intermediate solution for the `stable-5765` tag - see #1018 (https://github.com/jitsi/docker-jitsi-meet/issues/1018#issuecomment-821800483)

saghul (approved):

PR #1023, opened 2021-04-17
Add DEFAULT_LANGUAGE environment variable for web config

pruje:
To set defaultLanguage value in web/config.js, I added support of DEFAULT_LANGUAGE environment variable.

I hope everything is OK. I tried to put it in the same place than in config.js
Thanks!

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -57,6 +57,7 @@ services:
             - ENABLE_PREJOIN_PAGE
             - ENABLE_WELCOME_PAGE
             - ENABLE_CLOSE_PAGE
+            - DEFAULT_LANGUAGE
             - ENABLE_RECORDING
             - ENABLE_REMB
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -57,6 +57,7 @@ services:
             - ENABLE_PREJOIN_PAGE
             - ENABLE_WELCOME_PAGE
             - ENABLE_CLOSE_PAGE
+            - DEFAULT_LANGUAGE
```
saghul, line 60: Can you please put it alphabetically sorted ?

saghul (changes requested):
Minor comment, LGTM otherwise

docker-compose.yml
pruje: OK, it's done.

saghul (approved):

PR #342, opened 2020-03-31
Add support for sentry logging

sapkra:
This is adding the possibility to track warnings and errors for jicofo, jigasi and jvb with [Sentry](https://sentry.io).

depends on:
* https://github.com/jitsi/jigasi/pull/247
* https://github.com/jitsi/jicofo/pull/449
* https://github.com/jitsi/jitsi-videobridge/pull/1152

PR #1024, opened 2021-04-19
misc: add support for sentry logging

sapkra:
#342

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -210,6 +210,9 @@ services:
             - JVB_BREWERY_MUC
             - MAX_BRIDGE_PARTICIPANTS
             - OCTO_BRIDGE_SELECTION_STRATEGY
+            - SENTRY_DSN="${JICOFO_SENTRY_DSN}"
+            - SENTRY_ENVIRONMENT
+            - SENTRY_RELEASE
             - TZ
             - XMPP_DOMAIN
             - XMPP_AUTH_DOMAIN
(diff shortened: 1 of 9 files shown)


sapkra:
@saghul Can we merge this one again?

saghul:
I just took another look. I don't like having uncommented values in env.sample, because people won't update that, so I propose we provide the defaults in another way. WDYT about the commit I just added?

docker-compose.yml
```diff
@@ -210,7 +210,7 @@ services:
             - JVB_BREWERY_MUC
             - MAX_BRIDGE_PARTICIPANTS
             - OCTO_BRIDGE_SELECTION_STRATEGY
-            - SENTRY_DSN="${JICOFO_SENTRY_DSN}"
+            - SENTRY_DSN="${JICOFO_SENTRY_DSN:-0}"
```
sapkra, line 213: `SENTRY_DSN` is not something I define - it's a predefined config variable of sentry. So setting it to `0` is not like it is defined in their docs.

saghul: But if we take the fallback sentry won't be enabled, so that's ok, isn't it? I want to avoid the questions when people see warnings because they are undefined, when didn't configure any sentry related stuff.

sapkra: @saghul It's fine, you can merge it if you want.

saghul: Alright then! I'll make sure to land this next week.

saghul (approved):

PR #1038, opened 2021-05-03
Fix multi domain support for web UI

### LINKED ISSUE #927: Multi domain suppor behind reverse proxyt?
Hi,

with an older version it was possible to modify the host header to backends. Just rewrite the host header from second domain to the first one was enough to get it working. With the latest version it fails. I don't know if it related to changes in jitsi docker. But maybe it's just a problem with the domain name the websocket uses?

I the the PRIMARY _DOMAIN instead of the used SECONDARY_DOMAIN in chrome developer tools.
```
WebSocket connection to 'wss://<PRIMARY_DOMAIN>/xmpp-websocket?room=test123' failed: Error during WebSocket handshake: Unexpected response code: 403
_connect @ lib-jitsi-meet.min.js?v=4628:1
connect @ lib-jitsi-meet.min.js?v=4628:1
_interceptConnectArgs @ lib-jitsi-meet.min.js?v=4628:25
connect @ lib-jitsi-meet.min.js?v=4628:1
_connect @ lib-jitsi-meet.min.js?v=4628:1
connect @ lib-jitsi-meet.min.js?v=4628:1
c.connect @ lib-jitsi-meet.min.js?v=4628:10
e @ app.bundle.min.js?v=4628:83
(anonymous) @ app.bundle.min.js?v=4628:83
d @ app.bundle.min.js?v=4628:83
h @ app.bundle.min.js?v=4628:83
W @ app.bundle.min.js?v=4628:113
createInitialLocalTracksAndConnect @ app.bundle.min.js?v=4628:113
init @ app.bundle.min.js?v=4628:113
Logger.js:154 2021-01-31T12:44:59.409Z [JitsiMeetJS.js] <Object.getGlobalOnErrorHandler>:  UnhandledError: Strophe: Websocket error [object Event] Script: null Line: null Column: null StackTrace:  Error: Strophe: Websocket error [object Event]
```

Would it help to use the given domain name instead of the first / primary one for websocket connection?

With older versions it was easy to get the second domain working! Just rewrite the host header at the reverse proxy before pass it to the backend. Example with caddy swarm proxy configured by labels:
```
caddy: "<PRIMARY _DOMAIN> <SECONDARY_DOMAIN>"
caddy.reverse_proxy: "{{upstreams 80}}"
caddy.reverse_proxy.header_up: "Host <PRIMARY _DOMAIN>"
caddy.log.output: "stdout"
```
So backend gets <PRIMARY _DOMAIN> also if <SECONDARY_DOMAIN> was used from the user. Worked fine for me some time, but fails since I updated to latest jitsi docker version.

MartijnHols:
This fixes an issue where the xmpp-websocket was only available on the primary domain (the PUBLIC_URL). This change makes the URL relative to the domain that hosts Jitsi Web.

Partially fixes #927

### DIFF
--- a/web/rootfs/defaults/system-config.js
+++ b/web/rootfs/defaults/system-config.js
@@ -4,7 +4,6 @@
 {{ $ENABLE_SUBDOMAINS := .Env.ENABLE_SUBDOMAINS | default "false" | toBool -}}
 {{ $ENABLE_XMPP_WEBSOCKET := .Env.ENABLE_XMPP_WEBSOCKET | default "1" | toBool -}}
 {{ $JICOFO_AUTH_USER := .Env.JICOFO_AUTH_USER | default "focus" }}
-{{ $PUBLIC_URL_DOMAIN := .Env.PUBLIC_URL | default "https://localhost:8443" | trimPrefix "https://" | trimSuffix "/" -}}
 {{ $XMPP_AUTH_DOMAIN := .Env.XMPP_AUTH_DOMAIN -}}
 {{ $XMPP_DOMAIN := .Env.XMPP_DOMAIN -}}
 {{ $XMPP_MUC_DOMAIN := .Env.XMPP_MUC_DOMAIN -}}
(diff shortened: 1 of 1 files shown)


saghul:
This will not work when we enable WebSocket on mobile since there is no window.location.

MartijnHols:
In that case without a location how does the /http-bind path in that same file resolve? I might be able to use the same logic. 

I could change this to a conditional with the public url as fallback, but then secondary URLs would still be broken for mobile.

saghul:
BOSH works because you can set a relative URL but I don't it's possible to do that with WS URLs since you need to specify the transport.

PR #1040, opened 2021-05-04
Add missing websockets config to example .env

w4tsn:
The example environment file lacks the ENABLE_XMPP_WEBSOCKETS variable
which has to be configured in order for video calls to work. This should
be part of the initial example configuration so new folks won't get
confused and frustrated.

### DIFF
--- a/env.example
+++ b/env.example
@@ -44,6 +44,9 @@ TZ=UTC
 # Public URL for the web service (required)
 #PUBLIC_URL=https://meet.example.com:8443
 
+# Enable websocket connections required for participating video calls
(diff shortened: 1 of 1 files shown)


saghul:
Those are not needed since they already default to true.

Are you running the compose setup or something else?

w4tsn:
Yes, it seems that probably I missed some network configuration or was too impatient for the stack to start up which caused a 403 on the websocket connection. Based on an already closed issue in a kubernetes environment I added those configs and it worked. Thought I had verified it afterwards, but apparently today it works without those options set explicitly.

PR #1043, opened 2021-05-09
Added static file cache

supu2:
i added static file cache for improve performance


saghul:
This was already added, sorry we didn't notice the PR.

PR #1051, opened 2021-05-15
Add env variables for configuring videocodec

prayagsingh:
added env variables for configuring videocodecs


cloudslash7:
Screen Shot 2021-05-16 at 11 20 17 PM

I'm still seeing VP8. It IS p2p, though. Could that be the problem?

prayagsingh:
@saghul please review and let me know if you need any changes in env variables naming.

saghul:
Thanks for doing this! I'd rather not have the ENABLE_VIDEOQUALITY flag and just apply the values all the time, IF set. This means we should not have defaults, at least for the bitrates. I can think of us changing the default in lib-jitsi-meet and then if a value is set apps wouldn't benefit of the new value.

prayagsingh:
Thanks for the feedback @saghul 
I'll remove the default values.

prayagsingh:
@saghul Please review.

saghul:
Sorry I was on parental leave and forgot about this one. Did you intentionally close it?

prayagsingh:
yes. Some of the commits were showing `ghost` username because of some issue in git-config(configured using a different email ID and this email ID doesn't exist on github). I will create a new PR with these changes only.

PR #1074, opened 2021-06-12
configure external TURN server

tbutter:
Add 3 env variables to allow configuring an external TURN server

### DIFF
--- a/env.example
+++ b/env.example
@@ -354,6 +354,13 @@ JIBRI_STRIP_DOMAIN_JID=muc
 # Directory for logs inside Jibri container
 JIBRI_LOGS_DIR=/config/logs
 
+# Configure an external TURN server
+# TURN_CREDENTIALS="secret"
+# TURN_HOST=turnserver.example.com
+# TURN_PORT=443
+# TURNS_HOST=turnserver.example.com
+# TURNS_PORT=443
+
 # Disable HTTPS: handle TLS connections outside of this setup
 #DISABLE_HTTPS=1
 
--- a/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
+++ b/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
@@ -22,6 +22,26 @@ http_default_host = "{{ .Env.XMPP_DOMAIN }}"
 
 {{ $ENABLE_XMPP_WEBSOCKET := .Env.ENABLE_XMPP_WEBSOCKET | default "1" | toBool }}
 {{ $PUBLIC_URL := .Env.PUBLIC_URL | default "https://localhost:8443" -}}
+{{ $TURN_PORT := .Env.TURN_PORT | default "443" }}
+{{ $TURNS_PORT := .Env.TURNS_PORT | default "443" }}
+
+{{ if .Env.TURN_CREDENTIALS }}
+external_service_secret = "{{.Env.TURN_CREDENTIALS}}";
+{{ end }}
+
+{{ if or .Env.TURN_HOST .Env.TURNS_HOST }}
+external_services = {
+  {{ if .Env.TURN_HOST }}
+     { type = "turn", host = "{{ .Env.TURN_HOST }}", port = {{ $TURN_PORT }}, transport = "tcp", secret = true, ttl = 86400, algorithm = "turn" }
+  {{ end }}
+  {{ if and .Env.TURN_HOST .Env.TURNS_HOST }}
+  ,
+  {{ end }}
+  {{ if .Env.TURNS_HOST }}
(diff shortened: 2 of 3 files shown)


prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -22,6 +22,22 @@ http_default_host = "{{ .Env.XMPP_DOMAIN }}"
...
+{{ end }}
+
+{{ if .Env.TURN_HOST }}
+
+turncredentials_host = "{{ .Env.TURN_HOST }}";
+turncredentials_port = "{{ $TURN_PORT }}";
+
+turncredentials = {
+  { type = "turn", host = "{{ .Env.TURN_HOST }}", port = "{{ $TURN_PORT }}", transport = "tcp" },
+  { type = "turns", host = "{{ .Env.TURN_HOST }}", port = "{{ $TURN_PORT }}", transport = "tcp" }
```
saghul, line 38: I think we should have a separate env bar for the TURNS port.

tbutter: I think those usually are at the same port. Futhermore there is only one turncredentials_port option - i don't know if that would work with two ports.

damencho:
I think we should use the external_services module as the debian package 
https://github.com/jitsi/jitsi-meet/blob/b2985934f7ea43b28c4fbc33bea855e74fa244ab/doc/debian/jitsi-meet-prosody/prosody.cfg.lua-jvb.example#L6
https://github.com/jitsi/jitsi-meet/blob/b2985934f7ea43b28c4fbc33bea855e74fa244ab/doc/debian/jitsi-meet-prosody/prosody.cfg.lua-jvb.example#L52

tbutter:
> I think we should use the external_services module as the debian package
> https://github.com/jitsi/jitsi-meet/blob/b2985934f7ea43b28c4fbc33bea855e74fa244ab/doc/debian/jitsi-meet-prosody/prosody.cfg.lua-jvb.example#L6
> https://github.com/jitsi/jitsi-meet/blob/b2985934f7ea43b28c4fbc33bea855e74fa244ab/doc/debian/jitsi-meet-prosody/prosody.cfg.lua-jvb.example#L52

i can change it. what are the advantage of using that module?

damencho:
Supports extdisco:2 https://github.com/jitsi/jitsi-meet/blob/b2985934f7ea43b28c4fbc33bea855e74fa244ab/resources/prosody-plugins/mod_external_services.lua#L236
And this module will be in prosody 0.12 by default.

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
tbutter: By using the external_services module it was possible to set separate turn/turns servers

tbutter:
@damencho thanks for the explanation. changed it to use that module.

env.example
```diff
@@ -354,6 +354,13 @@ JIBRI_STRIP_DOMAIN_JID=muc
 # Directory for logs inside Jibri container
 JIBRI_LOGS_DIR=/config/logs
 
+# Configure an external TCP turn(s) server
```
saghul, line 357:
```suggestion
# Configure an external TURN server
```

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -91,6 +111,9 @@ VirtualHost "{{ .Env.XMPP_DOMAIN }}"
         "ping";
         "speakerstats";
         "conference_duration";
+        {{ if .Env.TURN_HOST }}
```
saghul, line 114: also check  if TURNS_HOST is set?

env.example
tbutter: thanks - fixed it

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
tbutter: thanks - fixed it

saghul (approved):
LGTM. @damencho can you also give it a quick read please?

seuffert:
i was confused and surprised that this was missing in the current implementation, thanks @tbutter ! i am currently using this PR without problems.

damencho:
> LGTM. @damencho can you also give it a quick read please?

LGTM

mayconmesquita:
@tbutter great implementation!
Is it possible to set a username, or only credential/pass?

tbutter:
@mayconmesquita credentials is the secret for computing temporary passwords for arbitrary usernames as defined in https://datatracker.ietf.org/doc/html/draft-uberti-behave-turn-rest-00 section 2.2.

I don't know if jitsi supports setting a static username/password, but I wouldn't recommend it in any case. The username/password is transmitted to the users and should therefore only be temporary valid.

mayconmesquita:
@tbutter Now I understand why there is no username. Thanks for the response!

PR #1079, opened 2021-06-23
Fix acme.sh pre and post hooks

driehuis:
On my system, letsencrypt did not auto-renew the acme certificate. Desk checking the code, I found the the pre- and post hooks, that stop nginx during the acme processing, check for the existence of a directory with `test -f`. That should probably be `test -d` or `test -e`. After I manually re-ran with `-d` in my container, the certificate was renewed.

I noticed that acme.sh --install writes a cronjob to a non-persistent location in the container, and that this cron job won't work anyway because it doesn't seem to have the pre/post hooks. I did not address those issues because if that amount of work gets put into fixing this, it would probably be better to rework the mechanism to use a `.well-known` directory served by nginx, which seems to be on the roadmap.

### DIFF
--- a/web/rootfs/etc/cont-init.d/10-config
+++ b/web/rootfs/etc/cont-init.d/10-config
@@ -25,8 +25,8 @@ if [[ $DISABLE_HTTPS -ne 1 ]]; then
             $STAGING \
             --issue \
             --standalone \
-            --pre-hook "if [[ -f /var/run/s6/services/nginx ]]; then s6-svc -d /var/run/s6/services/nginx; fi" \
-            --post-hook "if [[ -f /var/run/s6/services/nginx ]]; then s6-svc -u /var/run/s6/services/nginx; fi" \
+            --pre-hook "if [[ -d /var/run/s6/services/nginx ]]; then s6-svc -d /var/run/s6/services/nginx; fi" \
+            --post-hook "if [[ -d /var/run/s6/services/nginx ]]; then s6-svc -u /var/run/s6/services/nginx; fi" \
             -d $LETSENCRYPT_DOMAIN
         rc=$?
         if [[ $rc -eq 1 ]]; then


jkowens:
I'm running into issues with the letsencrypt cert not being renewed as well. Does the current setup require that the container is restarted periodically for the renewal to happen?

driehuis:
Indeed, restarting the container seems to be required for renewal, because the cronjob doesn't work and gets written to non-persistent storage to boot. Have you tried applying this patch and restarting the container?

jkowens:
Sorry, no I didn't find your patch until after I followed instructions to renew the cert in a closed issue 😞

driehuis:
Well, that's precisely the reason that I do not know myself if my patch covers it completely :-)  Al I know is that it needs fixing anyway.

jkowens:
Agreed!

saghul (approved):

saghul:
Good catch, thanks!

jkowens:
@saghul is there something in the works to get the cron job working for renewal?

saghul:
That should already be working, is it not working?

jkowens:
I see no cron jobs for root in the web container when I run `crontab -l`. It sounds like @driehuis has the same issue. We have to restart the container for the LetsEncrypt cert to renew.

saghul:
Can you share the container output when it boots up for the first time? Are you re-creating the container? I think the problem might be the contrab is not saved to persistent storage and when recreating the container it may get lost.

jkowens:
I'm not sure if this is what you're looking for:

docker logs docker-jitsi-meet_web_1
```
[s6-init] making user provided files available at /var/run/s6/etc...exited 0.
[s6-init] ensuring user provided files have correct perms...exited 0.
[fix-attrs.d] applying ownership & permissions fixes...
[fix-attrs.d] done.
[cont-init.d] executing container initialization scripts...
[cont-init.d] 01-set-timezone: executing... 
[cont-init.d] 01-set-timezone: exited 0.
[cont-init.d] 10-config: executing... 
[cont-init.d] 10-config: exited 0.
[cont-init.d] done.
[services.d] starting services
[services.d] done.
```

Then if I do `docker exec -it docker-jitsi-meet_web_1 /bin/bash` and run

```
# crontab -l
no crontab for root
```

I do use `docker-compose down` and `docker-compose up -d` when I push out updates to docker-jitsi-meet.

saghul:
That's weird, there is no acme related logs there. Did you already create the certa at least once?

jkowens:
Yes, do I need to delete some files on the host system before doing a `docker-compose up -d`?

saghul:
Wipe the entire Jitsi-Meet-cfg.

saghul:
This is just to test btw.

jkowens:
Well that was it. If I delete the `.jitsi-meet-cfg` directory the cron job is added to crontab. It's not ideal, but I can work with that. Thanks for the help @saghul  🙏

saghul:
I think it does make sense to fix this though. Can you open an issue so we don't forget?

jkowens:
Done!

driehuis:
Is the crontab entry still there if you restart the container? IIRC the crontab entry was created in the layer of the running container and got wiped upon restart (but I can't check at the moment).

saghul:
That's my suspicion too.

PR #1087, opened 2021-07-08
added TCP_HARVESTER_SSLTCP control in the env file

mghadam:
Added control for [videobridge.ice.tcp.ssltcp](https://github.com/jitsi/jitsi-videobridge/blob/master/CONFIG.md) variable in the .env file

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -245,6 +245,7 @@ services:
             - JVB_BREWERY_MUC
             - JVB_PORT
             - JVB_TCP_HARVESTER_DISABLED
+            - JVB_TCP_HARVESTER_SSLTCP
             - JVB_TCP_PORT
             - JVB_TCP_MAPPED_PORT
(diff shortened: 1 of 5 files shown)


saghul:
@bgrozev @JonathanLennox does it make sense to have this enabled by default?

damencho:
We recommend using coturn for the TCP connections, we have seen problems with jvb cpu usage when the tcp from jvb is used, that's why that is turned off by default as it is not recommended to be used.

damencho:
> > @bgrozev @JonathanLennox does it make sense to have this enabled by default?
> 
> Yes, it already defaults to `true`. This is the option to pretent to use SSL, not the option to enable/disable the TCP harvester.

Ups :) Sorry.

saghul:
Sorry this fell trough the cracks. We have completely disabled TCP support in the JVB so this no longer applies, sorry.

PR #1101, opened 2021-08-12
Enable A/V Moderation

khakulov:
With enviroment variable ENABLE_AV_MODERATION will be av_moderation enabled

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -125,6 +125,7 @@ services:
             - ENABLE_AUTH
             - ENABLE_GUESTS
             - ENABLE_LOBBY
+            - ENABLE_AV_MODERATION
             - ENABLE_XMPP_WEBSOCKET
             - GLOBAL_MODULES
             - GLOBAL_CONFIG
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -146,6 +146,7 @@ services:
             - XMPP_AUTH_DOMAIN
             - XMPP_GUEST_DOMAIN
             - XMPP_MUC_DOMAIN
+            - XMPP_MUC_DOMAIN_PREFIX
             - XMPP_INTERNAL_MUC_DOMAIN
             - XMPP_MODULES
             - XMPP_MUC_MODULES
--- a/env.example
+++ b/env.example
@@ -52,6 +52,9 @@ TZ=UTC
 # Control whether the lobby feature should be enabled or not
 #ENABLE_LOBBY=1
 
+# Control whether the A/V moderation should be enabled or not
+#ENABLE_AV_MODERATION=1
+
 # Show a prejoin page before entering a conference
 #ENABLE_PREJOIN_PAGE=0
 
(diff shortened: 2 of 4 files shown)


saghul:
I tested this and I'm getting the following error:

```
prosody_1  | avmoderation.meet.jitsi:av_moderation_component              warn	No room found found for /folkshipsslideobviously
```

Looks like the room name has a preceding / which is making this not work perhaps? Are you aware of this bug @damencho ?

khakulov:
https://github.com/jitsi/docker-jitsi-meet/blob/856e414084876551fe29f0c72d54d476d9905bd0/env.example#L221

but `util.lib.lua` uses `conference` by default, so that `muc_mapper_domain` is `conference.jitsi.meet`.
https://github.com/jitsi/jitsi-meet/blob/9914a5d14a9ede95fbbd9674a38401ed24bd1fa1/resources/prosody-plugins/util.lib.lua#L11

i have added new enviroment variable XMPP_MUC_DOMAIN_PREFIX, so that `muc_mapper_domain` should be correct

damencho:
> I tested this and I'm getting the following error:
> 
> ```
> prosody_1  | avmoderation.meet.jitsi:av_moderation_component              warn	No room found found for /folkshipsslideobviously
> ```
> 
> Looks like the room name has a preceding / which is making this not work perhaps? Are you aware of this bug @damencho ?

Nope. It works on alpha ...

saghul:
> https://github.com/jitsi/docker-jitsi-meet/blob/856e414084876551fe29f0c72d54d476d9905bd0/env.example#L221
> 
> 
> 
> but `util.lib.lua` uses `conference` by default, so that `muc_mapper_domain` is `conference.jitsi.meet`.
> 
> https://github.com/jitsi/jitsi-meet/blob/9914a5d14a9ede95fbbd9674a38401ed24bd1fa1/resources/prosody-plugins/util.lib.lua#L11
> 
> 
> 
> i have added new enviroment variable XMPP_MUC_DOMAIN_PREFIX, so that `muc_mapper_domain` should be correct

Good catch!

wernf:
Thanks for this patch.

I am wondering: would we add the av_moderation_component pending on the value of ENABLE_AV_MODERATION? It might not do any harm adding unused configuration, it would just make the config cleaner. But maybe I am missing something.

saghul:
> Thanks for this patch.
> 
> I am wondering: would we add the av_moderation_component pending on the value of ENABLE_AV_MODERATION? It might not do any harm adding unused configuration, it would just make the config cleaner. But maybe I am missing something.

Makes sense. @khakulov can you please guard the component?

khakulov:
> Makes sense. @khakulov can you please guard the component?

done

saghul:
Squashed and landed in c95f0d6, thanks @khakulov ! I ended up scrapping XMPP_MUC_DOMAIN_PREFIX, since we can infer it from XMPP_MUC_DOMAIN, as we do when generating the config file.

PR #1109, opened 2021-08-21
feat: env variables for configuring vp9

prayagsingh:


prayagsingh:
@saghul Please review it.

please don't merge it as of now because when VP9 is enabled then Screenshare is causing a crash if the chrome/chromium version is 92. It's working fine on chromium v93. I have tested it today only.
 OR 
we can add a note about updating the chrome/Brave/Edge browsers to version 93 if using VP9.

web/rootfs/defaults/settings-config.js
```diff
@@ -313,8 +313,6 @@ config.deploymentInfo.userRegion = '{{ $DEPLOYMENTINFO_USERREGION }}';
 
 
 // Testing
-//
```
saghul, line 316: LEave these please, I've used that style of comments to designate a "section"

saghul (approved):
LGTM with a minor nit. Since VP9 is not enabled by default this is safe to land.

web/rootfs/defaults/settings-config.js
prayagsingh: ok

saghul (approved):
LGTM

PR #1118, opened 2021-09-07
add enable-auto-login option to jicofo config

kennylee2008:
Once an authenticated user login, he is always logged in before the session timeout.  There should be an _enable-auto-login_ option in jicofo to control the auto login.  But you can not configure this option in _.env_ file when you self deploy with docker.

This pull request add an variable in _.env_ file, and modify the _docker-compose.xml_ and _jicofo.conf_ files to add this option.

_env.example_ file:
```javascript
# Enable auto login 
#ENABLE_AUTO_LOGIN=1
```
_docker-compose.xml_ file:
```javascript
            - ENABLE_AUTO_LOGIN
```

_jicofo.conf_ file:
```javascript
...
{{ $ENABLE_AUTO_LOGIN := .Env.ENABLE_AUTO_LOGIN | default "1" | toBool }}
...
      {{ if $ENABLE_AUTO_LOGIN }}
      enable-auto-login=true
      {{ else }}
      enable-auto-login=false
      {{ end }}
```

related discussion:
[Unable do disable autologin using Docker Installation](https://community.jitsi.org/t/unable-do-disable-autologin-using-docker-installation/98505)

[Default timeout for authenticated users](https://community.jitsi.org/t/default-timeout-for-authenticated-users/17664)

### DIFF
--- a/env.example
+++ b/env.example
@@ -142,9 +142,6 @@ ETHERPAD_SKIN_VARIANTS=super-light-toolbar super-light-editor light-background f
 # Select authentication type: internal, jwt or ldap
 #AUTH_TYPE=internal
 
-# Enable auto login 
-#ENABLE_AUTO_LOGIN=1
-
 # JWT authentication
 #
 
--- a/env.example
+++ b/env.example
@@ -142,6 +142,9 @@ ETHERPAD_SKIN_VARIANTS=super-light-toolbar super-light-editor light-background f
 # Select authentication type: internal, jwt or ldap
 #AUTH_TYPE=internal
 
+# Enable auto login 
+#ENABLE_AUTO_LOGIN=1
+
 # JWT authentication
 #
 
--- a/jicofo/rootfs/defaults/jicofo.conf
+++ b/jicofo/rootfs/defaults/jicofo.conf
@@ -3,7 +3,7 @@
 {{ $AUTH_TYPE := .Env.AUTH_TYPE | default "internal" }}
 {{ $ENABLE_RECORDING := .Env.ENABLE_RECORDING | default "0" | toBool }}
 {{ $ENABLE_OCTO := .Env.ENABLE_OCTO | default "0" | toBool }}
-{{ $ENABLE_AUTO_LOGIN := .Env.ENABLE_AUTO_LOGIN | default "0" | toBool }}
+{{ $ENABLE_AUTO_LOGIN := .Env.ENABLE_AUTO_LOGIN | default "1" | toBool }}
 
 jicofo {
     {{ if $ENABLE_AUTH }}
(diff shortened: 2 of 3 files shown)


env.example
```diff
@@ -142,6 +142,9 @@ ETHERPAD_SKIN_VARIANTS=super-light-toolbar super-light-editor light-background f
 # Select authentication type: internal, jwt or ldap
 #AUTH_TYPE=internal
 
+# Enable auto login 
+#ENABLE_AUTO_LOGIN=1
```
saghul, line 146: Please don't add it here. Document it on the handbook instead.

jicofo/rootfs/defaults/jicofo.conf
```diff
@@ -3,6 +3,7 @@
 {{ $AUTH_TYPE := .Env.AUTH_TYPE | default "internal" }}
 {{ $ENABLE_RECORDING := .Env.ENABLE_RECORDING | default "0" | toBool }}
 {{ $ENABLE_OCTO := .Env.ENABLE_OCTO | default "0" | toBool }}
+{{ $ENABLE_AUTO_LOGIN := .Env.ENABLE_AUTO_LOGIN | default "1" | toBool }}
```
saghul, line 6: What is the current default? We should preserve that.

jicofo/rootfs/defaults/jicofo.conf
```diff
@@ -22,6 +23,11 @@ jicofo {
       {{ else }}
       login-url = "{{ .Env.XMPP_DOMAIN }}"
       {{ end }}
+      {{ if $ENABLE_AUTO_LOGIN }}
```
saghul, line 26: You can write this as `enable-auto-login = {{ $ENABLE_AUTO_LOGIN }}`

saghul (changes requested):
Left some comments, thanks for adding this!

env.example
kennylee2008: In the source code, the default value is true, and I thought that adding this comment to _env.example_ was a good practice for others to know what the default value is and how it can be configured without having to read the documentation.

jicofo/rootfs/defaults/jicofo.conf
kennylee2008: Got it, Thanks @saghul

jicofo/rootfs/defaults/jicofo.conf
kennylee2008: The current default value is _true_, that's why I define _1_ as default.

env.example
kennylee2008: But of cause, if this pull request is accepted, I will create a new pull request on the handbook as well, to add this new configuration note.

saghul (approved):

PR #1121, opened 2021-09-10
Persist crontabs for letsencrypt

driehuis:
This seems to fix the persistance of the crontab entry for letsencrypt for me.


saghul:
Sweet, thanks!

PR #1134, opened 2021-10-06
docker-swarm files

prayagsingh:
Deploy jitsi using docker-swarm with Traefik.


prayagsingh:
It's easy to get the letsencrypt cert using Traefik for different services(monitoring services like grafana dashboard,telegraf) running on the same machine. I can do the SSL termination at Traefik and then forward the request to each of the services.

We can do it without Traefik also like we are doing it docker-compose setup.

saghul:
If we are to keep this here I'd like to cut out Traefik. Less moving parts.

prayagsingh:
ok. I'll make the changes.

dominiwe:
@saghul is this abandoned? The [docker compose file](https://github.com/jitsi/docker-jitsi-meet/blob/master/docker-compose.yml) is still quite hard to get working with swarm due to the incompatible network name...
Does everything here still work and could we combine it all into one compose file to be used with `docker stack` @prayagsingh?

saghul:
Sorry I dropped the ball. No I'm still keen on seeing if we put swarm configs here or in the contrib org. I just forgot sorry.

prayagsingh:
> Does everything here still work and could we combine it all into one compose file to be used with `docker stack` @prayagsingh?

This PR is a bit outdated now(but it will work) and I don't have enough bandwidth now so it'll be a little difficult for me to update this PR. But I'm happy to help if someone wants to take it up.

saghul:
Thanks for your effort so far @prayagsingh ! I still plan to see if it makes sense to have it here or in a contrib repo.

If anyone has ideas, I'm happy to listen!

prayagsingh:
@saghul  IMO, moving it to contrib repo makes sense since everything that is managed by the third party is present there. `docker-compose` is fine in the main repo and everything else moves to `jitsi-contrib`.

Issue #1135, opened 2021-10-06
Docker Swarm: JVB networking problems


slauth:
As reported in [this community thread](https://community.jitsi.org/t/docker-swarm/98091) I'm experiencing problems when using Jitsi Meet in a Docker Swarm.

Relevant nginx error:

```
jitsi_web.1.ki9oj4h7bary@aws-master    | 2021/04/20 12:22:36 [error] 1654#1654: *8 connect() failed (110: Connection timed out) while connecting to upstream, client: 10.0.0.2, server: _, request: "GET /colibri-ws/172.18.0.6/a1f3e6f6ce668ba5/a36db68d?pwd=5uqb98od1frm6tk6q74875iqg2 HTTP/1.1", upstream: "http://172.18.0.6:9090/colibri-ws/172.18.0.6/a1f3e6f6ce668ba5/a36db68d?pwd=5uqb98od1frm6tk6q74875iqg2", host: "jitsi.example.com"
jitsi_web.1.ki9oj4h7bary@aws-master    | 10.0.0.2 - - [20/Apr/2021:12:22:36 +0200] "GET /colibri-ws/172.18.0.6/a1f3e6f6ce668ba5/a36db68d?pwd=5uqb98od1frm6tk6q74875iqg2 HTTP/1.1" 502 580 "-" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.72 Safari/537.36"
```

prayagsingh:
Can you try this [example](https://github.com/prayagsingh/docker-jitsi-meet/tree/jitsi-docker-swarm/examples/docker-swarm)?

slauth:
@prayagsingh I'd like to stick with the officially supported Docker setup, which already works in swarm mode except for one tiny workaround (in JVB container use `hostname -i` as `LOCAL_ADDRESS`).

@saghul had an idea how to improve the JVB container's network configuration so that it works out-of-the-box in swarm mode also.

wiltonsr:
@slauth How do you setted `DOCKER_HOST_ADDRESS` in swarm with multiples `JVBs`?

PR #1139, opened 2021-10-12
jvb: try to use the correct IP as the default server ID

saghul:
This hopefully fixes WS issues on environment with multiple networks
such as Docker Swarm.

### DIFF
--- a/jvb/Dockerfile
+++ b/jvb/Dockerfile
@@ -3,7 +3,7 @@ ARG BASE_TAG=latest
 FROM ${JITSI_REPO}/base-java:${BASE_TAG}
 
 RUN apt-dpkg-wrap apt-get update && \
-    apt-dpkg-wrap apt-get install -y jitsi-videobridge2 jq curl iproute2 && \
+    apt-dpkg-wrap apt-get install -y jitsi-videobridge2 jq curl iproute2 dnsutils && \
     apt-cleanup
 
 COPY rootfs/ /
--- a/jvb/rootfs/defaults/jvb.conf
+++ b/jvb/rootfs/defaults/jvb.conf
@@ -1,12 +1,12 @@
+{{ $COLIBRI_REST_ENABLED := .Env.COLIBRI_REST_ENABLED | default "false" | toBool }}
 {{ $ENABLE_COLIBRI_WEBSOCKET := .Env.ENABLE_COLIBRI_WEBSOCKET | default "1" | toBool }}
 {{ $ENABLE_OCTO := .Env.ENABLE_OCTO | default "0" | toBool }}
 {{ $JVB_TCP_PORT := .Env.JVB_TCP_PORT | default "4443" }}
 {{ $JVB_TCP_MAPPED_PORT := .Env.JVB_TCP_MAPPED_PORT | default $JVB_TCP_PORT }}
 {{ $PUBLIC_URL_DOMAIN := .Env.PUBLIC_URL | default "https://localhost:8443" | trimPrefix "https://" | trimSuffix "/" -}}
-{{ $WS_DOMAIN := .Env.JVB_WS_DOMAIN | default $PUBLIC_URL_DOMAIN -}}
-{{ $WS_SERVER_ID := .Env.JVB_WS_SERVER_ID | default .Env.LOCAL_ADDRESS -}}
-{{ $COLIBRI_REST_ENABLED := .Env.COLIBRI_REST_ENABLED | default "false" | toBool }}
 {{ $SHUTDOWN_REST_ENABLED := .Env.SHUTDOWN_REST_ENABLED | default "false" | toBool }}
+{{ $WS_DOMAIN := .Env.JVB_WS_DOMAIN | default $PUBLIC_URL_DOMAIN -}}
+{{ $WS_SERVER_ID := .Env.JVB_WS_SERVER_ID | default .Env.JVB_WS_SERVER_ID_FALLBACK -}}
 
 videobridge {
     ice {
@@ -84,7 +84,6 @@ ice4j {
                 enabled = false
 {{ end -}}
             }
-
             static-mappings = [
 {{ if .Env.DOCKER_HOST_ADDRESS -}}
                 {
--- a/jvb/rootfs/etc/cont-init.d/10-config
+++ b/jvb/rootfs/etc/cont-init.d/10-config
@@ -1,8 +1,5 @@
 #!/usr/bin/with-contenv bash
 
-export LOCAL_ADDRESS=$(ip addr show dev "$(ip route|awk '/^default/ { print $5 }')" | grep -oP '(?<=inet\s)\d+(\.\d+){3}')
-export SENTRY_RELEASE="${SENTRY_RELEASE:-$(apt-cache policy jitsi-videobridge2 | sed -n '/Installed/p' | sed -e 's/[^:]*: //')}"
-
 if [[ -z $JVB_AUTH_PASSWORD ]]; then
     echo 'FATAL ERROR: JVB auth password must be set'
     exit 1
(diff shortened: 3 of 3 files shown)


saghul:
@prayagsingh Maybe you can help verify this fix since you've been tinkering with Swarm?

prayagsingh:
I tested it on my local setup(Docker for Windows) and it's not working fine.

**Logs**

```
2021/10/20 07:39:02 [error] 259#259: *14 upstream timed out (110: Connection timed out) while connecting to upstream, client: 10.0.0.2, server: _, request: "GET /colibri-ws/172.18.0.6/812c7596587f397a/7bfd39de?pwd=2m68r221aqmh0kgsjc0hkafvnk HTTP/1.1", upstream: "http://172.18.0.6:9090/colibri-ws/172.18.0.6/812c7596587f397a/7bfd39de?pwd=2m68r221aqmh0kgsjc0hkafvnk", host: "localhost:8443"
10.0.0.2 - - [20/Oct/2021:07:39:02 +0000] "GET /colibri-ws/172.18.0.6/812c7596587f397a/7bfd39de?pwd=2m68r221aqmh0kgsjc0hkafvnk HTTP/1.1" 504 590 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/95.0.4638.54 Safari/537.36"
```

**jvb.conf**

```
videobridge {
    ice {
        udp {
            port = 10000
        }
        tcp {
            enabled = false
            port = 4443

        }
    }
    apis {
        xmpp-client {
            configs {
                shard {
                    HOSTNAME = "xmpp.meet.jitsi"
                    DOMAIN = "auth.meet.jitsi"
                    USERNAME = "jvb"
                    PASSWORD = "2400fac916d1db0c9c47937b7cfb6eb8"
                    MUC_JIDS = "jvbbrewery@internal-muc.meet.jitsi"
                    MUC_NICKNAME = "ed34a130dd29"
                    DISABLE_CERTIFICATE_VERIFICATION = true
                }
            }
        }
    }
    stats {
        enabled = true
    }
    websockets {
        enabled = true
        domain = "localhost:8443"
        tls = true
        server-id = "172.18.0.6"
    }
    http-servers {
        private {
          host = 0.0.0.0
        }
        public {
            host = 0.0.0.0
            port = 9090
        }
    }

    }

ice4j {
    harvest {
        mapping {
            stun {

                addresses = [ "meet-jit-si-turnrelay.jitsi.net:443" ]

            }
        }
    }
}
```

**meet.conf**

```
server_name _;

client_max_body_size 0;

root /usr/share/jitsi-meet;

# ssi on with javascript for multidomain variables in config.js
ssi on;
ssi_types application/x-javascript application/javascript;

index index.html index.htm;
error_page 404 /static/404.html;

# Security headers
add_header X-Content-Type-Options nosniff;
add_header X-XSS-Protection "1; mode=block";




add_header Permissions-Policy "interest-cohort=()";


location = /config.js {
    alias /config/config.js;
}

location = /interface_config.js {
    alias /config/interface_config.js;
}

location = /external_api.js {
    alias /usr/share/jitsi-meet/libs/external_api.min.js;
}

# ensure all static content can always be found first
location ~ ^/(libs|css|static|images|fonts|lang|sounds|connection_optimization|.well-known)/(.*)$
{
    add_header 'Access-Control-Allow-Origin' '*';
    alias /usr/share/jitsi-meet/$1/$2;
}


# colibri (JVB) websockets
location ~ ^/colibri-ws/([a-zA-Z0-9-\.]+)/(.*) {
    proxy_pass http://$1:9090/colibri-ws/$1/$2$is_args$args;
    proxy_http_version 1.1;
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection "upgrade";
    tcp_nodelay on;
}


# BOSH
location = /http-bind {
    proxy_pass http://xmpp.meet.jitsi:5280/http-bind;
    proxy_set_header Host meet.jitsi;
}


# xmpp websockets
location = /xmpp-websocket {
    proxy_pass http://xmpp.meet.jitsi:5280/xmpp-websocket;
    proxy_http_version 1.1;

    proxy_set_header Connection "upgrade";
    proxy_set_header Upgrade $http_upgrade;

    proxy_set_header Host meet.jitsi;
    tcp_nodelay on;
}


location ~ ^/([^/?&:'"]+)$ {
    try_files $uri @root_path;
}

location @root_path {
    rewrite ^/(.*)$ / break;
}
```

sapkra:
@saghul It broke kubernetes deployments but there is a better solution for these deployments so we will apply it to the helm chart. Just to let you know.

saghul:
Damn, that's unfortunate, what is the better solution for k8s?

sapkra:
To directly reference the podIP as the `JVB_WS_SERVER_ID` env var in the kubernetes manifests.

```yaml
- name: JVB_WS_SERVER_ID
  valueFrom:
    fieldRef:
      fieldPath: status.podIP
```

saghul:
Makes sense. I assume you have a single JVB per pod then?

DanielMcAssey:
Can confirm, I amu using this in the same way with K8S using a stateful set for auto scaling JVBs

spijet:
I was tinkering around with possible solutions for `SERVER_ID` on Kubernetes, and found two that seem to be the most "proper" ways to set it:
* We can use a K8s service to use it's own auto-discovery mechanism (this may not work properly when there are multiple JVBs, maybe it'll work with "headless" services better);
* OR we can set the ID automatically using JVB Pod's metadata (just like in example posted by @sapkra).

You can check my PR to Jitsi Meet Helm Chart here: jitsi-contrib/jitsi-helm#28.

sapkra:
> I assume you have a single JVB per pod then?

Yes, to scale JVBs in Kubernetes you actually need multiple deployments or stateful sets because each JVB needs a different config (env vars in this case). Like I said before, the only way to achieve this properly in kubernetes is an operator but [the community is already working on one](https://github.com/jitsi-contrib/jitsi-kubernetes-operator). Anyways, that's a different topic.

nathanielhourt:
Hello, world!

Sadly, I must report that, in my environment, this isn't working. I found this patch as a result of following the trail from [the forum thread](https://community.jitsi.org/t/docker-swarm/98091), through #1135, and on to here, after finding that my deployment of Jitsi Meet in Docker Swarm wasn't passing any audio or video between participants, and on further inspection, it turned out that JVB was taking the `docker_gwbridge` private IP as mapping to its public IP, which is what, if I understand correctly, this patch is attempting to correct.

This patch being long since included in the release I'm using, I concluded that something has changed in the intermediate months and the problem is back.

To get my deployment working, I had to disable STUN (`JVB_DISABLE_STUN=1` and set public IPs in `JVB_ADVERTISE_IPS`) and then I can see that this problem is affecting me by observing near the top of JVB's output, this line:
`JVB 2023-01-11 10:36:49.335 INFO: [12] org.ice4j.ice.harvest.MappingCandidateHarvesters.initialize: Adding a static mapping: StaticMapping(localAddress=172.18.0.9, publicAddress=<addr from JVB_ADVERTISE_IPS>, localPort=null, publicPort=null, name=ip-0)`
That `localAddress=172.18.0.9` is the problem: that's the `docker_gwbridge` IP, which is indeed the route by which outgoing traffic goes to the internet... but not the route by which incoming traffic comes in, which is what we seem to need here. And the `jitsi.meet` network IP isn't either -- in Docker Swarm, traffic usually comes in by the ingress network, but it might come from another container over some other network (i.e., a load balancer or reverse proxy).

Well, I couldn't come up with a reliable way to determine where ingress traffic will come from, so I settled on making it an environment variable.

I revised `/etc/cont-init.d/10-config` wrapping the `LOCAL_ADDRESS` line like this:
```bash
# Local IP for the ice4j mapping harvester.
if [[ -z $JVB_INGRESS_SUBNET ]]; then
    export LOCAL_ADDRESS=$(ip route get 1 | grep -oP '(?<=src ).*' | awk '{ print $1 '})
else
    export LOCAL_ADDRESS=$(ip route get $JVB_INGRESS_SUBNET | grep -oP '(?<=src ).*' | awk '{ print $1 '})
fi
```

Then I added `JVB_INGRESS_SUBNET=10.0.0.0` to my JVB container environment. That did it for me. :-)
> Fwiw, check ingress subnet with `docker network inspect ingress` -- but put whatever subnet here that your traffic will come through.

I'm not sure whether a better solution can be found, but that's what I'm doing. HTH

saghul:
How come the inbound and outbound routes for UDP media are different?

Issue #1141, opened 2021-10-13
Property videobridge.rest.shutdown.enabled is always disabled


cinques:
https://github.com/jitsi/docker-jitsi-meet/blob/master/jvb/rootfs/defaults/jvb.conf
https://github.com/jitsi/jitsi-videobridge/blob/master/jvb/src/main/resources/reference.conf#L96

cinques:
done

PR #1142, opened 2021-10-13
allow to configure shutdown api

cinques:
#1141


cinques:
Ok
Added

xoxys:
Side note: This was a breaking change, as it looks like the default config on older releases has the colobri API enabled by default. We had to enable it again explicitly now to get our monitoring working again, while if it was working without a special config before. 

It is also not the best way to mix two features (Shutdown API/Colibir API) into one PR without any hint in the PR description. Makes it hard do search :)

f-asa:
@xoxys how did you fixed it ? you re-enable the --apis=${JVB_ENABLE_APIS:="none"} options for the api ?

f-asa:
and what is the Shutdown API ?

xoxys:
No, there are two new Env vars now to control the API's and `JVB_ENABLE_APIS` was removed,see https://github.com/jitsi/docker-jitsi-meet/blob/master/env.example#L406-L408

PR #1168, opened 2021-12-14
jibri: correct chromedriver mismatch

benbz:
Previously when `CHROME_RELEASE` is set to something specific (not `latest`), the `[ "${CHROMEDRIVER_MAJOR_RELEASE}" = "latest" ]` check is skipped due to the `||` before it. i.e. the original code is structured like:

```sh
CHROME_RELEASE=95.0.4638.69 CHROMEDRIVER_MAJOR_RELEASE=95 echo "Do initial things" && \
  [ "${CHROME_RELEASE}" = "latest" ] && \
  echo "Do latest chrome release things 1" && \
  echo "Do latest chrome release things 2"  || \
  [ "${CHROME_RELEASE}" != "latest" ] && \
  echo "Do specific chrome release 1" && \
  echo "Do specific chrome release 2" || \
  [ "${CHROMEDRIVER_MAJOR_RELEASE}" = "latest" ] && \
  CHROMEDRIVER_RELEASE="$(curl -4Ls https://chromedriver.storage.googleapis.com/LATEST_RELEASE)" || \
  CHROMEDRIVER_RELEASE="$(curl -4Ls https://chromedriver.storage.googleapis.com/LATEST_RELEASE_${CHROMEDRIVER_MAJOR_RELEASE})" && \
  echo "${CHROMEDRIVER_MAJOR_RELEASE} vs ${CHROMEDRIVER_RELEASE}"

Do initial things
Do specific chrome release 1
Do specific chrome release 2
95 vs 96.0.4664.45
```

The new code on the other hand:
```sh
CHROME_RELEASE=95.0.4638.69 CHROMEDRIVER_MAJOR_RELEASE=95 echo "Do initial things" && \
  [ "${CHROME_RELEASE}" = "latest" ] && \
  ( \
    echo "Do latest chrome release things 1" && \
    echo "Do latest chrome release things 2" \
  ) || \
  ( \
    echo "Do specific chrome release 1" && \
    echo "Do specific chrome release 2" \
  ) && \
  [ "${CHROMEDRIVER_MAJOR_RELEASE}" = "latest" ] && \
  CHROMEDRIVER_RELEASE="$(curl -4Ls https://chromedriver.storage.googleapis.com/LATEST_RELEASE)" || \
  CHROMEDRIVER_RELEASE="$(curl -4Ls https://chromedriver.storage.googleapis.com/LATEST_RELEASE_${CHROMEDRIVER_MAJOR_RELEASE})" && \
  echo "${CHROMEDRIVER_MAJOR_RELEASE} vs ${CHROMEDRIVER_RELEASE}"

Do initial things
Do specific chrome release 1
Do specific chrome release 2
95 vs 95.0.4638.69
```

This PR fixes this, makes things easier to debug in future and adds a fail-safe incase there's another issue

### DIFF
--- a/jibri/Dockerfile
+++ b/jibri/Dockerfile
@@ -13,16 +13,19 @@ RUN apt-dpkg-wrap apt-get update && \
     apt-dpkg-wrap apt-get install -y jibri libgl1-mesa-dri procps && \
     apt-cleanup && \
     [ "${CHROME_RELEASE}" = "latest" ] && \
-    wget -qO - https://dl-ssl.google.com/linux/linux_signing_key.pub | gpg --dearmour /etc/apt/trusted.gpg.d/google.gpg && \
-    echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" > /etc/apt/sources.list.d/google-chrome.list && \
-    apt-dpkg-wrap apt-get update && \
-    apt-dpkg-wrap apt-get install -y google-chrome-stable && \
-    apt-cleanup || \
-    [ "${CHROME_RELEASE}" != "latest" ] && \
-    curl -4so "/tmp/google-chrome-stable_${CHROME_RELEASE}-1_amd64.deb" "http://dl.google.com/linux/chrome/deb/pool/main/g/google-chrome-stable/google-chrome-stable_${CHROME_RELEASE}-1_amd64.deb" && \
-    apt-dpkg-wrap apt-get update && \
-    apt-dpkg-wrap apt-get install -y "/tmp/google-chrome-stable_${CHROME_RELEASE}-1_amd64.deb" && \
-    apt-cleanup || \
+    ( \
+        wget -qO - https://dl-ssl.google.com/linux/linux_signing_key.pub | gpg --dearmour /etc/apt/trusted.gpg.d/google.gpg && \
+        echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" > /etc/apt/sources.list.d/google-chrome.list && \
+        apt-dpkg-wrap apt-get update && \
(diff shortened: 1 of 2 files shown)


saghul:
Thanks! This image is already huge, so I think having an extra layer where we run a script (let's say some install-chrome.sh we copy) would be best for maintainability.

benbz:
> Thanks! This image is already huge, so I think having an extra layer where we run a script (let's say some install-chrome.sh we copy) would be best for maintainability.

Ack, sounds good to me. I'll try to update this PR with that this afternoon

saghul:
Awesome, thanks a ton!

benbz:
> > Thanks! This image is already huge, so I think having an extra layer where we run a script (let's say some install-chrome.sh we copy) would be best for maintainability.
> 
> Ack, sounds good to me. I'll try to update this PR with that this afternoon

Done in 4afeb5a:
* `set -e` in the script so we don't need to chain everything with `&&`
* Chrome & Chromedriver are installed in a separate layer so that changing their versions doesn't invalid the main apt layer. I'm happy to merge it all back down to a single layer if preferred
* I moved the installation of `jitsi-upload-integrations` & `jq` into the main `apt` run rather than a 3rd apt run at the end. I couldn't see a reason why they needed to be installed after Chrome but again happy to change this back if there's something I've missed
* Fixed an issue where the gnome GPG key wasn't being saved correctly when `CHROME_RELEASE=latest`
* Removed the check I added in 62a7154, it didn't work when `CHROMEDRIVER_MAJOR_RELEASE=latest` and then felt a bit pointless

saghul (approved):

lephat08:
Hi, I have issues with Jibri. I deployed Jibri on docker. However, when I started recording, I got problems. Could you help me how to resolve it? :(
image

PR #1171, opened 2021-12-14
fix(jigasi): participants joining conference without going through lobby

wfleischer:
This patch will make Jigasi using the guest domain for sip participants, if enabled. This fixes Jigasi users joining a conference without approval when lobby is enabled, due to the whitelisting of the auth domain in the Prosody lobby module.


saghul (approved):
LGTM, thanks Werner! @damencho can you PTAL too?

damencho (approved):

PR #1175, opened 2021-12-15
nginx: jvb websocket proxy avoid arbitrary redirect

aaronkvanmeerten:
This is intended to ensure the colibri-ws passthrough for the JVB websocket only goes to the proper location

### DIFF
--- a/web/rootfs/defaults/meet.conf
+++ b/web/rootfs/defaults/meet.conf
@@ -1,5 +1,6 @@
 {{ $ENABLE_COLIBRI_WEBSOCKET := .Env.ENABLE_COLIBRI_WEBSOCKET | default "1" | toBool }}
 {{ $ENABLE_XMPP_WEBSOCKET := .Env.ENABLE_XMPP_WEBSOCKET | default "1" | toBool }}
+{{ $WS_SERVER_ID := .Env.JVB_WS_SERVER_ID | default .Env.JVB_WS_SERVER_ID_FALLBACK -}}
 
 server_name _;
 
@@ -48,7 +49,7 @@ location ~ ^/(libs|css|static|images|fonts|lang|sounds|connection_optimization|.
 {{ if $ENABLE_COLIBRI_WEBSOCKET }}
 # colibri (JVB) websockets
 location ~ ^/colibri-ws/([a-zA-Z0-9-\.]+)/(.*) {
-    proxy_pass http://$1:9090/colibri-ws/$1/$2$is_args$args;
+    proxy_pass http://{{ $WS_SERVER_ID }}:9090/colibri-ws/$1/$2$is_args$args;
     proxy_http_version 1.1;
     proxy_set_header Upgrade $http_upgrade;
     proxy_set_header Connection "upgrade";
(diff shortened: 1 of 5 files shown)


bgrozev (approved):

bgrozev (approved):

DanielMcAssey:
Won't this break scalable infrastructure like K8s?

aaronkvanmeerten:
> Won't this break scalable infrastructure like K8s?
> 
> For example using multiple JVBs with one WS proxy won't work with these changes

This is true, I was considering only the docker-compose case.  It may be that we need to have a few flags to control this behavior, since by default the regexp will forward to any IP, and any path, on port 9090.  So likely if that's the desired behavior, we should have that behind a flag, and add something more restrictive for the default case.  I'll work on something to that effect.

web/rootfs/defaults/meet.conf
```diff
@@ -47,8 +48,8 @@ location ~ ^/(libs|css|static|images|fonts|lang|sounds|connection_optimization|.
 
 {{ if $ENABLE_COLIBRI_WEBSOCKET }}
 # colibri (JVB) websockets
-location ~ ^/colibri-ws/([a-zA-Z0-9-\.]+)/(.*) {
-    proxy_pass http://$1:9090/colibri-ws/$1/$2$is_args$args;
```
horazont, line 51: This is a problem if the server ID contains dots, as those will be interpreted as regex syntax.

saghul (commented):
This curtails the default installation to a single JVB doesn't it? The original approach tried to not impose that limitation so those that want to run more than one bridge with the default setup can, without resorting to building their own images.

prayagsingh:
> This curtails the default installation to a single JVB doesn't it?

Yes, it will restrict to a single JVB only.

horazont:
Couldn't one just allow a regex for the WS_SERVER_ID and carry the matched text over into the proxy_pass? I.e. just make what previously was `([a-zA-Z0-9-\.]+)` configurable.

saghul:
An option could be to use proxy_bind (https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_bind) and set it to the IP address used to route packets to the JVB, thus limiting its scope to the internal meet.jitsi network.

DanielMcAssey:
That would break Kubernetes deployments right?

aaronkvanmeerten:
> Couldn't one just allow a regex for the WS_SERVER_ID and carry the matched text over into the proxy_pass? I.e. just make what previously was `([a-zA-Z0-9-\.]+)` configurable.

It seems to me like this is the best path towards making this both configurable and secure-by-default.  The current setting is too open for a default, while also being a valid configuration to choose when desiring to proxy to arbitrary JVBs.  I believe we'll want to document the current value and suggest it to those folks deploying into larger environments (with caveats about its dangers too), while defaulting to a more locked down setup either via @saghul 's suggestion above of proxy_bind, or somthing based on my PR above.

Alternately, we could provide the above in the defaults as a new variable like WS_SERVER_REGEXP and and add an additional WS_SERVER_FROM_REGEXP boolean which defaults to false but can be simply set to true to re-enable the present behavior.

aaronkvanmeerten:
closing in favor of https://github.com/jitsi/docker-jitsi-meet/pull/1645

PR #1184, opened 2021-12-29
Add +search to resolve XMPP server aliases

dsmeytis:
Hello! I found out that when docker images are used within a k8s cluster `dig` util without `+search` option will not lookup to `search` section of the `/etc/resolv.conf` and thus it can't resolve XMPP server ip address.
Thanks for your attention and I hope it would be helpful.


saghul (approved):

saghul:
Nice catch!

PR #1186, opened 2022-01-04
allow to configure buttons in toolbar and pre-join screen

prayagsingh:
This PR will allow configuring the buttons in the toolbar and at the pre-join screen. For example, If we want to hide the `invite` button at the pre-join screen then we can set  `HIDE_PREMEETING_BUTTONS=invite`. Similarly setting `TOOLBAR_BUTTONS=camera,chat,fullscreen` will only show these buttons.

### DIFF
--- a/web/rootfs/defaults/settings-config.js
+++ b/web/rootfs/defaults/settings-config.js
@@ -364,3 +364,13 @@ config.disableReactions = {{ $DISABLE_REACTIONS }};
 
 // Polls
 config.disablePolls = {{ $DISABLE_POLLS }};
+
+// configure toolbar buttons
+{{ if .Env.TOOLBAR_BUTTONS -}}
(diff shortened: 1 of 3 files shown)


web/rootfs/defaults/settings-config.js
```diff
@@ -364,3 +364,13 @@ config.disableReactions = {{ $DISABLE_REACTIONS }};
 
 // Polls
 config.disablePolls = {{ $DISABLE_POLLS }};
+
+// configure toolbar buttons
```
saghul, line 368: Nit: can you properly capitalize `configure`?

saghul (approved):
LGTM with a nit.

web/rootfs/defaults/settings-config.js
prayagsingh: Sure

saghul (approved):

PR #1201, opened 2022-01-18
feat(web): cache versioned static files

wfleischer:


saghul (approved):

saghul:
@damencho does this LGTY too?

wfleischer:
I designed this patch to match this example: https://github.com/jitsi/jitsi-meet/blob/master/doc/debian/jitsi-meet/jitsi-meet.example#L72

However analysing meet.jit.si, it seems like you are caching all static files. For example, olm.wasm is cached, although not being versioned. If this would be recommended, then I would suggest to rather merge pull request https://github.com/jitsi/docker-jitsi-meet/pull/1043 from May 9, 2021, which I just got aware of and enables caching regardless of the version parameter:

saghul:
I see. Let's start with this one though. meet.jit.si works different because those assets come from CDN URLs that change when we make a release, so they are versioned in a way.

PR #1202, opened 2022-01-18
env.example : ETHERPAD_PUBLIC_URL : incl. /p/ path

### LINKED ISSUE #1200: ETHERPAD_PUBLIC_URL needs to contain /p/ path fragment
The `env.example` contains a key for an external Etherpad instance, that can be served through its `ETHERPAD_PUBLIC_URL`. Since the generated address will only append a current room's name to that base URL, it is necessary here to also add the `/p/` pad path fragment from Etherpad.

This way pads can be embedded and displayed.

Example:

```shell
ETHERPAD_PUBLIC_URL=https://pad.example.com/p/
```

The information about the necessity of the presence of the `/p/` fragment could be included in the `#` comment above the key in `env.example`.

almereyda:
Closes #1200

### DIFF
--- a/env.example
+++ b/env.example
@@ -97,8 +97,8 @@ TZ=UTC
 # Set etherpad-lite URL in docker local network (uncomment to enable)
 #ETHERPAD_URL_BASE=http://etherpad.meet.jitsi:9001
 
-# Set etherpad-lite public URL (uncomment to enable)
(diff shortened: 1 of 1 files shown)


saghul (approved):

saghul:
Cheers!

almereyda:
Same!

Issue #1216, opened 2022-02-04
Add labels to docker Dockerfiles


xoxys:
Hi, I would like to add (a subset) of [opencontainers](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys) labels to the Dockerfiles. A good starter would be: 

```
LABEL org.opencontainers.image.title=
LABEL org.opencontainers.image.url=
LABEL org.opencontainers.image.source=
LABEL org.opencontainers.image.documentation=
```

We use [renovate bot](https://docs.renovatebot.com/) to auto-create PRs for our infrastructure deployments, we would like to see the changelog in PRs created by renovate bot. This can be achieved by using the `org.opencontainers.image.source` label to point to the GitHub source directory, https://github.com/renovatebot/renovate/pull/2907.

saghul:
Sounds reasonable, let's have them!

xoxys:
Cool! I can provide a PR if you like.

saghul:
Please do!

rcarmo:
Hi again! 

In case my comment on the pull request is missed, the current labels are broken because they lack quotation marks for strings with spaces, and I had to remove them with 

```
sed -i "/^LABEL/d" $(find . -print |grep Dockerfile)
```

xoxys:
Thanks for the info! See https://github.com/jitsi/docker-jitsi-meet/pull/1220

prayagsingh:
Is anyone tried building an image on WSL2 using docker-desktop? I'm facing some issues. Below is the snippet.

```
docker image build -t prayagsingh/web:latest .
[+] Building 0.1s (2/2) FINISHED
 => [internal] load build definition from Dockerfile                                                                                                                                                       0.0s
 => => transferring dockerfile: 1.24kB                                                                                                                                                                     0.0s
 => [internal] load .dockerignore                                                                                                                                                                          0.0s
 => => transferring context: 2B                                                                                                                                                                            0.0s
failed to solve with frontend dockerfile.v0: failed to create LLB definition: no build stage in current context
``` 
There are a few suggestions available on StackOverflow about disabling the buildkit by setting `DOCKER_BUILDKIT=0` just before the `docker image build` command and it didn't work either.  Since I'm not familiar with the OCI standards hence asking this question. Is it mandatory to put the `LABEL` before the `FROM ${JITSI_REPO}/base:${BASE_TAG}` command?

xoxys:
[...]

PR #1219, opened 2022-02-05
chore: add opencontainers labels to Dockerfiles

xoxys:
Fixes: https://github.com/jitsi/docker-jitsi-meet/issues/1216

### DIFF
--- a/jibri/Dockerfile
+++ b/jibri/Dockerfile
@@ -1,3 +1,9 @@
+LABEL org.opencontainers.image.title=Jitsi Broadcasting Infrastructure (jibri)
+LABEL org.opencontainers.image.description=Components for recording and/or streaming a conference.
+LABEL org.opencontainers.image.url=https://github.com/jitsi/jibri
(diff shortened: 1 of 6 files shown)


saghul (approved):

saghul:
Cheers!

rcarmo:
Hi there! I'd just like to let you know that the labels broke the Dockerfiles for Ubuntu 20.04 LTS when using Ubuntu's pre-packaged Docker, since they have spaces in them (I'm pretty sure you need quotation marks, although all of my LABELs are generated automatically, so I have only come across this when using the latest commit). I had to run:

```
sed -I "/^LABEL/d" $(find . -print |grep Dockerfile)
```

...to remove them in place, which was easiest when doing my own build.

xoxys:
Hi @rcarmo good point! I never run into an issue while using labels without quotes, so it looks like they have "fixed" it on later docker versions... Nevertheless, should be easy to fix, sorry for the issue.

xoxys:
I've created a PR to add quotes around the label values.

saghul:
Thank you both!

PR #1240, opened 2022-03-15
feat: add ENABLE_LOAD_TEST option

hrenard:
This is useful for load testing with [jitsi-meet-torture's malleus script](https://github.com/jitsi/jitsi-meet-torture/blob/master/scripts/malleus.sh) using the `--use-load-test` flag.

Maybe the var name needs to be more explicit.

### DIFF
--- a/env.example
+++ b/env.example
@@ -423,3 +423,6 @@ RESTART_POLICY=unless-stopped
 
 # Hide the buttons at pre-join screen. Add the buttons name separated with comma
 #HIDE_PREMEETING_BUTTONS=
+
+# Enable the load test minimal ui used for jitsi-meet-torture's malleus.sh script.
+#ENABLE_LOAD_TEST=1
--- a/env.example
+++ b/env.example
@@ -423,6 +423,3 @@ RESTART_POLICY=unless-stopped
 
 # Hide the buttons at pre-join screen. Add the buttons name separated with comma
 #HIDE_PREMEETING_BUTTONS=
-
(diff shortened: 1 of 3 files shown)


env.example
```diff
@@ -423,3 +423,6 @@ RESTART_POLICY=unless-stopped
 
 # Hide the buttons at pre-join screen. Add the buttons name separated with comma
 #HIDE_PREMEETING_BUTTONS=
+
+# Enable the load test minimal ui used for jitsi-meet-torture's malleus.sh script.
+#ENABLE_LOAD_TEST=1
```
saghul, line 428: This is too advanced, please drop it from here.

web/rootfs/defaults/meet.conf
```diff
@@ -84,6 +84,17 @@ location = /xmpp-websocket {
 }
 {{ end }}
 
+{{ if $ENABLE_LOAD_TEST }}
```
saghul, line 87: add a | default "0" | toBool or it will fail if you don't specify it.

saghul (changes requested):

web/rootfs/defaults/meet.conf
hrenard: Sorry for the bad code...

saghul:
We have split out the load-test web app from the project: https://github.com/jitsi/jitsi-meet-load-test

I have no plaans to integrate it in the docker setup, sorry. It can be run from outside of the Docker setup.

PR #1244, opened 2022-03-16
prosody: support 0.12 again

aaronkvanmeerten:
Do not merge yet, only safe once the latest jitsi-meet #5955 hits stable


saghul (approved):

saghul:
Closing since upstream has incorporated the patch and will be part of 0.12.1.

PR #1255, opened 2022-03-28
prosody: update to 0.12

saghul:


saghul:
What's missing at the moment is the alternative to CORS handling, right now Prosody rejects all connection attempts...

prosody/Dockerfile
```diff
@@ -19,7 +19,7 @@ RUN apt-dpkg-wrap apt-get update && \
     cd /tmp/luarocks && ./configure && make && make install && cd - && \
     luarocks install cyrussasl 1.1.0-1 && \
     luarocks install net-url 0.9-1 && \
-    luarocks install luajwtjitsi 2.0-0
+    luarocks install luajwtjitsi 3.0-0
```
damencho, line 22: Have you tested that? There is a jitsi-meet PR that is not merged yet ...

prosody/rootfs/defaults/prosody.cfg.lua
```diff
@@ -173,7 +171,8 @@ network_settings = {
   tcp_backlog = 511;
 }
 
-component_interface = { "*" }
+http_ports = { 5280 }
+http_interfaces = { "*", "::" }
```
damencho, line 175: Do we need this for jitsi-meet default deb installation?

saghul: No because we proxy to localhost.

prosody/Dockerfile
saghul: No, but we cannot build the previous release with lua 5.4 due to a dependency problem, so the only way is forward!

DanielMcAssey:
Does this break CORS requests then? Noticed the removal of the CORS vars

saghul:
No it doesn't because Prosody 0.12.X (not .0) no longer does that by default AFAICT. We may want to add CORS at the web frontend level perhaps.

PR #1176, opened 2021-12-16
added lua-event to Prosody Dockerfile

Morpheus235:
added lua-event (libevent) to Prosody Build to enable the use of the epoll backend

PR #1257, opened 2022-03-29
✨(prosody) enable libevent

lunika:
In the prosody [documentation] it is recommanded to enable libevent in
order to have better performance. For this, the lua-event debian package
should be installed and then we have to enable it in the configuration,
uncommenting the use_libevent instruction.

[documentation]: https://prosody.im/doc/libevent

### DIFF
--- a/prosody/Dockerfile
+++ b/prosody/Dockerfile
@@ -11,6 +11,7 @@ RUN apt-dpkg-wrap apt-get update && \
       libsasl2-dev \
       libssl-dev \
       libreadline-dev \
+      lua-event \
       git \
       unzip \
       wget && \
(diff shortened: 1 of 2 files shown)


DanielMcAssey:
I believe it already uses the native epoll event, as mentioned here: https://github.com/jitsi/docker-jitsi-meet/pull/1176

saghul:
We use the native epoll backend, which is better: https://github.com/jitsi/docker-jitsi-meet/blob/6f6fe7717ce9eca34de4d5d86bc3da2038e6d5ee/prosody/rootfs/defaults/prosody.cfg.lua#L170

libevent is not adding anything at this point.

lunika:
Thanks for your feedback, I didn't notice before that the `network_backend` was used in the configuration file :+1:

PR #1267, opened 2022-04-20
Update s6 overlay and add secrets support

jlisher:
Firstly, Sorry for the large commit. 

### Changes:
- Updated the s6 overlay to the latest format
- Increased generated password complexity
- Added support for using secrets
- Added GPG fingerprint verification
- Added `.editorconfig` to help standardise the file formatting

### Untested images:
I have not verified that these images are perfectly functional, however, I haven't made any changes outside of updating the s6 service scripts.
- jibri
- jigasi

### Notes:
The timeouts may need to be adjusted depending on the systems used. 
These can be adjusted easily using the `{image}/rootfs/etc/s6-overlay/s6-rc.d/{service}/timeout-up` file, which is simply the timeout limit in `ms`


jlisher:
@saghul will do :)

saghul:
Thank you!

Issue #1214, opened 2022-02-02
arm64 image support


saghul:
I'll pin this issue so people can find it easily and track it.

luongdnt:
any news on this issue? @saghul

saghul:
No.

luongdnt:
do we have any solution for by pass this issue on M1 devices? @saghul

saghul:
Please stop asking. You are not going make it happen faster, on the contrary, you'll make developers stressed and not keen on working on what you really want.

wfleischer:
Thank you @saghul for working on this feature! If there is any specific task I could help with, please let me know.

saghul:
> Thank you @saghul for working on this feature! If there is any specific task I could help with, please let me know.

I have a mostly ready branch, the only thing that I'm not a fan of is that buildx doesn't allow building multi arch images and then loading them locally instead of directly pushing them, but I think that's unavoidable at the moment without complicating things too much...

wfleischer:
> > Thank you @saghul for working on this feature! If there is any specific task I could help with, please let me know.
> 
> I have a mostly ready branch, the only thing that I'm not a fan of is that buildx doesn't allow building multi arch images and then loading them locally instead of directly pushing them, but I think that's unavoidable at the moment without complicating things too much...

Just got the Jitsi backend running on my dev M1 mac. I only had to make few adjustments. You have already done a tremendous work. Thank you so much Saúl!

For anyone interested: use the latest release which already includes several patches to support arm64.

saghul:
Yep I was thinking about skipping those for the first release indeed.

nikwise29:
> > > Thank you @saghul for working on this feature! If there is any specific task I could help with, please let me know.
> > 
> > 
> > I have a mostly ready branch, the only thing that I'm not a fan of is that buildx doesn't allow building multi arch images and then loading them locally instead of directly pushing them, but I think that's unavoidable at the moment without complicating things too much...
> 
> Just got the Jitsi backend running on my dev M1 mac. I only had to make few adjustments. You have already done a tremendous work. Thank you so much Saúl!
> 
> For anyone interested: use the latest release which already includes several patches to support arm64.

@wfleischer would you mind highlighting those few adjustments so that one can reproduce it on their M1s.  Thanks!

wfleischer:
[...]

PR #1269, opened 2022-04-21
Update GH action to also build for arm64

### LINKED ISSUE #1069: ARM64 Support for docker images
There are already many Jitsi related packages available for all arches, including arm64: https://download.jitsi.org/stable/

The whole docker stuff however still is only available for amd64, although there are already pretty powerful (and mostly cheaper) ARM64 Servers available on some cloud providers. Not to mention it would then also work on Raspberries (however don't know if they would be fast enough).

Can we somehow make the whole docker stuff also run on ARM64? There already was working version available on this fork, but since maintaining it alone without the community it's lacking behind a lot in terms of updates: https://github.com/amaumene/docker-jitsi-meet/

Firefishy:
Add arm64 build support. 
Add arm64 download logic in `Dockerfile`s

Closes #1069 #1214

### DIFF
--- a/.github/workflows/unstable.yml
+++ b/.github/workflows/unstable.yml
@@ -47,6 +47,8 @@ jobs:
           build-args: |
             JITSI_RELEASE=unstable
           platforms: linux/amd64,linux/arm64
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
       - name: Dryrun
         uses: docker/build-push-action@v2
         if: ${{ github.event_name == 'pull_request' }}
@@ -58,6 +60,8 @@ jobs:
           build-args: |
             JITSI_RELEASE=unstable
           platforms: linux/amd64,linux/arm64
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
 
   base-java:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


Firefishy:
`jibri` image uses chrome + chromedriver which do not appear to be available for `arm64` on linux. :-(

adam-at-mobomo:
> `jibri` image uses chrome + chromedriver which do not appear to be available for `arm64` on linux. :-(

I believe chromium is available along with chromedriver support:

* https://vitux.com/debian-chromium-browser/
* https://packages.debian.org/bullseye/chromium-driver
* https://askubuntu.com/questions/1383789/install-chrome-on-ubuntu-debian-with-arm64

Firefishy:
@saghul Nudge. Interested in this pull request? Happy to make any chances as required. Else feel free to take parts of it.

saghul:
Sorry for the lack of a reply here. I've been a bit busy lately. Will try to get to this next week.

saghul:
Quick update: I noticed rebasing my branch would take longer than taking yours :-) So I'm adding the missing parts to make builds with the Makefile and I'll refactor the release process to use buildx and we should be good to go.

saghul:
Question, what's that commit about the cache?

saghul:
Thanks! We have our first unstable builds with arm64 now :-) Thanks a lot!

PR #1275, opened 2022-04-28
jicofo: add optional XMPP_PORT value

aaronkvanmeerten:


env.example
```diff
@@ -233,6 +233,9 @@ XMPP_DOMAIN=meet.jitsi
 # Internal XMPP server
 XMPP_SERVER=xmpp.meet.jitsi
 
+# Internal XMPP server c2s port
+#XMPP_PORT=5222
```
saghul, line 237: Shouldn't prosody also listen on this port? That's what the name suggests.

env.example
```diff
@@ -233,6 +233,9 @@ XMPP_DOMAIN=meet.jitsi
 # Internal XMPP server
 XMPP_SERVER=xmpp.meet.jitsi
 
+# Internal XMPP server c2s port
+XMPP_PORT=5222
```
saghul, line 237: Please leave it commented out since we are providing a default

jibri/rootfs/defaults/jibri.conf
```diff
@@ -1,6 +1,7 @@
 {{ $JIBRI_USAGE_TIMEOUT := .Env.JIBRI_USAGE_TIMEOUT | default "0" -}}
 {{ $JIBRI_RECORDING_RESOLUTION := .Env.JIBRI_RECORDING_RESOLUTION | default "1280x720" -}}
 {{ $XMPP_TRUST_ALL_CERTS := .Env.XMPP_TRUST_ALL_CERTS | default "true" | toBool -}}
+{{ $ENABLE_XMPP_PORT := .Env.XMPP_PORT | default "0" | toBool }}
```
saghul, line 4: Why not use a single variable and use | default 5222 instead?

aaronkvanmeerten: Because I wanted to leave the default behavior and only specify a value explicitly if the variable is set. This way I'm not forcing the override to the default value but instead only providing the override when needed

env.example
aaronkvanmeerten: Done

env.example
aaronkvanmeerten: This has been implemented now

jibri/rootfs/defaults/jibri.conf
saghul: I don't see any value in that, and that's not how other options are implemented. Being explicit allows one to know exactly what the actual setting is.

docker-compose.yml
```diff
@@ -255,6 +255,7 @@ services:
             - XMPP_MUC_DOMAIN
             - XMPP_RECORDER_DOMAIN
             - XMPP_SERVER
+            - XMPP_PORT
```
saghul, line 258: you need to pass it to all services.

aaronkvanmeerten: done

jibri/rootfs/defaults/jibri.conf
aaronkvanmeerten: done

saghul (approved):
LGTM, but I haven't tested it. Thanks for going all the way on this one :-)

PR #1276, opened 2022-05-02
feature: support multiple XMPP servers via list

aaronkvanmeerten:


sawall (approved):

jvb/rootfs/defaults/jvb.conf
```diff
@@ -7,6 +7,8 @@
 {{ $WS_DOMAIN := .Env.JVB_WS_DOMAIN | default $PUBLIC_URL_DOMAIN -}}
 {{ $WS_SERVER_ID := .Env.JVB_WS_SERVER_ID | default .Env.JVB_WS_SERVER_ID_FALLBACK -}}
 {{ $XMPP_PORT := .Env.XMPP_PORT | default "5222" -}}
+{{ $XMPP_SERVERS := splitList "," .Env.XMPP_SERVER -}}
+{{ $c := . -}}
```
saghul, line 11: I don't understand this $c thing. It's not changed anywhere, so it's just a "." literal?

aaronkvanmeerten: It's because inside the range the . Context changed meaning to be the loop iterator. So we assign the context to another variable first so we can still reference it in the loop.

saghul: Got it. It looks very weird though. How about we save .Env in $ENV (and add a comment so we don't forget) ? It should be more obvious I think.

aaronkvanmeerten: Ok I'll test that

saghul (approved):

Issue #1207, opened 2022-01-25
Docker instructions and/or container broken


Myridium:
Followed instrructions here: https://jitsi.github.io/handbook/docs/devops-guide/devops-guide-docker

Ran
`docker-compose up`

Get this output:
```
docker-jitsi-meet-stable-6826-jicofo-1   | INFO: Started ServerConnector@2bc9a775{HTTP/1.1, (http/1.1)}{0.0.0.0:8888}
docker-jitsi-meet-stable-6826-jicofo-1   | Jan 25, 2022 2:32:42 AM org.eclipse.jetty.server.Server doStart
docker-jitsi-meet-stable-6826-jicofo-1   | INFO: Started @1312ms
docker-jitsi-meet-stable-6826-jvb-1      | Jan 25, 2022 2:32:42 AM org.eclipse.jetty.server.handler.ContextHandler doStart
docker-jitsi-meet-stable-6826-jvb-1      | INFO: Started o.e.j.s.ServletContextHandler@6f798482{/,null,AVAILABLE}
docker-jitsi-meet-stable-6826-jvb-1      | Jan 25, 2022 2:32:42 AM org.eclipse.jetty.server.AbstractConnector doStart
docker-jitsi-meet-stable-6826-jvb-1      | INFO: Started ServerConnector@6c000e0c{HTTP/1.1, (http/1.1)}{0.0.0.0:8080}
docker-jitsi-meet-stable-6826-jvb-1      | Jan 25, 2022 2:32:42 AM org.eclipse.jetty.server.Server doStart
docker-jitsi-meet-stable-6826-jvb-1      | INFO: Started @1389ms
docker-jitsi-meet-stable-6826-jvb-1      | Jan 25, 2022 2:32:46 AM org.jitsi.utils.logging2.LoggerImpl log
docker-jitsi-meet-stable-6826-jvb-1      | WARNING: Cannot set presence extension: not connected.
docker-jitsi-meet-stable-6826-jvb-1      | Jan 25, 2022 2:32:51 AM org.jitsi.utils.logging2.LoggerImpl log
docker-jitsi-meet-stable-6826-jvb-1      | INFO: Performed a successful health check in PT0.000884S. Sticky failure: false
docker-jitsi-meet-stable-6826-jvb-1      | Jan 25, 2022 2:32:51 AM org.jitsi.utils.logging2.LoggerImpl log
[...]

PR #1277, opened 2022-05-02
Add IPv6 note in readme

xeruf:
This might be the issue behind https://github.com/jitsi/docker-jitsi-meet/issues/1207

As this is a temporary case, I think putting it into the readme rather than the handbook is suitable, as people who are going to report an issue will see it then


saghul:
On the contrary, please add it to the handbook, since that's where all the installation instructions currently are. I don't want to have them split, sorry.

PR #1284, opened 2022-05-12
Add support for secrets

jlisher:
Added support for using secrets to store passwords and secrets. 

The file `/etc/cont-init.d/02-load-secrets`, added to the base image, fetches the secrets from `/run/secrets/` and stores them as new environment variables by creating a file with the name of the secret in `/run/s6/container_environment/`. The s6 init system will then make the secrets available as environment variables.
Note that there doesn't need to be an environment variable defining the location of the secrets, the existence of the secret in `/run/secrets/` (the default location) will make it available to the container. This should help prevent the need to edit any init or service run scripts.

There is also a `SECRETS_DIR` environment variable added to the `env.example` file. This should be used to specify the directory on the host where the secrets are stored. `gen-passwords.sh` has also been updated to use `SECRETS_DIR` as the location to store the generated secrets.

Edit: I just realised the environment variables could be left to exist and the use of secrets can be optional, the secrets will only replace the environment variables if they exist, otherwise the environment variables will remain untouched.


saghul:
Thoughts @aaronkvanmeerten ?

PR #1287, opened 2022-05-17
Simplify config

saghul:


jibri/rootfs/defaults/jibri.conf
```diff
@@ -84,7 +95,7 @@ jibri {
       }
     }
     recording {
-      recordings-directory = "{{ .Env.JIBRI_RECORDING_DIR }}"
+      recordings-directory = "{{ .Env.JIBRI_RECORDING_DIR | defailt "/config/recordings" }}"
```
prayagsingh, line 98: @saghul `default` typo here.

saghul: Thanks for taking a look!

PR #1301, opened 2022-06-02
Added prosody reservation config for docker

### LINKED ISSUE #1300: Jicofo Reservation Not Working on Jitsi
The reservation system is not working on the latest stable jitsi docker implementation. I have created endpoints on a server that's locally hosted and none of the endpoints are hit when the participant joins a room. 

I have added JICOFO_RESERVATION_ENABLED and JICOFO_RESERVATION_REST_BASE_URL to the env file.

I was digging into the jicofo source code and found that reservation feature was removed. 

But the documentation has nothing mentioned about it.

Mahdhir:
Fixes #1300


saghul:
Thank you! Shouldn't you also be deleting the existing reservation system config from jicofo, since it's no longer functional?

Mahdhir:
@saghul noted and removed the existing reservation system config from jicofo.
Also renamed the variables as they were not related Jicofo anymore.
Might need to update the documentation as well.

saghul (approved):

saghul:
Thank you!

Mahdhir:
Awesome.
@saghul  I have put a PR for the documentation as well.

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -161,6 +166,10 @@ VirtualHost "{{ $XMPP_DOMAIN }}"
     {{ end }}
     {{ end }}
 
+    {{ if $PROSODY_RESERVATION_ENABLED }}
+    reservations_api_prefix = { "{{ $PROSODY_RESERVATION_REST_BASE_URL }}" }
```
shawnchin, line 170: Should the curly braces around the var be there? That var is expected to be a string, not a table.

Should probably be:
```
reservations_api_prefix = "{{ $PROSODY_RESERVATION_REST_BASE_URL }}" 
```

Mahdhir: Noted. Will fix it

PR #1306, opened 2022-06-08
added advertise-private-candidates

tarokamikaze:


saghul:
This should be configurable. There are cases in which one wants the server's private address to be used, like in a LAN only deployment.

PR #1309, opened 2022-06-11
Update logging.properties

jvkassi:
Hello,

A quick PR to fix quotes in template file


saghul:
What was wrong about it?

PR #1313, opened 2022-06-13
jigasi: build on arm

saghul:


saghul:
WIP until we update the arch in the Debian files.

PR #1316, opened 2022-06-14
fix(web): start with clean config.js

wfleischer:
Fixes a bug introduced by https://github.com/jitsi/docker-jitsi-meet/pull/1287 causing configs to be constantly appended to config.js.


saghul (approved):

saghul:
Good catch!

PR #1319, opened 2022-06-14
feat(jvb): allow underscore in JVB_WS_SERVER_ID

wfleischer:
Swarm adds `_<service>` to build the service / container name. Adding the underscore to the regex will allow setting `JVB_WS_SERVER_ID` to the docker service name.


saghul:
Thanks Werner!

Issue #1310, opened 2022-06-12
Dumb Question


WriteCodeEveryday:
I'm using the prosody docker container.

I added ["muc_max_occupants" to XMPP_MUC_MODULES... ](https://github.com/jitsi/docker-jitsi-meet/blob/7c29b57914125ba37ea0df90242d2344b5d8f7fa/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua#L227) ... [how can I add the configurations to the block below it through ENV variables](https://github.com/jitsi/docker-jitsi-meet/blob/7c29b57914125ba37ea0df90242d2344b5d8f7fa/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua#L243=) since it resets the config every time the docker container starts up.

I'm [using the following config set](https://community.jitsi.org/t/solved-where-to-add-the-muc-max-occupants-to-limit-participants-in-a-meeting/73968/2)

 muc_max_occupants = "2"
 muc_access_whitelist = { "recorder@auth.meet.example.com" }

saghul:
It's currently not possible to do that.

saghul:
Given it's a module we supply we should make in configurable with env variables. Would you like to send a PR?

WriteCodeEveryday:
Yeah, it seems likely but I don't know if that would require a new release from jitsi? I'm currently on an old stable build.

saghul:
In order to be incorporated into the images yeah it would.

saghul:
A release of the images that is, not of the Jitsi Meet frontend or backend, that is.

PR #1324, opened 2022-06-20
Allow configuration of MUC Modules through ENV variables

WriteCodeEveryday:
This one is allowing configuration of added XMPP_MUC_MODULES for issue #1310


saghul:
Why not allow the 2 options you need (which are generic) instead of this catch-all?

WriteCodeEveryday:
Because doing so will allow us to configure a lot of modules that may require extra configuration.
If we're allowing any MUC module to be loaded by separating them by commas we should allow the same options for configuration @saghul

saghul:
Fair enough.

saghul (approved):

PR #1332, opened 2022-06-24
Bugfix

WriteCodeEveryday:
I clearly don't understand the synthax here. Can someone confirm that this is correct?

I got a config that had my data but was poorly formatted and had one too many strings and an ending semi-colon.

   muc_room_default_public_jids = true
    "muc_max_occupants="2""
"muc_other="30";


saghul:
You can download and test tpl here: https://github.com/jitsi/tpl

saghul (approved):

PR #1335, opened 2022-06-27
Add setting for encoded transform support

jan-di-cgm:
Add environment variable to configure `config.enableEncodedTransformSupport`


saghul:
Are you aware of the fact that this doesn't work 100% well yet?

jan-di-cgm:
Yes, we know that the feature is currently experimental. Our developers would like to test e2ee in production like environments.

saghul:
I see. You can also enable it with a config-custom.js file, would that work for you?

jan-di-cgm:
Yes, that is actually how we currently doing it. We extending the official images, where the `-custom` files are built via environment variables. This works - but as those are mostly generic settings, I would like to contribute them to the official image.

saghul (approved):

saghul:
Sorry, can you rebase please?

PR #1338, opened 2022-06-29
etherpad: add initial WBO integration

saghul:


web/rootfs/defaults/meet.conf
```diff
@@ -98,7 +98,20 @@ location ^~ /etherpad/ {
...
+location ^~ /wbo/ {
+    proxy_buffering off;
+    proxy_cache_bypass $http_upgrade;
+
+    proxy_http_version 1.1;
+    proxy_set_header Upgrade $http_upgrade;
+    proxy_set_header Connection "upgrade";
+    proxy_set_header X-Forwarded-For $remote_addr;
+
+    proxy_pass http://wbo.meet.jitsi/;
```
mihhu, line 114: Why is the port not necessary here?

saghul: Because the WBO container esposes port 80, and it doesn't need specifying since it's the default for HTTP.

jvkassi:
Hello, Asking politely, when can this be merged ?

saghul:
We are exploring the possibilities. It may never get merged in its current shape.

jvkassi:
What's blocking the merge ? I can take a look

jvkassi:
yeah, it's not the best. But there is no other WBO.
But can it do for now ? I mean it's the only way for now

Excalidraw seems like the best alternative way. But there is no plugin yet for jitsi

saghul:
You can already use WBO via Etherpad today, by using your own Etherpad installation.

The point of this PR is to make it easier for us to test and decide if that's the way we want to go or not.

jvkassi:
Yeah I can do that, but I prefered the way you did with plugins already installed, proxy in the nginx config

jvkassi:
I found this package https://github.com/cracker0dks/whiteboard
used by this fork of  jitsi :  https://meetzi.de/

Once this [feature](https://github.com/jitsi/jitsi-meet/pull/11052) is merged, it should be easy to use like https://github.com/cracker0dks/whiteboard/issues/78#issuecomment-757963187

saghul:
We have decided to not go forward with WBO. We are working on natively integrating another open source whiteboard solution. Stay tuned :-)

saghul:
I may rescue this PR to improve the Etheroad integration though, since one needs to build their own image to add plugins.

PR #1356, opened 2022-08-05
Added functionality for docker container to set max occupants allowed in a single room

RhysRdm:
I have added the option to be able to set a max number of occupants into a room via the docker jitsi-meet.cfg.lua template file.

It keys off if a user has set the ENVIRONMENT variable MAX_PARTICIPANTS in there env file or docker run command. Ie to specify 25 occupants max in a room set an env variable MAX_PARTICIPANTS=25

Please let me know what you think or if you have any comments

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -207,6 +207,7 @@ services:
             - LDAP_START_TLS
             - LDAP_URL
             - LDAP_USE_TLS
+            - MAX_PARTICIPANTS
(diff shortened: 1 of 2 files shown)


saghul (changes requested):
Looking good! Can you please add the env var to the compose file too please?

RhysRdm:
Thank you for the quick review @saghul

saghul (approved):

PR #1365, opened 2022-08-12
Update etherpad.yml

filwu8:
set db for eatherpad
Fixed when container restarts missing text


saghul:
Ditto as in https://github.com/jitsi/docker-jitsi-meet/pull/1364

PR #1367, opened 2022-08-15
add(web): enableFeaturesBasedOnToken support

debendraoli:


saghul:
@damencho did we talk about deprecating this feature or did I dream about it?

damencho:
We had dropped part of it. But that is still available, whether UI to respect what is in the token.

saghul:
> We had dropped part of it. But that is still available, whether UI to respect what is in the token.

Let's sync with Emil, IIRC this was meant to disappear to one point after Meet Pro was sidelined.

damencho:
People from jaas are also using it, so doubt that to be dropped.

saghul:
After discussing with @damencho, the plan is to axe the setting altogether: https://github.com/jitsi/jitsi-meet/pull/12028

debendraoli:
Looks good then.

PR #1379, opened 2022-08-19
JVB: allow configuration of videobridge.ice.advertise-private-candidates

benbz:
Similar to https://github.com/jitsi/docker-jitsi-meet/pull/1306 but keeps the current default and is configurable via an ENV var

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -298,6 +298,7 @@ services:
             - ENABLE_COLIBRI_WEBSOCKET
             - ENABLE_OCTO
             - ENABLE_MULTI_STREAM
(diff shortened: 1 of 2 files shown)


saghul:
When would this be helpful?

benbz:
> When would this be helpful?

We've got some JVBs in Kubernetes. Currently they're advertising both the node's public IP (as discovered by STUN) and the pod IP. It all works now but the pod IP is never going to be usable by users and just adds noise when we're investigating issues

saghul (approved):

saghul:
Gotcha, makes sense!

PR #1382, opened 2022-08-21
add jigasi as admin

tbutter:
When token authentication is enabled jigasi is not able to join rooms that request transcription. This pull request adds jigasi as admin to be able to join any room.

### DIFF
--- a/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
+++ b/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
@@ -3,6 +3,7 @@
 {{ $ENABLE_RECORDING := .Env.ENABLE_RECORDING | default "0" | toBool }}
 {{ $AUTH_TYPE := .Env.AUTH_TYPE | default "internal" }}
 {{ $JICOFO_AUTH_USER := .Env.JICOFO_AUTH_USER | default "focus" -}}
+{{ $JIGASI_XMPP_USER := .Env.JIGASI_XMPP_USER | default "jigasi" -}}
 {{ $JVB_AUTH_USER := .Env.JVB_AUTH_USER | default "jvb" -}}
 {{ $JWT_ASAP_KEYSERVER := .Env.JWT_ASAP_KEYSERVER | default "" }}
 {{ $JWT_ALLOW_EMPTY := .Env.JWT_ALLOW_EMPTY | default "0" | toBool }}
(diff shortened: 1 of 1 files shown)


debendraoli:
Likely related to https://github.com/jitsi/jitsi-meet/issues/11967

tbutter:
my setup does not allow to join a room without a token. i don't think its related. the behaviour seems to be correct (jigasi sends no token -> no entry to room). adding jigasi to admins allows it to join any room - just like jicofo.

damencho:
I think that is the same, the same happens to jibri. 
And if you set `enable_domain_verification = false` you will be able to join with jigasi without it being an admin.

saghul:
Si question @damencho: does the JVB user need to be admin?

damencho:
> Si question @damencho: does the JVB user need to be admin?

Here yes, https://github.com/jitsi/jitsi-meet/blob/40637aa3dcc25be6a1a506787dd995882547d93e/doc/debian/jitsi-meet-prosody/prosody.cfg.lua-jvb.example#L106
It is gor the case when you restart prosody, who will be the first to join to create the brewery.

saghul:
Ack! Don't we have the same case for the jigasi brewery here then?

debendraoli:
From what I observed, I see jigasi brewery  passed down to the jicofo and it is taken care of.

I ain't expert of XMPP, I believe brewery are some sort of identifier to filer messages or so. IDK.  


https://github.com/jitsi/jicofo/blob/c03962f24242210d77efd2bd4885b83b12dbb58e/src/main/kotlin/org/jitsi/jicofo/jigasi/JigasiConfig.kt#L30

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -31,8 +32,17 @@
 {{ $ENABLE_SUBDOMAINS := .Env.ENABLE_SUBDOMAINS | default "true" | toBool -}}
 {{ $PROSODY_RESERVATION_ENABLED := .Env.PROSODY_RESERVATION_ENABLED | default "false" | toBool }}
 {{ $PROSODY_RESERVATION_REST_BASE_URL := .Env.PROSODY_RESERVATION_REST_BASE_URL | default "" }}
+{{ $JIBRI_XMPP_USER := .Env.JIBRI_XMPP_USER | default "jibri" -}}
```
saghul, line 35: Please sort these alphabetically.

saghul (approved):
LGTM. @damencho ?

damencho (approved):

PR #1384, opened 2022-08-25
fix: Fixes missing lua dependency for inspect.

damencho:


damencho:
Hey @saghul can you test this with the room-info endpoint, does that fixes the issue not loading that module. Thanks.

saghul (approved):

Issue #1389, opened 2022-08-30
Endconference component not working.


debendraoli:
The last merged feature seems incomplete.

When I try to end the conference for all, there is no effect on room and following error is displayed on prosody logs.

`endconference.jitsi.meet:end_conference`                      `warn`    `No room found found for /roomName`

saghul:
Are you using `unstable` ? Ping @wfleischer :-)

debendraoli:
yes, Also I am getting same error with breakout room too. Perhaps something to do with subdomain enabled?

saghul:
Maybe. Do you observe any problems if you don't use a subdomain?

debendraoli:
I cannot join any meetings at all without any logs.

debendraoli:
I cannot join any meetings at all. I guess it is not subdomain issue as it is there for months.

wfleischer:
Hi @debendraoli. Thanks for reporting. End conference works well with the following plain setup:

```
git clone https://github.com/jitsi/docker-jitsi-meet
cd docker-jitsi-meet
cp env.example .env
./gen-passwords.sh  
docker-compose up -d
```

The log line you provided is from https://github.com/jitsi/jitsi-meet/blob/master/resources/prosody-plugins/mod_end_conference.lua#L55. Obviously the `session.jitsi_web_query_room` and `session.jitsi_web_query_prefix` values are not populated accordingly. That also explains why breakout rooms are broken for you (compare https://github.com/jitsi/jitsi-meet/blob/master/resources/prosody-plugins/mod_muc_breakout_rooms.lua#L259).

I tend to think that there is an issue with your configuration. Can you provide details that will allow us to duplicate?

saghul:
We have no indication that it's broken. Please share your config and full logs.

debendraoli:
Above logs is all I got, I manged to fix it by properly populating  `jitsi_web_query_prefix` with right subdomain.

debendraoli:
`jitsi_web_query_prefix` value is empty so it cannot get room when subdomain is used.

It is expecting prefix from http request but its is not sent from jitsi web hence empty. 

https://github.com/jitsi/jitsi-meet/blob/6fe353cfa4774a7c52c14651b0bbefa7895d4bf5/resources/prosody-plugins/mod_jitsi_session.lua#L23

And in breakout or end conference module it is trying to get room using the prefix as subdomain but it is empty so the same error message I reported above.

[...]

PR #1390, opened 2022-08-31
fix(web): nginx pass prefix for subdomains

debendraoli:
fixes: https://github.com/jitsi/docker-jitsi-meet/issues/1389


DanielMcAssey:
Should probably be wrapped in $ENABLE_SUBDOMAINS

debendraoli:
> 

That would complicate things, initial value will just do the work.

saghul (approved):

PR #1395, opened 2022-09-05
web: add ability to configure the room password digit size

saghul:


damencho:
This switches from any character to just digits by default. Is this ok?

saghul:
> This switches from any character to just digits by default. Is this ok?

I don't know TBH. For JaaS components we need to change it to digits, how did you handle that in the Debian package?

damencho:
That setting is off for the debian packages and anyone can override it if desired. 
So you say we need to change that for the JaaS components? I'm currently doing some modifications around that for the debian packages.

saghul:
Yep because for jigasi we need digits only right?

Maybe we can do the following: if JaaS components are enabled  set the default to 10 digits, unless the user has explicitly set it to something.

WDYT?

damencho:
Make sense.

damencho (approved):

damencho (approved):

PR #1396, opened 2022-09-05
fix JWT_ENABLE_DOMAIN_VERIFICATION missing in docker-compose.yml

chaosbiber:
variable JWT_ENABLE_DOMAIN_VERIFICATION was added in stable-7648 to prosody files but enabling in .env had no effects


saghul (approved):
Oops, my bad!

Issue #1394, opened 2022-09-03
Matrix auth broken


ChrisPortman:
Prosidy fails to start with auth set to matrix because it cant find the luajwtjitsi module.

It appears the following line was removed from the dockerfile a few releases ago:

luarocks install luajwtjitsi 3.0-0

(Apologies for the rubbish text, doing this from my phone. Can add more verbose output later if required)

Thanks!

amg-web:
also jwt does not work

saghul:
The plugin is now builtin, as in, it's provided by the Jitsi Meet package. Do you have an error trace?

saghul:
> also jwt does not work

Do you have an error you can share? What release? Last I checked it was working.

ChrisPortman:
The release is 7648-4. Just updated today.

Ill be able to get a trace shortly.

Thanks

ChrisPortman:
```
prosody_1   | modulemanager                                                        error	Error initializing module 'auth_matrix_user_verification' on 'vtc.jitsi.internal': /usr/share/lua/5.4/prosody/util/startup.lua:202: module 'luajwtjitsi' not found:
prosody_1   | 	No LuaRocks module found for luajwtjitsi
prosody_1   | 	no field package.preload['luajwtjitsi']
prosody_1   | 	no file '/usr/local/share/lua/5.4/prosody/luajwtjitsi.lua'
prosody_1   | 	no file '/usr/local/share/lua/5.4/luajwtjitsi.lua'
prosody_1   | 	no file '/usr/local/share/lua/5.4/prosody/luajwtjitsi/init.lua'
prosody_1   | 	no file '/usr/local/share/lua/5.4/luajwtjitsi/init.lua'
prosody_1   | 	no file '/usr/local/lib/lua/5.4/prosody/luajwtjitsi.lua'
prosody_1   | 	no file '/usr/local/lib/lua/5.4/luajwtjitsi.lua'
prosody_1   | 	no file '/usr/local/lib/lua/5.4/prosody/luajwtjitsi/init.lua'
prosody_1   | 	no file '/usr/local/lib/lua/5.4/luajwtjitsi/init.lua'
prosody_1   | 	no file '/usr/share/lua/5.4/prosody/luajwtjitsi.lua'
prosody_1   | 	no file '/usr/share/lua/5.4/luajwtjitsi.lua'
prosody_1   | 	no file '/usr/share/lua/5.4/prosody/luajwtjitsi/init.lua'
[...]

PR #1400, opened 2022-09-09
Update Dockerfile

thermaq:
Added missing requirement for matrix authentication

### DIFF
--- a/prosody/Dockerfile
+++ b/prosody/Dockerfile
@@ -11,10 +11,13 @@ RUN apt-dpkg-wrap apt-get update && \
       libreadline-dev \
       git \
       unzip \
+      openssl \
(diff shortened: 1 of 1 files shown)


saghul:
PLease check the open issues, this is a bug on the MAtrix side after we changed the way we bundle the library.

thermaq:
Which open issue, there are 200 of them? You bundle a mod_auth_matrix_ now out of the box and it just needs one package so it seems like it needs to be fixed here.

saghul:
My bad, the user closed the issue: https://github.com/jitsi/docker-jitsi-meet/issues/1394

The JWT plugin is builtin now so am I port needs to be adjusted, no plugin is missing.

PR #1412, opened 2022-10-05
fix: missing configuration options

charles-zablit:
This PR fixes missing configurations from the latest version of Jitsi-meet (2.0.7830)

The [new configuration options for transcription](https://github.com/jitsi/jitsi-meet/blob/da9b5a91568f4d2544a93ab0def7bb67bdc9009c/config.js#L376-L407) are now implemented.

Please let me know if there is anything I can edit.

PS: I have signed the CLA.

### DIFF
--- a/web/rootfs/defaults/settings-config.js
+++ b/web/rootfs/defaults/settings-config.js
@@ -22,6 +22,12 @@
 {{ $ENABLE_TALK_WHILE_MUTED := .Env.ENABLE_TALK_WHILE_MUTED | default "false" | toBool -}}
 {{ $ENABLE_TCC := .Env.ENABLE_TCC | default "true" | toBool -}}
 {{ $ENABLE_TRANSCRIPTIONS := .Env.ENABLE_TRANSCRIPTIONS | default "false" | toBool -}}
+{{ $TRANSLATION_LANGUAGES := .Env.TRANSLATION_LANGUAGES | default "['en', 'fr']" -}}
+{{ $TRANSLATION_LANGUAGES_HEAD := .Env.TRANSLATION_LANGUAGES_HEAD | default "['en']" -}}
+{{ $USE_APP_LANGUAGE := .Env.USE_APP_LANGUAGE | default "true" | toBool -}}
+{{ $PREFERRED_LANGUAGE := .Env.PREFERRED_LANGUAGE | default "en-US" -}}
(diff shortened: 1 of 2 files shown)


web/rootfs/defaults/settings-config.js
```diff
@@ -22,6 +22,12 @@
 {{ $ENABLE_TALK_WHILE_MUTED := .Env.ENABLE_TALK_WHILE_MUTED | default "false" | toBool -}}
 {{ $ENABLE_TCC := .Env.ENABLE_TCC | default "true" | toBool -}}
 {{ $ENABLE_TRANSCRIPTIONS := .Env.ENABLE_TRANSCRIPTIONS | default "false" | toBool -}}
+{{ $TRANSLATION_LANGUAGES := .Env.TRANSLATION_LANGUAGES | default "['en', 'fr']" -}}
```
saghul, line 25: I don't think we should provide a default here.

saghul (changes requested):

web/rootfs/defaults/settings-config.js
charles-zablit: fixed 👍

web/rootfs/defaults/settings-config.js
```diff
@@ -22,6 +22,12 @@
 {{ $ENABLE_TALK_WHILE_MUTED := .Env.ENABLE_TALK_WHILE_MUTED | default "false" | toBool -}}
 {{ $ENABLE_TCC := .Env.ENABLE_TCC | default "true" | toBool -}}
 {{ $ENABLE_TRANSCRIPTIONS := .Env.ENABLE_TRANSCRIPTIONS | default "false" | toBool -}}
+{{ $TRANSLATION_LANGUAGES := .Env.TRANSLATION_LANGUAGES -}}
```
saghul, line 25: Doesn't this need a default?

charles-zablit: Sorry, I'm confused, should I add a default? Your previous change request said we shouldn't add defaults.

Would an empty default like below be appropriate?
`{{ $TRANSLATION_LANGUAGES := .Env.TRANSLATION_LANGUAGES | default "[]" -}}`

saghul: Yeah, that looks ok.

charles-zablit: Alright, fixed 😄

saghul (approved):

charles-zablit:
No problem, glad I could contribute!

PR #1422, opened 2022-10-21
Adding single-use-mode flag to Jibri config

Tobaroo:
Adding the missing `single-use-mode` flag to the `jibri.conf` template, defaulting to `false`. The flag is already present in the [reference file](https://github.com/jitsi/jibri/blob/master/src/main/resources/reference.conf) for Jibri.

It defines whether or not Jibri should return to idle state after handling (successfully or unsuccessfully) a request. A value of 'true' here means that a Jibri will NOT return back to the IDLE state and will need to be restarted in order to be used again.

### DIFF
--- a/jibri/rootfs/defaults/jibri.conf
+++ b/jibri/rootfs/defaults/jibri.conf
@@ -3,6 +3,7 @@
 {{ $JIBRI_USAGE_TIMEOUT := .Env.JIBRI_USAGE_TIMEOUT | default "0" -}}
 {{ $JIBRI_RECORDING_RESOLUTION := .Env.JIBRI_RECORDING_RESOLUTION | default "1280x720" -}}
 {{ $JIBRI_BREWERY_MUC := .Env.JIBRI_BREWERY_MUC | default "jibribrewery" -}}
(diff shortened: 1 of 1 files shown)


jibri/rootfs/defaults/jibri.conf
```diff
@@ -20,6 +21,11 @@
 jibri {
     // A unique identifier for this Jibri
     id = "{{ .Env.JIBRI_INSTANCE_ID }}"
+    // Whether or not Jibri should return to idle state after handling
+    // (successfully or unsuccessfully) a request.  A value of 'true'
+    // here means that a Jibri will NOT return back to the IDLE state
+    // and will need to be restarted in order to be used again.
+    single-use-mode = {{ .Env.JIBRI_SINGLE_USE_MODE }}
```
saghul, line 28: You are not using the variable you decalred here. Use $JIBRI_SINGLE_USE_MODE.

saghul (commented):
Left a comment. You also need to add this new variable to jibri.yml

jibri/rootfs/defaults/jibri.conf
Tobaroo: You are so right. I corrected that line.

saghul (approved):

PR #1424, opened 2022-10-24
Use 'stable' version for docker images

marius-bardan:


saghul:
git master targets unstable, that's why is the default, we then change it to the specific tag during the release process, so a release has pinned images.

I have tried this before, but it's a fool's errand, because we add options for new features that are not yet in stable.

marius-bardan:
got it. thanks @saghul

PR #1440, opened 2022-11-15
Register jaas account automatically

oanaianc:


damencho:
Do we need a copy of that if its available in the web container that holds jitsi-meet?

register-jaas-account.sh
```diff
@@ -0,0 +1,85 @@
...
+    echo "Validating domain failed. Status: ${ready_error}"
+    echo "Response: "
+    echo "${ready_data}" | jq -r
+    echo "${SUPPORT_MSG}"
+    echo
+    exit 3
+fi
+
+SLEEP_TIME=0
+WAIT_BEFORE_CHECK=10
```
saghul, line 41: Could we lower this to 5?

saghul:
LEt's put this in web/rootfs/etc/service.d/jaas-account/run

saghul:
> Do we need a copy of that if its available in the web container that holds jitsi-meet?

It will be adapted to work in the container environment.

web/rootfs/etc/services.d/jaas-account/run
```diff
@@ -0,0 +1,122 @@
...
+rm ${CHALLENGE_FILE} || true
+
+fi
+stop_service
+
+function stop_service() {
+  s6-svc -O /var/run/s6/services/jaas-account
+}
+
+function check_nginx() {
```
saghul, line 120: Since we only use this in 1 place, perhaps better to inline it?

web/rootfs/etc/services.d/jaas-account/run
```diff
@@ -0,0 +1,122 @@
...
+CHALLENGE_FILE="/usr/share/jitsi-meet/.well-known/jitsi-challenge.txt"
+SUPPORT_MSG="Reach out to JaaS support at https://jaas.8x8.vc/#components"
+JAAS_ACCOUNT_FILE="/config/jaas-account-created.txt"
+
+if [[ $DISABLE_HTTPS -ne 1 ]] && [[ $ENABLE_LETSENCRYPT -eq 1 ]]
+  && [[ $ENABLE_JAAS_COMPONENTS -eq 1 ]]  && [[ -e $EMAIL ]] && [[ -e $DOMAIN ]]; then
+
+if [ -f $JAAS_ACCOUNT_FILE ]; then
+    echo "JaaS account already exists"
+    stop_service
```
saghul, line 18: We need to exit here, or maybe just exit in the stop_service function. Here and elsewhere.

register-jaas-account.sh
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/with-contenv bash
```
saghul, line 1: Drop this file now?

saghul (changes requested):

register-jaas-account.sh
oanaianc: yes

saghul (approved):

PR #1449, opened 2022-12-06
Change TURN transport from TCP to UDP

adamsko:
This pull request changes the transport mode for "turn" server from TCP to UDP.

This change allows peer to peer connections trough the TURN server to use UDP transport for the client->TURN server communication.

Using UDP for turn transport is also what meet.jit.si uses in its ICE server candidates.

Community thread about this: https://community.jitsi.org/t/jitis-docker-turn-transport-tcp-should-be-udp/119374

### DIFF
--- a/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
+++ b/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
@@ -67,7 +67,7 @@ external_service_secret = "{{.Env.TURN_CREDENTIALS}}";
 {{ if or .Env.TURN_HOST .Env.TURNS_HOST }}
 external_services = {
   {{ if .Env.TURN_HOST }}
-     { type = "turn", host = "{{ .Env.TURN_HOST }}", port = {{ $TURN_PORT }}, transport = "tcp", secret = true, ttl = 86400, algorithm = "turn" }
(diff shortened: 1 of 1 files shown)


saghul:
This is not what I mentioned in the community forum. We should add an option for both, not just replace tcp with udp.

adamsko:
Would adding an env variable called TURN_TRANSPORT with default set to tcp (to keep current behaviour) be an acceptable solution?

adamsko:
I updated the PR. By both do you mean for turns as well? I can of course add a TURNS_TRANSPORT variable but I am unsure what transport=udp for turns means or if it is actually meaningful. Some googling seems to indicate that turns+udp means use dtls which might not be supported by webrtc implementations.

bgrozev:
I think the client already has code to exclude the unnecessary entries for the jvb connection, so just defalting to enable all of turn/tcp, turn/udp and turns is a reasonable solution IMO

saghul:
> I updated the PR. By both do you mean for turns as well? I can of course add a TURNS_TRANSPORT variable but I am unsure what transport=udp for turns means or if it is actually meaningful. Some googling seems to indicate that turns+udp means use dtls which might not be supported by webrtc implementations.

A single variable is fine, but you need to split on the comma and add an entry for each transport.

adamsko:
It now adds a turn entry for every transport specified in TURN_TRANSPORT (comma separated). If TURN_TRANSPORT is not specified it only adds an entry with transport=tcp as before.

saghul:
Wonderful!

saghul (approved):

PR #1452, opened 2022-12-13
fix: missing quotes on string.

debendraoli:
This is breaking config.js loading.


saghul (approved):

saghul:
Thank you! I'll do a -1 release.

PR #1453, opened 2022-12-15
Ability to configure whiteboard.

debendraoli:
- Ability to configure whiteboard.
- Replace deprecated es function `substr()` with `substring()`.

### DIFF
--- a/web/rootfs/defaults/settings-config.js
+++ b/web/rootfs/defaults/settings-config.js
@@ -63,7 +63,8 @@
 {{ $ENABLE_LOCAL_RECORDING_SELF_START := .Env.ENABLE_LOCAL_RECORDING_SELF_START | default "false" | toBool -}}
 {{ $DISABLE_PROFILE := .Env.DISABLE_PROFILE | default "false" | toBool -}}
 {{ $ROOM_PASSWORD_DIGITS := .Env.ROOM_PASSWORD_DIGITS | default "false" -}}
-
+{{ $WHITEBOARD_COLLAB_SERVER_BASE_URL := .Env.WHITEBOARD_COLLAB_SERVER_BASE_URL | default "" -}}
+{{ $WHITEBOARD_ENABLED := .Env.WHITEBOARD_ENABLED | default "false" | toBool -}}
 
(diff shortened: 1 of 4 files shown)


web/rootfs/defaults/settings-config.js
```diff
@@ -63,7 +63,8 @@
 {{ $ENABLE_LOCAL_RECORDING_SELF_START := .Env.ENABLE_LOCAL_RECORDING_SELF_START | default "false" | toBool -}}
 {{ $DISABLE_PROFILE := .Env.DISABLE_PROFILE | default "false" | toBool -}}
 {{ $ROOM_PASSWORD_DIGITS := .Env.ROOM_PASSWORD_DIGITS | default "false" -}}
-
+{{ $WHITEBOARD_COLLAB_SERVER_BASE_URL := .Env.WHITEBOARD_COLLAB_SERVER_BASE_URL | default "" -}}
```
saghul, line 66: Can you please rename this to be `WHITEBOARD_COLLAB_SERVER_PUBLIC_URL` ? That would make it analogous to Etherpad, where use the "base" URL when it's builtin, and the other one when it's external.

saghul (changes requested):
The room_metadata plugin needs to be loaded too. We probably want to do that regardless of the whiteboard option, since it's generic.

saghul (approved):

andys448:
Hi! 

ENV - WHITEBOARD_COLLAB_SERVER_BASE_URL not working correctly. After filling url - generated variable is still empty in web / config.js

debendraoli:
correct env name is `WHITEBOARD_COLLAB_SERVER_PUBLIC_URL`.

andys448:
> correct env name is `WHITEBOARD_COLLAB_SERVER_PUBLIC_URL`.

Thanks! Works fine)

jvkassi:
Hello, the button whiteboard is only avaible for moderators. Is the default ? Can i change this config  ?

jvkassi:
i was missing this config 
https://github.com/jitsi/docker-jitsi-meet/pull/1457

PR #1489, opened 2023-03-06
Fix "<no value>" issue in prosody config

orandev:
If XMPP_DOMAIN and XMPP_MUC_DOMAIN are not configured in the .env file (which is the default situation), we get the following lines in jitsi-meet.cfg.lua:

`    end_conference_component = "endconference.<no value>"`

```
Component "endconference.<no value>" "end_conference"
    muc_component = "<no value>"
```


saghul:
You need the $ sign to use it as a variable: $XMPP_DOMAIN.

saghul (approved):

PR #1494, opened 2023-03-15
jicofo: enable rest interface

aaronkvanmeerten:


docker-compose.yml
```diff
@@ -255,6 +255,8 @@ services:
     jicofo:
         image: jitsi/jicofo:${JITSI_IMAGE_VERSION:-unstable}
         restart: ${RESTART_POLICY:-unless-stopped}
+        expose:
```
saghul, line 258: I think using port and controlling the host there (like we do with the JVB colibri endpoint) makes more sense.

Jicofo would also be exporting the port because from the container perspective localhost is the container itself, so nobody would have access to it, not even the Docker host.

Or did I miss something?

saghul (commented):
Left some quick feedback.

docker-compose.yml
aaronkvanmeerten: I think that's true, good point.  I've updated based on your feedback
.

saghul (approved):

PR #1523, opened 2023-04-04
workflow: disable schedule runs on new packages

aaronkvanmeerten:


damencho (approved):

saghul:
When are the unstable images going to get built then?

saghul:
But people don't know how to pair each container versions, do they? Having a daily unstable marks a checkpoint. Any specific reason why we no longer want that, in addition to the other tags?

Screen Shot 2023-06-06 at 14 20 25

aaronkvanmeerten:
Seems ok to leave them on then

PR #1549, opened 2023-05-16
jibri: pre-warm chrome first before starting jibri

aaronkvanmeerten:


jibri/rootfs/etc/services.d/40-jibri/run
```diff
@@ -4,5 +4,7 @@
 HOME=/home/jibri
 
 DAEMON=/opt/jitsi/jibri/launch.sh
+# pre-warm google chrome before jibri launches to ensure fast chrome launch during recordings
+s6-setuidgid jibri /usr/bin/google-chrome --timeout=1000 --headless https://example.com/
```
saghul, line 8: We should use PUBLIC_URL here, I think.

PR #1554, opened 2023-05-23
fix: update TAG as stable instead of latest

emrahcom:
This commit sets the tag as `stable` because `latest` is not available as a tag anymore.


saghul:
The reason why that tag is set to latest is for development. If you just `make` the generated images will have a tag of latest.

When building for release we specify it.

emrahcom:
Sorry, I though that there is a TAG issue when I could not build the image by `docker build`

PR #1211, opened 2022-01-28
prosody: allow multiple turn servers

wfleischer:

PR #1559, opened 2023-06-01
prosody: allow multiple TURN

emrahcom:
This PR allows to set multiple `TURN` or `TURNS` host for `prosody`. 

It uses currently available environment variables, `TURN_HOST` and `TURNS_HOST`. So these variables may be set by using only one host or comma seperated multiple host list.

This PR is based on @wfleischer's https://github.com/jitsi/docker-jitsi-meet/pull/1211 with some updates.


saghul:
@aaronkvanmeerten can you PTAL?

aaronkvanmeerten (approved):
Looks good to me, I don't love the comma logic but I took a stab at a better if statement that took into account idx1 or idx2 greater than 1 and it was even uglier.  Nice work.

wfleischer:
Thanks, @emrahcom, for this initiative!

wfleischer:
> Looks good to me, I don't love the comma logic but I took a stab at a better if statement that took into account idx1 or idx2 greater than 1 and it was even uglier. Nice work.

This is how it was done at #1211: https://github.com/jitsi/docker-jitsi-meet/pull/1211/files#diff-199f45103a3a23a514583143479a6ce9a61672842b115a183202503dbf315459R43.

PR #1560, opened 2023-06-02
prosody: update the comma logic while listing TURNs

emrahcom:
This PR is related with PR https://github.com/jitsi/docker-jitsi-meet/pull/1559 which updates the comma logic while listing TURN servers.

@aaronkvanmeerten  can you PTAL?


aaronkvanmeerten (approved):
It looks good to me, so as long as it actually parses I like it better. Thanks for taking the time to make this more aesthetically pleasing to me :)

PR #1561, opened 2023-06-05
New env var JICOFO_ENABLE_AUTH

luiru72:
Introduction of a new JICOFO_ENABLE_AUTH environment variable, to make it possible to enable authentication in Jicofo indipendently from the other containers.

For example, you need to enable JWT authentication (so ENABLE_AUTH should be true), but you want to use the token affiliation plugin (so authentication should be false in Jicofo).

See: https://community.jitsi.org/t/plugin-token-affiliation-mod-token-affiliation-lua-not-working/123852/15


docker-compose.yml
```diff
@@ -274,6 +274,7 @@ services:
             - BRIDGE_AVG_PARTICIPANT_STRESS
             - BRIDGE_STRESS_THRESHOLD
             - ENABLE_AUTH
+            - JICOFO_ENABLE_AUTH
```
saghul, line 277: Please put this one alphabetically sorted.

env.example
```diff
@@ -113,6 +113,9 @@ ETHERPAD_SKIN_VARIANTS="super-light-toolbar super-light-editor light-background
 # Enable authentication
 #ENABLE_AUTH=1
 
+# Enable authentication in Jicofo: defaults to ENABLE_AUTH
+#JICOFO_ENABLE_AUTH=1
```
saghul, line 117: No need to document it here. Do so in the handbook please. The sample env is just for the essentials.

saghul (commented):
A couple of minor comments!

docker-compose.yml
luiru72: Done! :)

saghul (approved):

saghul:
Thank you!

PR #1562, opened 2023-06-05
jibri: add autoscaler-sidecar service support

aaronkvanmeerten:


jibri.yml
```diff
@@ -13,6 +13,15 @@ services:
             - CHROMIUM_FLAGS
             - DISPLAY=:0
             - ENABLE_STATS_D
+            - AUTOSCALER_URL
```
saghul, line 16: Alpha sort please.

jibri/rootfs/defaults/autoscaler-sidecar.config
```diff
@@ -0,0 +1,14 @@
+export PORT={{ .Env.AUTOSCALER_SIDECAR_PORT }}
```
saghul, line 1: This port is not present in the compose ports / expose section. Does it need to? Also no default?

jibri/rootfs/defaults/autoscaler-sidecar.config
```diff
@@ -0,0 +1,14 @@
...
+export GRACEFUL_SHUTDOWN_SCRIPT="/opt/jitsi/jibri/wait_graceful_shutdown.sh"
+export TERMINATE_SCRIPT="/opt/jitsi/jibri/shutdown.sh"
+export RECONFIGURE_SCRIPT=
```
saghul, line 4: ?

jibri/rootfs/defaults/autoscaler-sidecar.config
```diff
@@ -0,0 +1,14 @@
...
+export ENABLE_REPORT_STATS=true
+export POLLING_URL="{{ .Env.AUTOSCALER_URL  }}/sidecar/poll"
+export STATUS_URL="{{ .Env.AUTOSCALER_URL }}/sidecar/status"
+export STATS_RETRIEVE_URL="http://localhost:2222/jibri/api/v1.0/health"
```
saghul, line 8: Isn't that port conficurable?

jibri/rootfs/defaults/autoscaler-sidecar.config
```diff
@@ -0,0 +1,14 @@
...
+export STATS_REPORT_URL="{{ .Env.AUTOSCALER_URL }}/sidecar/stats"
+export ASAP_SIGNING_KEY_FILE="{{ .Env.AUTOSCALER_SIDECAR_KEY_FILE }}"
+export ASAP_JWT_KID="{{ .Env.AUTOSCALER_SIDECAR_KEY_ID }}"
+export INSTANCE_TYPE="jibri"
```
saghul, line 12: Should this be configurable?

jibri.yml
```diff
@@ -13,6 +13,15 @@ services:
...
+            - AUTOSCALER_SIDECAR_KEY_FILE
+            - AUTOSCALER_SIDECAR_KEY_ID
+            - AUTOSCALER_SIDECAR_PORT
+            - AUTOSCALER_SIDECAR_INSTANCE_ID
+            - AUTOSCALER_SIDECAR_REGION
+            - AUTOSCALER_SIDECAR_GROUP_NAME
+            - LOCAL_ADDRESS
```
saghul, line 23: What is this supposed to do?

jibri/rootfs/defaults/autoscaler-sidecar.config
```diff
@@ -0,0 +1,14 @@
...
+export INSTANCE_ID="{{ .Env.AUTOSCALER_SIDECAR_INSTANCE_ID }}"
+export INSTANCE_METADATA='{"environment":"{{ .Env.XMPP_ENV_NAME }}","region":"{{ .Env.AUTOSCALER_SIDECAR_REGION }}","group":"{{ .Env.AUTOSCALER_SIDECAR_GROUP_NAME }}","name":"{{ .Env.JIBRI_INSTANCE_ID }}","version":"{{ .Env.JIBRI_VERSION }}","privateIp":"{{ .Env.LOCAL_ADDRESS }}"}'
```
saghul, line 14: is the private IP relevant?

jibri/rootfs/etc/cont-init.d/10-config
```diff
@@ -30,6 +30,40 @@ fi
...
+    fi
+
+    if [ -f "$AUTOSCALER_SIDECAR_KEY_FILE" ]; then
+        echo "AUTOSCALER_URL found, enabling autoscaler sidecar"
+
+        export JIBRI_VERSION="$(dpkg -s jibri | grep Version | awk '{print $2}' | sed 's/..$//')"
+
+        [ -z "$AUTOSCALER_SIDECAR_PORT" ] && export AUTOSCALER_SIDECAR_PORT="6000"
+        [ -z "$JIBRI_WEBHOOK_SUBSCRIBERS" ] && export JIBRI_WEBHOOK_SUBSCRIBERS="http://localhost:$AUTOSCALER_SIDECAR_PORT/hook"
+        [ -z "$AUTOSCALER_SIDECAR_INSTANCE_ID" ] && export AUTOSCALER_SIDECAR_INSTANCE_ID="$JIBRI_INSTANCE_ID"
```
saghul, line 50: Does it make sense for the 2 ids to be different?

jibri/rootfs/defaults/autoscaler-sidecar.config
aaronkvanmeerten: It's only used internally by the jibri running in the same container.  It's default is set in the shell script

jibri/rootfs/defaults/autoscaler-sidecar.config
aaronkvanmeerten: No this is the type of instance we are running for the jibri container.  For other containers we will set it differently.

jibri/rootfs/defaults/autoscaler-sidecar.config
aaronkvanmeerten: Only for informational purposes to identify the container or node, can be filled in by nomad, kube etc as desired.

jibri/rootfs/defaults/autoscaler-sidecar.config
aaronkvanmeerten: Why yes, it is!

jibri.yml
aaronkvanmeerten: done

saghul (approved):

jvkassi:
how does autoscaler works ?

aaronkvanmeerten:
> how does autoscaler works ?

It's a work in progress, but the basic idea is that the sidecar sends stats to a microservice which then controls the number of instances and their various states (drain vs. in use).  The existing microservice currently only supports nomad deployments of containers, but could probably be extended to do kubernetes as well.  The source is here: https://github.com/jitsi/jitsi-autoscaler

PR #1564, opened 2023-06-14
base: migrate to Debian Bookworm

saghul:


saghul:
Not working yet because of https://github.com/adoptium/adoptium-support/issues/816

emrahcom:
Hi @saghul,

I tried the followings to install  `openjdk-11-jre-headless` on `bookworm` container and it work

- Run `docker.io/library/debian:bookworm-slim`
- Add `bullseye` repositories into `/ec/apt/sources.list.d/bullseye.list`

```
deb http://deb.debian.org/debian bullseye main
deb http://deb.debian.org/debian bullseye-updates main
deb http://security.debian.org/debian-security bullseye-security main
```

- Update repo info

```bash
apt-get update
```

- Install `openjdk-11-jre-headless`

```bash
apt-get install openjdk-11-jre-headless

# or

apt-get install openjdk-11-jre-headless --no-install-recommends
```

Notes:
- All dependencies come from `bookworm` repo except `libgraphite2-3`
- `apt-get dist-upgrade` doesn't install any new package and doesn't break the system
- `java` version after these steps

```bash
java --version

openjdk 11.0.18 2023-01-17
OpenJDK Runtime Environment (build 11.0.18+10-post-Debian-1deb11u1)
OpenJDK 64-Bit Server VM (build 11.0.18+10-post-Debian-1deb11u1, mixed mode, sharing)
```

saghul:
The AdopOpenJDK repo has been updated now but I haven't been able to update the PR as I was at a conference.

Any reason to prefer Debian old stable repo to that one?

The advantage I see with the latter is that they provide builds for all LTS Java versions, unlike Debian.

emrahcom:
Actually I don't have much idea about using the AdopOpenJDK repo. After a quick check, it looks like that `openjdk-11` will be available until `Oct 2024` on Adoptium but it will be available until `June 2026` on Debian Bullseye.

https://adoptium.net/support/
https://wiki.debian.org/LTS

saghul:
Interesting points! I'll dig a little here.

saghul:
re: the OpenJDK source: https://whichjdk.com/#adoptium-eclipse-temurin

We don't need to settle it in stone, but I think Temurin sounds like a good way forward. I assume we'll be running 17 before 11 is EOL anyway, so relying on Debian's support feels unnecessary.

emrahcom:
I think that a simpler PR will work now since the latest `jibri` works with `OpenJDK-17`.

saghul:
> I think that a simpler PR will work now since the latest `jibri` works with `OpenJDK-17`.

Yes indeed!

saghul:
I hope to pick back up early next year.

saghul:
Superseded by https://github.com/jitsi/docker-jitsi-meet/pull/1770

PR #1567, opened 2023-06-24
jicofo, prosody: allow to set different AUTH_TYPE

emrahcom:
This PR adds `PROSODY_AUTH_TYPE` and `JICOFO_AUTH_TYPE` environment variables to allow to set different authentication type for these two services.

As an example to a real use-case, it may be possible to select `JWT` for `prosody` while enabling the `internal` auth for `jicofo`.  This makes possible to use `wait for host` with `JWT`.

The old behaviour works as expected.


saghul:
Ha, that's a cool hack!

saghul (approved):

PR #1577, opened 2023-07-10
add support for max-ssrcs-per-user config

luocongqiu:


jicofo/rootfs/defaults/jicofo.conf
```diff
@@ -24,6 +24,8 @@
 {{ $XMPP_RECORDER_DOMAIN := .Env.XMPP_RECORDER_DOMAIN | default "recorder.meet.jitsi" -}}
 {{ $XMPP_PORT := .Env.XMPP_PORT | default "5222" -}}
 {{ $XMPP_SERVER := .Env.XMPP_SERVER | default "xmpp.meet.jitsi" -}}
+{{ $MAX-SSRCS-PER-USER := .Env.MAX-SSRCS-PER-USER | default "20" -}}
```
saghul, line 27: Please use the underscore _ instead of the hyphen - for the variable name.

saghul: Is 20 the JVB default?

jicofo/rootfs/defaults/jicofo.conf
```diff
@@ -24,6 +24,8 @@
...
+{{ $MAX-SSRC-GROUPS-PER-USER := .Env.MAX-SSRC-GROUPS-PER-USER | default $MAX-SSRCS-PER-USER -}}
```
saghul, line 28: What's the default in the JVB?

saghul (changes requested):

jicofo/rootfs/defaults/jicofo.conf
luocongqiu: @saghul    default value is 20 in jicofo

image

saghul (approved):

saghul (approved):

Issue #1447, opened 2022-11-24
Is there any plan to run container as non root user?


ghost:

saghul:
Yes, but it hasn't been a priority.

If you are willing to help, he first step would be to upgrade the s6 overlay to v3, which does allow running as non-root.

Once that is done, switching to non-root would be a next step.

bsobbe:
@saghul  is it gonna be a priority this year?

saghul:
I don't have any indication to believe it will be. As I mentioned earlier (https://github.com/jitsi/docker-jitsi-meet/issues/1447#issuecomment-1326094680) migrating to S6 v3 is a pre-requisite to that.

bsobbe:
@saghul  I have successfully migrated docker-jitsi-meet's base image to the latest version of s6-overlay (v3.1.5.0). After a few days of refactoring some of the cont-init and services.d and of course, some changes in Dockerfiles, seems like all the services are now running error-free with the new base.
I am going to do a few more tests during the week just to be sure before moving to add support for non-root users.
I will make sure to create a PR after it's ready, but meanwhile don't hesitate to let me know if there are any specific tests I need to perform.

saghul:
This is awesome, thank you!

bsobbe:
@saghul  I made the PR just now: https://github.com/jitsi/docker-jitsi-meet/pull/1584
Looking forward to getting some feedback.

neuhaus:
I'm hoping we can revive this important issue... @saghul 
There are tons of standards and benchmarks that mandate this practice:

- NIST SP 800-190 (Application Container Security Guide)
- CIS Benchmarks (Center for Internet Security)
- PCI DSS
- NSA and CISA Kubernetes Hardening Guidance
- ISO/IEC 27001 Annex A.9
- BSI IT-Grundschutz
- ENISA Guidelines
- ANSSI & SecNumCloud
- NIS2 Directive
- DORA
- NCSC Cloud Security Principles
- ...

The list goes on. The benefit will be improved security for everone running Jitsi in a container.

PR #1584, opened 2023-08-01
Adding support for non-root containers

bsobbe:
Steps:
1. Upgrading s6-overlay to v3.1.5.0 as suggested in [this issue](https://github.com/jitsi/docker-jitsi-meet/issues/1447) in base image.
2. Refactoring the scripts and making sure they have sufficient permission.
3. Changing ownerships and permissions of files and directories needed for the service to run.
4. Running a test call in a non-root k8s environment using services: web, jvb, jicofo, prosody.

Useful links during my process:
https://github.com/just-containers/s6-overlay/blob/master/MOVING-TO-V3.md
https://github.com/just-containers/s6-overlay/issues/539
https://github.com/just-containers/s6-overlay/issues/526


docker-compose.yml
```diff
@@ -4,6 +4,12 @@ services:
     # Frontend
     web:
         image: jitsi/web:${JITSI_IMAGE_VERSION:-unstable}
+        # build:
```
saghul, line 7: I suppose this is for testing?

jibri/Dockerfile
```diff
@@ -15,12 +15,37 @@ ARG USE_CHROMIUM=0
...
+COPY --chown=noroot:noroot rootfs/ /
 
 RUN apt-dpkg-wrap apt-get update && \
     apt-dpkg-wrap apt-get install -y -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" jibri libgl1-mesa-dri procps jitsi-upload-integrations jitsi-autoscaler-sidecar jq pulseaudio dbus dbus-x11 rtkit unzip fonts-noto && \
     /usr/bin/install-chrome.sh && \
-    apt-cleanup && \
-    adduser jibri rtkit
+    apt-cleanup
+
+RUN mkdir -p /etc/jitsi/autoscaler-sidecar
```
saghul, line 25: All of these RUN operations really need to be a single RUN.

jibri/Dockerfile
```diff
@@ -15,12 +15,37 @@ ARG USE_CHROMIUM=0
...
+RUN mkdir -p /config/logs
+RUN mkdir -p /config/recordings
+RUN mkdir -p /jibri
```
saghul, line 28: Where is this directory used?

jibri/Dockerfile
```diff
@@ -15,12 +15,37 @@ ARG USE_CHROMIUM=0
...
+
+RUN chown -R noroot:noroot /etc/cont-init.d
```
saghul, line 30: Many of these look generic enough to do it in the base image, any reason not to do it there?

jibri/rootfs/etc/services.d/10-xorg/run
```diff
@@ -1,5 +1,5 @@
-#!/usr/bin/with-contenv bash
+#!/command/with-contenv bash
 
 DAEMON="/usr/bin/Xorg -nocursor -noreset  +extension RANDR +extension RENDER -logfile /tmp/xorg.log  -config /etc/jitsi/jibri/xorg-video-dummy.conf ${DISPLAY}"
-exec s6-setuidgid jibri /bin/bash -c "exec $DAEMON"
+exec s6-setuidgid /bin/bash -c "exec $DAEMON"
```
saghul, line 4: Why do we still need s6-setuidgid if we are not specifying the user?

jibri/rootfs/etc/services.d/20-icewm/run
```diff
@@ -1,5 +1,5 @@
-#!/usr/bin/with-contenv bash
+#!/command/with-contenv bash
 
 DAEMON="/usr/bin/icewm-session"
-exec s6-setuidgid jibri /bin/bash -c "exec $DAEMON"
+exec s6-setuidgid /bin/bash -c "exec $DAEMON"
```
saghul, line 4: Ditto.

jibri/rootfs/etc/services.d/30-pulse/run
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/with-contenv bash
+#!/command/with-contenv bash
 
-HOME=/home/jibri
-exec s6-setuidgid jibri /bin/bash -c "exec /usr/bin/pulseaudio"
+HOME=/jibri
```
saghul, line 3: Ah, so here is where we use /jibri. Why is /home/jibri not good?

jibri/rootfs/etc/services.d/40-jibri/run
```diff
@@ -1,10 +1,10 @@
...
 
 # we have to set it, otherwise chrome won't find ~/.asoundrc file
-HOME=/home/jibri
+HOME=/jibri
 
 DAEMON=/opt/jitsi/jibri/launch.sh
 # pre-warm google chrome before jibri launches to ensure fast chrome launch during recordings
-s6-setuidgid jibri /usr/bin/google-chrome --timeout=1000 --headless about:blank
-exec s6-setuidgid jibri /bin/bash -c "exec $DAEMON"
+s6-setuidgid /usr/bin/google-chrome --timeout=1000 --headless about:blank
```
saghul, line 8: Same question s6-setuidgid

jicofo/Dockerfile
```diff
@@ -12,6 +12,22 @@ RUN apt-dpkg-wrap apt-get update && \
     apt-dpkg-wrap apt-get install -y jicofo && \
     apt-cleanup
 
-COPY rootfs/ /
+COPY --chown=noroot:noroot rootfs/ /
+
+RUN mkdir -p /config
```
saghul, line 17: Please combine all these RUNs.

saghul, line 17: Do we really need to try and create /config?

jicofo/Dockerfile
```diff
@@ -12,6 +12,22 @@ RUN apt-dpkg-wrap apt-get update && \
...
+RUN chown -R noroot:noroot /etc/cont-init.d
```
saghul, line 18: Most of these should go in the base image.

jicofo/rootfs/etc/services.d/jicofo/run
```diff
@@ -1,7 +1,7 @@
-#!/usr/bin/with-contenv bash
+#!/command/with-contenv bash
 
 JAVA_SYS_PROPS="-Djava.util.logging.config.file=/config/logging.properties -Dconfig.file=/config/jicofo.conf"
 DAEMON=/usr/share/jicofo/jicofo.sh
 DAEMON_DIR=/usr/share/jicofo/
 
-exec s6-setuidgid jicofo /bin/bash -c "cd $DAEMON_DIR; JAVA_SYS_PROPS=\"$JAVA_SYS_PROPS\" exec $DAEMON"
+exec /bin/bash -c "cd $DAEMON_DIR; JAVA_SYS_PROPS=\"$JAVA_SYS_PROPS\" exec $DAEMON"
```
saghul, line 7: So here we are not using s6-setuidgid... I guess that's the way to go, but then we should be consistent in all services.

jigasi/Dockerfile
```diff
@@ -14,6 +14,23 @@ RUN apt-dpkg-wrap apt-get update && \
     apt-dpkg-wrap apt-get install -y jigasi jq && \
     apt-cleanup
 
-COPY rootfs/ /
+COPY --chown=noroot:noroot rootfs/ /
+
+RUN mkdir /config
```
saghul, line 19: Ditto: combine in a single RUN

jigasi/Dockerfile
```diff
@@ -14,6 +14,23 @@ RUN apt-dpkg-wrap apt-get update && \
...
+RUN mkdir -pm777 /tmp/transcripts
```
saghul, line 20: Maybe we should change this to /transcripts or better yet /config/transcripts while we're at it.

jigasi/Dockerfile
```diff
@@ -14,6 +14,23 @@ RUN apt-dpkg-wrap apt-get update && \
...
+
+RUN chown -R noroot:noroot /etc/cont-init.d
+RUN chown -R noroot:noroot /etc/services.d
+RUN chown -R noroot:noroot /tmp
+RUN chown -R noroot:noroot /run
+RUN chown -R noroot:noroot /config
+RUN chown -R noroot:noroot /var
+RUN chown -R noroot:noroot /usr/share
```
saghul, line 28: Why do we need shuch a broad chown?

jigasi/rootfs/etc/services.d/jigasi/run
```diff
@@ -1,8 +1,8 @@
-#!/usr/bin/with-contenv bash
+#!/command/with-contenv bash
 
 JAVA_SYS_PROPS="-Djava.util.logging.config.file=/config/logging.properties"
 
 DAEMON=/usr/share/jigasi/jigasi.sh
 DAEMON_OPTS="--nocomponent=true --configdir=/ --configdirname=config --min-port=${JIGASI_PORT_MIN:-20000} --max-port=${JIGASI_PORT_MAX:-20050}"
 
-exec s6-setuidgid jigasi /bin/bash -c "JAVA_SYS_PROPS=\"$JAVA_SYS_PROPS\" exec $DAEMON $DAEMON_OPTS"
+exec s6-setuidgid /bin/bash -c "JAVA_SYS_PROPS=\"$JAVA_SYS_PROPS\" exec $DAEMON $DAEMON_OPTS"
```
saghul, line 8: ?

jvb/rootfs/etc/cont-init.d/10-config
```diff
@@ -32,14 +32,12 @@ export LOCAL_ADDRESS=$(ip route get 1 | grep -oP '(?<=src ).*' | awk '{ print $1
 export SENTRY_RELEASE="${SENTRY_RELEASE:-$(apt-cache policy jitsi-videobridge2 | sed -n '/Installed/p' | sed -e 's/[^:]*: //')}"
 
 if [[ -f /config/custom-sip-communicator.properties ]]; then
-    cat /config/custom-sip-communicator.properties > /config/sip-communicator.properties
+  cat /config/custom-sip-communicator.properties > /config/sip-communicator.properties;
```
saghul, line 35: Please leave 4 space indentation.

prosody/Dockerfile
```diff
@@ -61,11 +61,38 @@ RUN wget -qO /etc/apt/trusted.gpg.d/prosody.gpg https://prosody.im/files/prosody
...
+COPY --chown=noroot:noroot rootfs/ /
 
-COPY --from=builder /usr/local/lib/lua/5.4 /usr/local/lib/lua/5.4
-COPY --from=builder /usr/local/share/lua/5.4 /usr/local/share/lua/5.4
+COPY --chown=noroot:noroot --from=builder /usr/local/lib/lua/5.4 /usr/local/lib/lua/5.4
+COPY --chown=noroot:noroot --from=builder /usr/local/share/lua/5.4 /usr/local/share/lua/5.4
+
+RUN chown -R noroot:noroot /etc/cont-init.d
+RUN chown -R noroot:noroot /etc/services.d
+RUN touch /etc/saslauthd.conf && chown noroot:noroot /etc/saslauthd.conf
```
saghul, line 71: This is unnecessary, since the file needs to be templated with tpl.

prosody/rootfs/defaults/prosody.cfg.lua
```diff
@@ -215,7 +215,7 @@ http_interfaces = { "*", "::" }
 http_interfaces = { "*" }
 {{ end }}
 
-data_path = "/config/data"
+data_path = "/config/noroot_data"
```
saghul, line 218: Don't do this please. It will break existing installations for no good reason.

saghul (changes requested):
I made a first round of review, impressive work, thank you!

jibri/rootfs/etc/services.d/10-xorg/run
bsobbe: deleted

jibri/rootfs/etc/services.d/30-pulse/run
bsobbe: this is because we stopped creating jibri user, so initially this path would not exist. but now I'm creating /home/jibri and assigning the right permissions and therefore reverted this back to /home/jibri to minimize issues. PR will be updated soon..

jicofo/Dockerfile
bsobbe: Some of these mkdir statements used to be in the init script. considering the limited permissions of noroot, these directories need to be created with noroot as owner before that script in order to get over permission errors while running the container.

jicofo/rootfs/etc/services.d/jicofo/run
bsobbe: done

jigasi/Dockerfile
bsobbe: done. all paths have been updated to /config/transcripts

prosody/rootfs/defaults/prosody.cfg.lua
bsobbe: this is the only way I could get over permission issues while generating TLS certificates by prosody to /config/data.
it appears to me that something sets the permission of /config/data to root no matter what we set it in the Dockerfile. After creating noroot_data and setting the right ownerships, the SSL issue was solved and I was finally able to make a call. otherwise, JVB and Jicofo both failed while trying to reach prosody internally because it doesn't have TLS certificated.

If you have any ideas that can preserve the current setup and still solve the SSL issue with this user I am definitely open to hearing. but I'm curious to know why this will break the existing installations. since the /config/data is still there and the prosody PID is also still in /config/data. this is just changing the data directory to noroot_data to be able to generate SSLs. It appears to me that once the 10-config script is done it will even move the certificates to /config/certs and removes them from noroot_data at the end of the file.

saghul: What user does the 10-config script run as? A cursory look at https://github.com/bjc/prosody/blob/master/util/prosodyctl/cert.lua reveals some ownership changes if the user is root. Or maybe we should set `prosody_user` and `prosody_group` to the new user on the config file.

jicofo/Dockerfile
saghul: Then let's put that in the base image please.

bsobbe:
> I made a first round of review, impressive work, thank you!

Thanks for the detailed review. I made a few changes, cleaned up Dockerfiles, and tried to adapt to the requests as much as possible. feel free to check it out.

prosody/rootfs/defaults/prosody.cfg.lua
bsobbe: The user is noroot, with user and group ids = 1000. I bet that's why it can't create the certs in /config/data since apparently we can't change that one's ownership but it can create the certs in noroot_data and move them to the correct location since it's the owner. I could not find a better way to make this work, so I reverted it back to noroot_data so prosody doesn't break.

jibri/Dockerfile
```diff
@@ -15,12 +15,38 @@ ARG USE_CHROMIUM=0
...
-    apt-cleanup && \
-    adduser jibri rtkit
+    apt-cleanup
+
+RUN mkdir -p /etc/jitsi/autoscaler-sidecar && \
+    mkdir -p /config/logs && \
+    mkdir -p /config/recordings && \
+    mkdir -p /home/jibri && chown -R noroot:noroot /home/jibri
+
+RUN chown -R noroot:noroot /etc/cont-init.d && \
```
saghul, line 30: These 2 are already done in the base image. Is it necessary?

jibri/Dockerfile
```diff
@@ -15,12 +15,38 @@ ARG USE_CHROMIUM=0
...
+    chown -R noroot:noroot /etc/services.d && \
+    chown -R noroot:noroot /etc/jitsi && \
```
saghul, line 32: Why is this necessary? Even if files are owned by root we can read them, right?

jibri/Dockerfile
```diff
@@ -15,12 +15,38 @@ ARG USE_CHROMIUM=0
...
+    chown -R noroot:noroot /run && \
+    chown -R noroot:noroot /config && \
+    chown -R noroot:noroot /var && \
+    chown -R noroot:noroot /opt/jitsi && \
+    chown -R noroot:noroot /tmp && \
+    chown -R noroot:noroot /defaults && \
+    chown -R noroot:noroot /usr/bin
+
+RUN chmod +x /etc/cont-init.d/* && \
+    chmod +x /etc/services.d/10-xorg/* && \
```
saghul, line 42: Any chance we can use wildcards to use a single command here?

jicofo/Dockerfile
```diff
@@ -12,6 +12,21 @@ RUN apt-dpkg-wrap apt-get update && \
     apt-dpkg-wrap apt-get install -y jicofo && \
     apt-cleanup
 
-COPY rootfs/ /
+COPY --chown=noroot:noroot rootfs/ /
+
+RUN chown -R noroot:noroot /etc/cont-init.d && \
```
saghul, line 17: Same, is this necessary? If it is, which I'd like you to double-check, let's a script in the base image which all mages call, instead of duplicating the code on each of them.

jicofo/Dockerfile
```diff
@@ -12,6 +12,21 @@ RUN apt-dpkg-wrap apt-get update && \
...
+    chown -R noroot:noroot /etc/services.d && \
+    chown -R noroot:noroot /run && \
+    chown -R noroot:noroot /var && \
+    chown -R noroot:noroot /usr/share/jicofo && \
+    chown -R noroot:noroot /etc/jitsi && \
+    chown -R noroot:noroot /etc/jitsi/jicofo && \
+    mkdir -p /config && chown -R noroot:noroot /config
```
saghul, line 24: This should go on the base image.

prosody/Dockerfile
```diff
@@ -61,11 +61,36 @@ RUN wget -qO /etc/apt/trusted.gpg.d/prosody.gpg https://prosody.im/files/prosody
...
 
-COPY rootfs/ /
+COPY --chown=noroot:noroot rootfs/ /
 
-COPY --from=builder /usr/local/lib/lua/5.4 /usr/local/lib/lua/5.4
-COPY --from=builder /usr/local/share/lua/5.4 /usr/local/share/lua/5.4
+COPY --chown=noroot:noroot --from=builder /usr/local/lib/lua/5.4 /usr/local/lib/lua/5.4
+COPY --chown=noroot:noroot --from=builder /usr/local/share/lua/5.4 /usr/local/share/lua/5.4
+
+RUN mkdir -pm777 /var/run/saslauthd && \
```
saghul, line 69: Please use a single RUN block, here and elsewhere.

prosody/rootfs/defaults/prosody.cfg.lua
saghul: But we can configure the prosody user and group with the options I showed above, did you test that? Why can noroot create stuff in /config/data tough?

We need to fix this before this PR can land.

saghul (changes requested):
Good progress! Some some comments. A couple of generic things:

- Use a single RUN
- Make the permission fixes geenric with a script if needed so we don't keep repeating that

prosody/rootfs/defaults/prosody.cfg.lua
bsobbe: @saghul  I tried the following:

1.  Changing everything back to `/config/data`. 
2. Setting the correct ownership of `noroot:noroot` on `/config` recursively.
3. Setting `prosody_user` and `prosody_group` to 'noroot' in the prosody.cfg.lua as the following: 
  ```
  prosody_user = "noroot"
  prosody_group = "noroot"
```
it still can't write in the /config/data due to the following error:
`The directory /config/data is not owned by the current user, won't be able to write files to it`

Seems like the error is coming from this line of the file you sent before:
https://github.com/bjc/prosody/blob/b5d791c8bffcaa3e200876cb159ff92de2ba1c05/util/prosodyctl/cert.lua#L282

which is specifically checking for the user to be uid 0 (aka root) or own the directory. It is not possible in cloud setups that do not allow running any container as root, so we definitely hit the first part of that `if`. and for the second part, changing the ownership of /config/data does not work either since it reverts back to root:daemon somehow..

Personally, I could not find any solution that works apart from using a backup directory like /config/noroot_data for certs. But I'm open to more suggestions.

saghul: Since it uses `~=` that line checks if the user is NOT root and is NOT the owner of the certs directory, doesn't it? Maybe try creating the certs directory inside data?

bsobbe: @saghul yes precisely. our `uid = 1000` and the owner is `root:daemon`. therefore we hit that if statement and get the error: `The directory `/config/data `is not owned by the current user, won't be able to write files to it`. 

I don't see why putting` /config/certs` directory inside` /config/data `would help solve this problem. I feel like there is a misunderstanding about the nature of the issue. Ownership over /config/certs is **not** the issue here. Ownership over `/config/data` is the actual problem.
Prosody expects noroot to be the owner of that directory. But, regardless of changing that ownership to noroot in the dockerfile, something always overwrites its ownership to `root:daemon`. My guess is it has something to do with `prosody.pid` being there (not actually sure if it can be because of this). 

The bottom line is, if we can't use noroot_data as a temp path for generating certificates, we need to find a way to change ownership of `/config/data`. Otherwise, it will keep trying to write the certs in that one and floods the container log with failed to load TLS certificates errors.

saghul: > yes precisely. our `uid = 1000` and the owner is `root:daemon`. therefore we hit that if statement and get the error: `The directory `/config/data `is not owned by the current user, won't be able to write files to it`.

Oh hold on. Why can't we change the permissions of that?



> But, regardless of changing that ownership to noroot in the dockerfile, something always overwrites its ownership to `root:daemon`. My guess is it has something to do with `prosody.pid` being there (not actually sure if it can be because of this).

Aha. Maybe try with chattr -i ?

bsobbe: @saghul did not work. tried this:
// to install chattr cause it did not exist by default.
RUN apt-get install -y e2fsprogs
// Create /config/data (cause it does not exist at this stage) - change ownership - change attribute and remove immutability.
RUN mkdir -p /config/data && chown -R noroot:noroot /config/data && chattr -R -i /config/data

all of that resulted in:
│ The directory /config/data is not owned by the current user, won't be able to write files to it                            │
│ The directory /config/data is not owned by the current user, won't be able to write files to it                            │
│ mv: cannot stat '/config/data/*.crt': No such file or directory                                                            │
│ mv: cannot stat '/config/data/*.key': No such file or directory

same old same old.

saghul: Hum, that's very odd, as if the user running `prosodyctl` is root or something. Can you please take care of the rest of the items and leave this one for last? I'll try to take a look too.

saghul:
I've just had an idea about the certificate generation @bsobbe . We could stop using `prosodyctl` to generate them and use the `openssl` command line instead.

saghul:
@bsobbe do you intend to pick this back up?

PR #1585, opened 2023-08-03
prosody: add hybrid_matrix_token as a new authentication method

emrahcom:
`Prosody` uses only one authentication method at a time in a single setup. Therefore one setup cannot be used by participant with a token and by a Matrix member (_through Element Jitsi widget_) at the same time even both methods are based on `JWT`.

This pull-request allows `Jitsi` to be used by both groups at the same by adding [auth_hybrid_matrix_token](https://github.com/jitsi-contrib/prosody-plugins/tree/main/auth_hybrid_matrix_token) as a new authentication method for `Prosody`.


saghul (approved):

saghul:
Cheers!

PR #1606, opened 2023-09-21
JVB configurable Jetty sendServerVersion

viakom-choppe:
Added 2 Env new variables which allow to configure, if the jetty http server should include the jetty version in its HTTP Headers and error pages.

Default behaviour is defined in jicoco "JettyBundleActivatorConfig.kt" and is also set to true.


saghul:
@damencho @bgrozev is there a reason to ever send the version?

bgrozev:
> @damencho @bgrozev is there a reason to ever send the version?

Not that I can think of. I think we were just preserving behavior when we introduced the flag, but it makes sense to me to change the default to `false`.
https://github.com/jitsi/jicoco/pull/189

viakom-choppe:
@saghul Done.
I was not sure if I should already squash the commits or keep both. So I kept them for now.

saghul (approved):

bgrozev:
Do we still need this after the default upstream was changed to `false`?

saghul:
Just making it more explicit since it hasn't reached stable yet.

PR #1611, opened 2023-09-27
prosody: visitor mode support

aaronkvanmeerten:


prosody/rootfs/defaults/conf.d/visitors.cfg.lua
```diff
@@ -0,0 +1,191 @@
...
+    ssl = {
+        key = "/config/certs/v{{ $VISITOR_INDEX }}.{{ $XMPP_DOMAIN }}.key";
+        certificate = "/config/certs/v{{ $VISITOR_INDEX }}.{{ $XMPP_DOMAIN }}.crt";
+    }
+    modules_enabled = {
+      'bosh';
+      'ping';
+      "external_services";
+      "smacks";
+      "jiconop";
```
saghul, line 123: Do we still need this?

prosody/rootfs/defaults/conf.d/visitors.cfg.lua
```diff
@@ -0,0 +1,191 @@
...
+    }
+    main_muc = 'conference.v{{ $VISITOR_INDEX }}.{{ $XMPP_DOMAIN }}';
```
saghul, line 125: Why are we hardcoding the MUC name? The base config doesn't.

prosody/rootfs/defaults/prosody.cfg.lua
```diff
@@ -143,6 +165,42 @@ c2s_interfaces = { "*", "::" }
 c2s_interfaces = { "*" }
 {{ end }}
 
+{{ if $ENABLE_S2S -}}
+-- set s2s port
+s2s_ports = { {{ $S2S_PORT }} } -- Listen on specific s2s port
+
+s2s_whitelist = {
+	{{ if $ENABLE_VISITORS -}}
+    'conference.{{ $XMPP_DOMAIN }}'; -- needed for visitors to send messages to main room
```
saghul, line 174: This should probably use the variable or won't work, will it?

prosody/rootfs/etc/cont-init.d/10-config
```diff
@@ -25,10 +25,24 @@ if [[ "$(stat -c %U /prosody-plugins-custom)" != "prosody" ]]; then
...
+    echo "Prosody visitor mode, using alternate config"
+    PROSODY_SITE_CFG="visitors.cfg.lua"
+    rm /config/conf.d/jitsi-meet.cfg.lua
+    # force jicofo into auth domain for visitor-mode prosody
+    [ -z "$XMPP_AUTH_DOMAIN" ] && XMPP_AUTH_DOMAIN="auth.meet.jitsi"
+    export PROSODY_ADMINS="focus@$XMPP_AUTH_DOMAIN"
+else
+    echo "Prosody normal mode, using default config"
+    PROSODY_SITE_CFG="jitsi-meet.cfg.lua"
+    rm /config/conf.d/visitors.cfg.lua
```
saghul, line 42: Would it make sense to delete all *.lua from there at the start?

saghul (commented):
Left some comments, PTAL.

prosody/rootfs/etc/cont-init.d/10-config
aaronkvanmeerten: I was pondering that, but then considered that someone might want to mount other files in there to provide custom configuration, so I didn't want to assume anything there.

prosody/rootfs/defaults/conf.d/visitors.cfg.lua
aaronkvanmeerten: Yes, this is the mod_jiconop module, not the deprecated node service unfortunately named the same thing.

prosody/rootfs/defaults/conf.d/visitors.cfg.lua
aaronkvanmeerten: Good point, I'll work on this next.

prosody/rootfs/defaults/prosody.cfg.lua
aaronkvanmeerten: Excellent point, good catch.

prosody/rootfs/etc/cont-init.d/10-config
saghul: Good point!

saghul (approved):

PR #1629, opened 2023-10-23
Add sip-jibri container

emrahcom:
This PR adds `sip-jibri` container.

Even it is working, there are still some parts that need improvement. Please take this as a starting point. For now:

- `pjsua` executable is downloaded from [here](https://emrah.com/files/pjsua_2.10-jitsi3). I will update this when it will be available as a package.
- not tested on `arm64`
- `alsa` is used instead of `pulse`
- depends on the host config because of `v4l2loopback` and `alsa`
- no document for now for the new environment variables and the host config in the Jitsi handbook


saghul:
> * `pjsua` executable is downloaded from [here](https://emrah.com/files/pjsua_2.10-jitsi3). I will update this when it will be available as a package.

It should be possible to use an intemediate image to build it and then copy it to the final image. See how the prosody image uses an intermediate one to build some libraries. I'm not sure we intend to package it.

> * `alsa` is used instead of `pulse`
> 
> * depends on the host config because of `v4l2loopback` and `alsa`

Is it possible to use PulseAudio's userland loopback capabilities? I'm not sure if there is a `v4l2loopback` replacement that we could use :-/

emrahcom:
> > * `pjsua` executable is downloaded from [here](https://emrah.com/files/pjsua_2.10-jitsi3). I will update this when it will be available as a package.
> 
> It should be possible to use an intemediate image to build it and then copy it to the final image. See how the prosody image uses an intermediate one to build some libraries. I'm not sure we intend to package it.

I will check this and update PR.

> > * `alsa` is used instead of `pulse`
> > * depends on the host config because of `v4l2loopback` and `alsa`
> 
> Is it possible to use PulseAudio's userland loopback capabilities? I'm not sure if there is a `v4l2loopback` replacement that we could use :-/

I didn't find a replacement for `v4l2loopback` and since the container is already dependent on host's config, I didn't try hardly to replace `alsa`. Looks like `pulseaudio` makes things more complicated unlike the standard `jibri` which has only need one loopback device.

Maybe `pipewire` will be a better alternative for the future.

emrahcom:
@saghul,

I added an intermediate image to build `pjsua` and it works in my tests

emrahcom:
@saghul what do you think about the current status of PR?

saghul:
That's quite some work, thanks @emrahcom ! I'll take a look.

PR #1635, opened 2023-11-02
bug: escape/encapsulate string in env.example

HannesOberreiter:
Currently if you use the default `source` command in linux and the default `.env` file as declared in the `env.example`, it will fail because of the space in the string.

Using double-quotes around the string will solve this issue.


saghul (approved):

saghul:
Cheers!

Issue #1647, opened 2023-11-25
allow jvb max memory value to be altered


gottaeat:
users who deploy via docker are unable to alter the default value of 3072 in a permanent matter as the `VIDEOBRIDGE_MAX_MEMORY` env var (https://github.com/jitsi/jitsi-videobridge/blob/814bffd6cf050001a020288e70e942fa0d505d64/jvb/resources/jvb.sh#L19) is not defined within the docker compose or the `env.example`

edit:
doing a little more digging, although not expressed anywhere within the docker related jitsi files, simply exporting 
- JICOFO_MAX_MEMORY=N
- VIDEOBRIDGE_MAX_MEMORY=N
in the compose allows the max mem to be altered.

gottaeat:
```diff
compose yaml
--- a/docker-compose.yml  2023-09-22 08:44:45.000000000 +0000
+++ b/docker-compose.yml  2023-11-25 22:14:05.479998766 +0000
@@ -326,6 +326,7 @@
             - XMPP_RECORDER_DOMAIN
             - XMPP_SERVER
             - XMPP_PORT
+            - JICOFO_MAX_MEMORY
         depends_on:
             - prosody
         networks:
@@ -374,6 +375,7 @@
             - XMPP_INTERNAL_MUC_DOMAIN
             - XMPP_SERVER
             - XMPP_PORT
+            - VIDEOBRIDGE_MAX_MEMORY
         depends_on:
             - prosody
         networks:
```

```sh
# .env
JICOFO_MAX_MEMORY=1536m
VIDEOBRIDGE_MAX_MEMORY=1536m
```

saghul:
A PR would be welcome!

PR #1648, opened 2023-11-26
allow jvb and jicofo max_mem env vars to be altered through docker

gottaeat:
as described in https://github.com/jitsi/docker-jitsi-meet/issues/1647, for non containerized setups of jitsi, users can alter the maximum memory limit for java to be set in the respective jvb and jicofo configuration files, but these files are not exposed in `/config` for dockerized installations.

while that is the case, the run-scripts for both jvb and jicofo read `JICOFO_MAX_MEMORY` and `VIDEOBRIDGE_MAX_MEMORY` from the environment.

these patches allow for the same functionality to be restored without the config files needing to be exposed to the main namespace.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -338,6 +338,7 @@ services:
             - JICOFO_ENABLE_HEALTH_CHECKS
             - JICOFO_ENABLE_REST
             - JICOFO_HEALTH_CHECKS_USE_PRESENCE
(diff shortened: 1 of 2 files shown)


saghul (approved):

saghul:
Sorry, I forgot to push the button.

gottaeat:
cheers.

PR #1660, opened 2023-12-04
Allow adding custom modules to guest muc domain

jmallach:


damencho:
What modules do you want to add there?

jmallach:
I'm trying to populate the `participants` metadata in my setup (using JWT/OIDC auth), so I am following this solution that is doing the job for me https://community.jitsi.org/t/jibri-metadata-json/20254/20. I'd like to restrict this module to the guest virtual domain, though, as currently I need to add it to `GLOBAL_MODULES`.

PR #1663, opened 2023-12-06
Add transcriber authentication to hidden domain

zobadaniel:
It took me a while to understand what kind of configuration is required to have jigasi's transcriber join a conference as hidden participant. This allows normal participants to set up transcripts for themselves after the moderator has enabled transcripts (and thus invited transcriber to the meeting).

It seems that this configuration has been partly implemented already, e.g. some of the environment variables were already there. The commits just complete the config for jigasi and prosody, and add a variable for enabling/disabling.

It is confusing though that the authentication has to be for the same domain as the recorder, because there can be only one hidden domain in the frontend - and I did not want one to exclude the other. So now to use this feature, recording needs to be configured as well. I have no idea how to streamline this. Nevertheless I wanted to publish this, since the desired feature can be set up just by config.


env.example
```diff
@@ -116,6 +116,10 @@ ETHERPAD_SKIN_VARIANTS="super-light-toolbar super-light-editor light-background
 # Enable guest access
 #ENABLE_GUESTS=1
 
+# Enable auth for Transcriber in Jigasi
+# This can only be enabled if the other auth options above are not used
+#JIGASI_TRANSCRIBER_AUTHENTICATION=1
```
saghul, line 121: This is very confusing. The transcriber should be authenticated always, like a recorder. Can this be dropped?

jigasi/rootfs/etc/cont-init.d/10-config
```diff
@@ -51,3 +51,10 @@ if [[ $ENABLE_TRANSCRIPTIONS -eq 1 || $ENABLE_TRANSCRIPTIONS == "true" ]]; then
         }' \
         > /config/key.json
 fi
+
+# check that no conflicting authentications are configured
+if [[ $ENABLE_AUTH -eq 1 && $JIGASI_TRANSCRIBER_AUTHENTICATION -eq 1 ]]; then
```
saghul, line 56: Yeah nobody will understand this. Let's make it work authenticated always, like a recorder please.

prosody/rootfs/etc/cont-init.d/10-config
```diff
@@ -104,6 +105,15 @@ if [[ ! -z $JIGASI_XMPP_PASSWORD ]]; then
     prosodyctl --config $PROSODY_CFG register $JIGASI_XMPP_USER $XMPP_AUTH_DOMAIN $JIGASI_XMPP_PASSWORD
 fi
 
+if [[ ! -z $JIGASI_TRANSCRIBER_PASSWORD ]]; then
+    OLD_JIGASI_TRANSCRIBER_PASSWORD=passw0rd
```
saghul, line 109: This never existed, so we don't need to check for it.

saghul (changes requested):
Left some commentas, please take a look!

env.example
zobadaniel: I guess that eventually, this can be dropped and the authentication of the transcriber can be made mandatory. I don't know whether this would be a "breaking change" though? Maybe it was never meant that the transcriber works without authentication, but I can't tell. Please see also my generic comment below.

jigasi/rootfs/etc/cont-init.d/10-config
zobadaniel: Agree, not actually intuitive.

prosody/rootfs/etc/cont-init.d/10-config
zobadaniel: Honestly I just copied the handling of other paswords. I don't know why this password would have to be handled specially. Or is this some sort of legacy stuff that is simply no longer necessary for newly introduced passwords?

saghul: The latter.

zobadaniel:
Currently, the recorder is the only authenticated user that utilizes the "hidden domain" - and there is only one "hidden domain". 

If we want to add the transcriber, we might have to give up on the idea that the hidden domain is defined and used solely by the recorder, and managed by the "ENABLE_RECORDING" option.

Is this reasonable? I might try and unbundle this, and have an explicit option for setting up the "hidden domain" which both the recorder and the transcriber need to authenticate to.

saghul:
> If we want to add the transcriber, we might have to give up on the idea that the hidden domain is defined and used solely by the recorder, and managed by the "ENABLE_RECORDING" option.

Certainly!

> Is this reasonable? I might try and unbundle this, and have an explicit option for setting up the "hidden domain" which both the recorder and the transcriber need to authenticate to.

Sounds like a good plan!

surajprakashchoudhary:
Hi @saghul and @zobadaniel 

Can you please tell us when it will be available in production? Can we expect to get this feature in the next Docker-Jitsi-Meet stable release?

Does this pull request (PR) solve both problems:

- Hide transcriber frames for all participants.
- Enable or show the translation to both the moderator and other participants.

zobadaniel:
@surajprakashchoudhary This PR is intended to ease/allow the configuration of the transcriber as "hidden participant" (e.g. it will not be visible the same way as a "normal" participant without video). And if a moderator then enables subtitles, the subtitle menu will be visible to all participants (and be removed once the last person switches off subtitles again). 

During the discussion with saghul I learned that this PR shall be reworked to make the config better comprehensible. This means reworking a part of the meeting recorder config as well.

I can't make any promises to when I will have a revised version ready for discussion. Nevertheless I wanted to post my first version, to get feedback on how to set up this feature properly. It took me a while to collect the info and consolidate into this PR.

saghul:
Thanks for your work @zobadaniel !

@surajprakashchoudhary With the holiday season approaching, we cannot make any promises. Everyone deserves a rest :-)

saghul:
@zobadaniel Do you intend to pick this back up?

zobadaniel:
Yes, but it might take a while until I find the time to do so. I can't promise to have this done short term due to other stuff with unfortunately higher priority.

saghul:
No worries, just checking in!

zobadaniel:
Closing this one as https://github.com/jitsi/docker-jitsi-meet/pull/1737 now covers a more generic attempt.

PR #1674, opened 2023-12-19
Introduce DESKTOP_SHARING_FRAMERATE_AUTO to allow browser to determine sharing fps automatically

### LINKED ISSUE #1668: Screen share starts with a low resolution, screens are unreadable for ~30 seconds
Jitsi version: stable-9111

Since the last two releases, when a user starts a screen share, the resolution of the share usually starts very low. When starting the screen share, it starts with a resolution of 512x288 (totally unreadable). Then after a few seconds, it upgrades itself to 1027x576 (better, but still pretty hard to read text) and finally it goes to 1536x864 (perfect).

The problem eventually corrects itself, but it takes a good 30 seconds for the screen to become perfectly readable for people watching. This is quite a long time, and other tools (like Teams) start with a decent screen share resolution immediately.

Is there any way to configure Jitsi to start with a specific screen share resolution, possibly in a percentage of the screen/window size that is being shared, as this can vary on the size of the window or the resolution of the full screen being shared. Ideally, you would want the share to start at 80% or maybe even 100% of the original screen size. There really is no point to start lower, since a screen share has a different function than a camera view: lowering the resolution makes no sense. While a lower camera resolution is still usable, a lower screen share resolution is not.

**Update**: I've noticed that if you shared screen in a call, then stopped sharing screen but remain in the call, and then later start sharing your screen again, it immediately starts at the higher resolution. As if it remembers that you have shared at this resolution before, and then immediately starts with it. Now only to have this same feature at the first time to share screen would be perfect.

cedricroijakkers:
This PR Fixes #1668

The docker containers allow the option to set the screen sharing fps via two environment variables `DESKTOP_SHARING_FRAMERATE_MIN` and `DESKTOP_SHARING_FRAMERATE_MAX`. This is fine for people who want to tweak these settings, but on meet.jit.si, these options are left to the defaults. These are that min and max framerate are not configured at all, and that they are therefore left to the browser to determine.

This PR introduces a new environment variable `DESKTOP_SHARING_FRAMERATE_AUTO` which when set to `true`, which is the default, does not configure the desktop sharing parameters in the `web` container at all, i.e. leaving them up to the browser to determine (which is also the behaviour on meet.jit.si). For those who still want to configure their own framerate, they can set `DESKTOP_SHARING_FRAMERATE_AUTO` to `false` and configure `DESKTOP_SHARING_FRAMERATE_MIN` and `DESKTOP_SHARING_FRAMERATE_MAX` as they wish.


saghul (approved):

saghul:
Thank you!

PR #1675, opened 2023-12-19
task: support logging to file in JVB container

aaronkvanmeerten:


bgrozev (approved):

saghul (approved):
Why would we want this rather than do that redirection at the Docker runtime level?

aaronkvanmeerten:
> Why would we want this rather than do that redirection at the Docker runtime level?

When running in-container jvb-rtcstats-push which also scrapes a logfile for inclusion to the server.

saghul:
Gotcha!

PR #1679, opened 2023-12-19
jvb: expose public IP in autoscaler sidecar config

aaronkvanmeerten:


saghul (approved):
Note: it's a comma separated list.

aaronkvanmeerten:
> Note: it's a comma separated list.

It's intended to be informational, to tell one JVB from another in the autoscaler reports, so CSV is OK

PR #1685, opened 2024-01-02
web: trim deprecated options and defaults

aaronkvanmeerten:


jallamsetty1 (approved):

web/rootfs/defaults/settings-config.js
```diff
@@ -344,11 +346,12 @@ config.roomPasswordNumberOfDigits = {{ $ROOM_PASSWORD_DIGITS }};
 // Advanced.
 //
 
-// Lipsync hack in jicofo, may not be safe.
-config.enableLipSync = {{ $ENABLE_LIPSYNC }};
-
-config.enableRemb = {{ $ENABLE_REMB }};
-config.enableTcc = {{ $ENABLE_TCC }};
+{{ if not $ENABLE_REMB -}}
```
saghul, line 349: Why go the verbose route? Setting the value always since it has a sane default is a good idea IMHO.

aaronkvanmeerten: @jallamsetty1 and I discussed, and since the code itself has sane defaults, it was felt that an additional default wasn't needed here.  Open to reverting that pieces as I don't feel strongly on it.

saghul: Right now most settings are set by taking defaults in the template itself.

Generally speaking we have avoided changing defaults drastically, because it leads to problems.

jallamsetty1: Yes, the current practice is to set the defaults in config.js but for settings that affect the core media functionality, I don't think we should make them configurable anymore. I do not see a reason why we should let users disable REMB, TCC, RTX or simulcast for that matter. IMHO not adding the defaults to the template is a first step in that direction.

saghul: Why do we have a setting outside of the testing section for them then?

Not arguing, just curious.

jallamsetty1: I guess they were added when the features were implemented but we didn't bother to remove them after the default behavior was implemented in code. That happens with a lot of features that we implement. We conditionally enable them, then enable them by default and don't remove the config.js flag.

saghul: I see.

PR #1686, opened 2024-01-02
fix: Fix default JIBRI_PENDING_TIMEOUT.

bgrozev:


aaronkvanmeerten (approved):

saghul:
Pl prefix the container in the commit message going forward, "fix: Fix xyz " vs "jibri: fix xyz"

PR #1693, opened 2024-01-02
prosody: enable muc_meeting_id in lobby component

aaronkvanmeerten:


prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -406,6 +406,7 @@ Component "lobby.{{ $XMPP_DOMAIN }}" "muc"
     muc_room_locking = false
     muc_room_default_public_jids = true
     modules_enabled = {
+        "muc_meeting_id";
```
bgrozev, line 409: Does the file always exist?

aaronkvanmeerten (commented):
> Does the file always exist? 

yes, it's in the open source jitsi-meet prosody modules that get copied into the container every time.

bgrozev (approved):

PR #1707, opened 2024-01-12
Default ENABLE_OCTO_SCTP to ENABLE_SCTP not .Env.ENABLE_SCTP.

garloff:
The latter may not be passed and then the tpl initializtion for jicofo.conf fails, resulting in an empty conf file and a disfunctional setup.

### DIFF
--- a/jicofo/rootfs/defaults/jicofo.conf
+++ b/jicofo/rootfs/defaults/jicofo.conf
@@ -7,7 +7,7 @@
 {{ $ENABLE_SCTP := .Env.ENABLE_SCTP | default "0" | toBool -}}
 {{ $ENABLE_RECORDING := .Env.ENABLE_RECORDING | default "0" | toBool -}}
 {{ $ENABLE_OCTO := .Env.ENABLE_OCTO | default "0" | toBool -}}
-{{ $ENABLE_OCTO_SCTP := .Env.ENABLE_OCTO_SCTP | default .Env.ENABLE_SCTP | toBool -}}
+{{ $ENABLE_OCTO_SCTP := .Env.ENABLE_OCTO_SCTP | default $ENABLE_SCTP | toBool -}}
 {{ $ENABLE_AUTO_LOGIN := .Env.ENABLE_AUTO_LOGIN | default "1" | toBool -}}
 {{ $ENABLE_REST := .Env.JICOFO_ENABLE_REST | default "0" | toBool -}}
 {{ $ENABLE_JVB_XMPP_SERVER := .Env.ENABLE_JVB_XMPP_SERVER | default "0" | toBool -}}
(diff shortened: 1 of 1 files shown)


jicofo/rootfs/defaults/jicofo.conf
```diff
@@ -7,7 +7,7 @@
 {{ $ENABLE_SCTP := .Env.ENABLE_SCTP | default "0" | toBool -}}
 {{ $ENABLE_RECORDING := .Env.ENABLE_RECORDING | default "0" | toBool -}}
 {{ $ENABLE_OCTO := .Env.ENABLE_OCTO | default "0" | toBool -}}
-{{ $ENABLE_OCTO_SCTP := .Env.ENABLE_OCTO_SCTP | default .Env.ENABLE_SCTP | toBool -}}
+{{ $ENABLE_OCTO_SCTP := .Env.ENABLE_OCTO_SCTP | default $ENABLE_SCTP | toBool -}}
```
aaronkvanmeerten, line 10: This would work, except that we need to use the toBool which only accepts a string. I had it this way to begin with but it fails

aaronkvanmeerten (changes requested):
I do not have the exact solution here but unfortunately this is not a working one

aaronkvanmeerten:
There may be something to try with parentheses and defaults. I will try to work on this in an hour or two unless you get time before me.

jicofo/rootfs/defaults/jicofo.conf
garloff:
```
{{ $ENABLE_SCTP_TMP := .Env.ENABLE_SCTP | default "0" }}
{{ $ENABLE_SCTP := $ENABLE_SCTP_TMP | toBool -}}
{{ $ENABLE_OCTO_SCTP := .Env.ENABLE_OCTO_SCTP | default $ENABLE_SCTP_TMP | toBool -}}
```
Not really elegant, but works.

garloff:
> There may be something to try with parentheses and defaults. I will try to work on this in an hour or two unless you get time before me.

I pushed a fix to the branch. Maybe you find a more elegant solution without a temporary variable.

jicofo/rootfs/defaults/jicofo.conf
saghul: Hum, I suppose we could fix tpl to do it? https://github.com/jitsi/tpl/blob/0a1fc48a5bcb8cbd88db78bb59b8f76818058af4/func.go#L45

aaronkvanmeerten:
Here's my attempt:

https://github.com/jitsi/docker-jitsi-meet/pull/1708

Let me know if that works for you?  I tested by making a candidate template file in a running container and then running tpl on it with various env vars exported, and it seems to do what we want.

garloff:
Superceded  by #1708

PR #1720, opened 2024-01-22
prosody: c2s encryption flag from environment

aaronkvanmeerten:
allows c2s_require_encrption flag in meet config to be controlled environment


bgrozev (approved):

saghul:
This changes the default from false to true. Was that intentional?

PR #1726, opened 2024-01-30
Remove PROSODY_ADMINS

bgrozev:
It is overriden in jitsi-meet.cfg.lua and therefore only usable in
visitors mode. Move the declaration to visitors.cfg.lua


saghul (approved):

saghul:
Si we need this?

damencho:
O nice, few days ago I was complaining to @aaronkvanmeerten that the admins declaration is a mess and we do it twice in some cases because of the main prosody cfg and the inclusion

damencho:
I need to check, but I think this is the right way l.

bgrozev:
Testing some more, this may not be safe. Code running between the two declarations will see the first value. I don't know if there's any code which depends on this or not :(

Issue #1599, opened 2023-09-11
Video Quality Impaired - The bridge Channel has been disconnect 


iarlemozaga:
Hello, I'm using Jitsi Docker Version 8922

When its only 2 people connected it pops up everytime the message: 

Video Quality Impaired - The bridge Channel has been disconnect 

And when there are more than 2 people, the connection it pops up the same alert and everyone lost the video, screenshare and audio.

Does someone knows how to fix that? I already editted my .env added my domain, I'm also forwarding the ports

I've noticed a warning in the logs 

**docker-jitsi-meet-stable-8922-jvb-1:** AllocationSettingsWrapper.setBandwidthAllocationSettings#174: Ignoring assumed-bandwidth-bps, not allowed in config. 

Does someone had the same issue? I already tried to find something but no success

saghul:
The problem is caused by the users not being able to reach the JVB. Did you set JVB_ADVERTISED_IPS to a correct value?

iarlemozaga:
I'm running into a VPS with public ip, I did set the JVB_ADVERTISED_IPS to the public ip, should I set to a different one?

saghul:
That should be ok. Is port 10000 filtered somehow?

iarlemozaga:
I'm using behind Nginx Proxy Manager, I'm forwarding the 8000 and using SSL with nginx, the ports are openned also in the Firewall, 10000 its openned too

iarlemozaga:
Port 10000 Its already openned

saghul:
Then we need the browser console logs to further diagnose.

iarlemozaga:
Error: 2023-09-12T18:02:08.554Z [modules/RTC/BridgeChannel.js] <e.onclose>:  Channel closed: 1001 endpoint closed

Warning: MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 rtc.forwarded_sources_changed listeners added. Use emitter.setMaxListeners() to increase limit

saghul:
Check the network tab to see the WS connection attempt.

iarlemozaga:
Is that tab?

![image](https://github.com/jitsi/docker-jitsi-meet/assets/105409779/3e60dd6a-4322-4762-88cf-205dd5736fc3)

iarlemozaga:
![image](https://github.com/jitsi/docker-jitsi-meet/assets/105409779/86db6d9b-35f1-4ade-ba47-30c2ce14515e)

[...]

Issue #1531, opened 2023-04-13
websocket connection failed


bp2468:
websocket connection xmpp-websocket failed.

saghul:
It's impossible to help you without any logs.

Are you running behind a reverse proxy? Did you make sure WebSocket connections are proxied?

prathibhacdac:
The above issue is resolved by adding ENABLE_HTTP_REDIRECT=1 in .env
 But now I am getting the following error: Video is not getting displayed
WebSocket connection to 'wss://bp.erss.in/colibri-ws/172.28.0.5/3f45edae982d3a91/082e38a9?pwd=7kivrblerg80eh5a7gq5va2e53' failed:

prathibhacdac:
I'm getting the error message on the web page as:

"The bridge channel has been disconnected and thus video quality is limited to its lowest setting."

prathibhacdac:
How to resolve this?

saghul:
Start by checking the logs.

prathibhacdac:
It is not taking the port 9090. 

wss://bp.erss.in/colibri-ws
It should be wss://bp.erss.in:9090/colibri-ws
How to resolve this?

prathibhacdac:
which log should I check?

saghul:
The JVB container logs.

kunimasu:
I faced to same message "The bridge channel has been disconnected and thus video quality is limited to its lowest setting.".

<img width="348" alt="Screenshot 2023-09-08 at 22 37 04" src="https://github.com/jitsi/docker-jitsi-meet/assets/23184567/3bcf137d-083d-4491-8acb-0aea1f4af4ad">

My usecase:
- Jitsi server is Raspberry pi 4 model B 8GB.
- OS: Linux raspberrypi 6.1.21-v8+ #1642 SMP PREEMPT Mon Apr  3 17:24:16 BST 2023 aarch64 GNU/Linux
- Jitsi works by self hosted with Docker container (jitsi/jvb, jicofo, prosody, web stable-8719)
- When connected 2 clients of Chrome browser(Version 116.0.5845.140 (Official Build) (x86_64)) is no problem. I mean video and audio can be communicated.
- When connected 3 clients, that message shows me and stop video and audio.

[...]

PR #1739, opened 2024-02-09
add jvb dependency to web container

BifbofII:
since web runs dig jvb on startup, if jvb isn't up yet, this can fail or point to a wrong ip, leading the colibri websocket connection to fail

this might be the cause for #1531 and #1599, at least I fixed the same symptoms like this for me


saghul:
Good catch!

saghul (approved):

PR #1742, opened 2024-02-16
base-java: install nodejs 20 from nodesource

aaronkvanmeerten:


aaronkvanmeerten (commented):
It's failed twice on jibri builds, due to something wrong with chromedriver, but it doesn't appear to be related to the changes here.

jibri/Dockerfile
```diff
@@ -11,7 +11,7 @@ LABEL org.opencontainers.image.documentation="https://jitsi.github.io/handbook/"
 ARG USE_CHROMIUM=0
 #ARG CHROME_RELEASE=latest
 # https://googlechromelabs.github.io/chrome-for-testing/
-ARG CHROME_RELEASE=121.0.6167.85
+ARG CHROME_RELEASE=121.0.6167.184
```
saghul, line 14: Can you try backing out this change?

aaronkvanmeerten: Yes, but this was my attempt to fix the problem. It did not solve it.

aaronkvanmeerten: No good I believe this is a separate problem that we need to solve

saghul: Not sure, I built it like that 2 days ago no problem.

I see the problem seems to be the zip file layout, but it would be very odd that they changed the layout for a given version.

saghul: Never mind, the nightly build fails the same way: https://github.com/jitsi/docker-jitsi-meet/actions/runs/7939771467

Please drop that change and I'll look into fixing it.

aaronkvanmeerten: It's already dropped.

saghul (approved):

PR #1744, opened 2024-02-18
Use storage.googleapis.com location for chromedriver.

garloff:
This is the place listed on
https://googlechromelabs.github.io/chrome-for-testing/ which would seem to be the least likely to silently break. Currently, the edgedl.me.gvt1.com location is broken.


saghul:
Thank you!

saghul (approved):

PR #1746, opened 2024-02-21
Fix prosody vars

denzs:
If XMPP_AUTH_DOMAIN is not set in Environment it leads to this rendered config inside of the prosody container:

```
    muc_password_whitelist = {
        "focus@<no value>"
    }
```

This PR fixes it by using the default if not overriden by Env.


saghul (approved):
Thank you!

PR #1760, opened 2024-03-18
jigasi: skip SIP configuration when configured as a transcriber

saghul:


aaronkvanmeerten (approved):

saghul:
Is there a world in which we deem running a Jigasi for both dialing and transcriptions in the same container is ok?

PR #1761, opened 2024-03-19
fix(web) :bug: wrong quote in default config

HannesOberreiter:
Quotes in the config for subdomain were wrong.

### DIFF
--- a/web/rootfs/defaults/system-config.js
+++ b/web/rootfs/defaults/system-config.js
@@ -22,7 +22,7 @@ config.focusUserJid = 'focus@{{$XMPP_AUTH_DOMAIN}}';
 
 {{ if $ENABLE_SUBDOMAINS -}}
 var subdir = '<!--# echo var="subdir" default="" -->';
-var subdomain = "<!--# echo var="subdomain" default="" -->";
+var subdomain = '<!--# echo var="subdomain" default="" -->';
 if (subdir.startsWith('<!--')) {
     subdir = '';
 }


web/rootfs/defaults/system-config.js
```diff
@@ -22,7 +22,7 @@ config.focusUserJid = 'focus@{{$XMPP_AUTH_DOMAIN}}';
 
 {{ if $ENABLE_SUBDOMAINS -}}
 var subdir = '<!--# echo var="subdir" default="" -->';
-var subdomain = "<!--# echo var="subdomain" default="" -->";
+var subdomain = '<!--# echo var="subdomain" default="" -->';
```
aaronkvanmeerten, line 25: I'm curious what impact you saw from this?  

The nginx container shoudl be replacing the full contents from  using SSI directives before any javascript is parsed by the browser.  So there should not be any point at which the quotes confliect.

HannesOberreiter: Currently setting up a dev environment to make some changes on the jitsi meet server code and was struggling to load the config and though maybe its this, but just found out it was something different. So it is no bug only style and it is now the same as in the jitsi-meet repo:

https://github.com/jitsi/jitsi-meet/blob/master/config.js

aaronkvanmeerten: Excellent, I'm all for consistency but I just wanted to ensure I understood the impact of the existing code.

aaronkvanmeerten (approved):

PR #1767, opened 2024-03-21
feat(prosody): Cleans up gc settings.

damencho:
Reverts 69bba17 and some parts of 2798426.


saghul:
Why are we removing the ability to customize that?

bgrozev:
> Why are we removing the ability to customize that?

We're not, these are already defined in `prosody.cfg.lua`

bgrozev (approved):

damencho:
I have missed that the main prosody one that has the settings is always used and the general config and we just include all configs in conf.d, where for standalone we have a separate general config for visitors.

PR #1772, opened 2024-03-25
Add access-control-allow-origin env variable in web meet.conf

simon-kaiser:
Added cors header variable in meet.conf so that access-control-allow-origin header can be set by a dynamic value / domain and replaces the default *.


saghul (approved):
LGTM, thank you! @aaronkvanmeerten thoughts?

aaronkvanmeerten (approved):
lgtm

Issue #1771, opened 2024-03-24
JWT Auth - RS512 unsupported, because cannot possibly supply private key in the JWT_APP_SECRET variable


DevelopingEntitiesWithFuntations:
In JWT Auth environment variables there's no alternative to provide a private key file or to add it directly to the JWT_APP_SECRET variable, because it will complain about the first "+" in the private key.

I wanted to implement RS512 JWT checking, but realized it's not possible to do that with Docker at least, because the Lua script will simply fail at one of those characters.

It would be nice to mention that in the documentation with the supported algs.

DevelopingEntitiesWithFuntations:
Weird because here they mention RS256 for example: https://github.com/jitsi/lib-jitsi-meet/blob/master/doc/tokens.md

saghul:
What images version are you using and how does your .env file look like?

DevelopingEntitiesWithFuntations:
@saghul Thank you for your prompt reply. I am running `stable-9364` and the `.env` looks like this (only the essential stuff so that I don't flood with unnecessary entries):
```
# Enable authentication
ENABLE_AUTH=1

# Enable guest access
ENABLE_GUESTS=1

# Select authentication type: internal, jwt, ldap or matrix
AUTH_TYPE=jwt

# JWT authentication
#

# Application identifier
JWT_APP_ID=myjitsiid

# Application secret known only to your token generator
JWT_APP_SECRET=-----BEGIN RSA PRIVATE KEY-----
keycontentshere
-----END RSA PRIVATE KEY-----
```

Perhaps I just need to dearmor it, but I haven't seen any functionality in the code to actually support RS512 like that as mentioned above

damencho:
To use RS512 you need to setup: JWT_ASAP_KEYSERVER
https://github.com/jitsi/docker-jitsi-meet/blob/03b6ca24f21d1eaae7b4e58d2f15ca9af514b346/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua#L151

```
For the Public Key Validation a key server must be provided via asap_key_server .... 
```

DevelopingEntitiesWithFuntations:
@damencho ahh the documentation states or looks as if the two variables `JWT_APP_ID` and `JWT_APP_SECRET` were required, but blog posts and similar seem to state that the one is optional in that case...

damencho:
Yep it is secret or asap server.

damencho:
[...]

PR #1776, opened 2024-03-30
JWT Sign Type Fixing (#1771)

DevelopingEntitiesWithFuntations:
Added and exposed the required env variable.

### DIFF
--- a/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
+++ b/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
@@ -13,7 +13,7 @@
 {{ $JWT_ALLOW_EMPTY := .Env.JWT_ALLOW_EMPTY | default "0" | toBool -}}
 {{ $JWT_AUTH_TYPE := .Env.JWT_AUTH_TYPE | default "token" -}}
 {{ $JWT_ENABLE_DOMAIN_VERIFICATION := .Env.JWT_ENABLE_DOMAIN_VERIFICATION | default "false" | toBool -}}
-{{ $JWT_SIGN_TYPE := .Env.JWT_SIGN_TYPE | default "null" -}}
+{{ $JWT_SIGN_TYPE := .Env.JWT_SIGN_TYPE | default "RS256" -}}
 {{ $MATRIX_UVS_ISSUER := .Env.MATRIX_UVS_ISSUER | default "issuer" -}}
 {{ $MATRIX_UVS_SYNC_POWER_LEVELS := .Env.MATRIX_UVS_SYNC_POWER_LEVELS | default "0" | toBool -}}
 {{ $MATRIX_LOBBY_BYPASS := .Env.MATRIX_LOBBY_BYPASS | default "0" | toBool -}}
--- a/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
+++ b/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
@@ -136,7 +136,7 @@ VirtualHost "jigasi.meet.jitsi"
     authentication = "token"
     {{ if .Env.JWT_SIGN_TYPE }}
        signature_algorithm = "{{ .Env.JWT_SIGN_TYPE }}"
-       {{ end }}
+   {{ end -}}
     app_id = "jitsi";
     
     asap_key_server = "https://jaas-public-keys.jitsi.net/jitsi-components/prod-8x8"
--- a/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
+++ b/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
@@ -13,7 +13,7 @@
 {{ $JWT_ALLOW_EMPTY := .Env.JWT_ALLOW_EMPTY | default "0" | toBool -}}
 {{ $JWT_AUTH_TYPE := .Env.JWT_AUTH_TYPE | default "token" -}}
 {{ $JWT_ENABLE_DOMAIN_VERIFICATION := .Env.JWT_ENABLE_DOMAIN_VERIFICATION | default "false" | toBool -}}
-{{ $JWT_SIGN_TYPE := .Env.JWT_SIGN_TYPE | default "RS256" -}}
+{{ $JWT_SIGN_TYPE := .Env.JWT_SIGN_TYPE -}}
 {{ $MATRIX_UVS_ISSUER := .Env.MATRIX_UVS_ISSUER | default "issuer" -}}
 {{ $MATRIX_UVS_SYNC_POWER_LEVELS := .Env.MATRIX_UVS_SYNC_POWER_LEVELS | default "0" | toBool -}}
 {{ $MATRIX_LOBBY_BYPASS := .Env.MATRIX_LOBBY_BYPASS | default "0" | toBool -}}
@@ -135,7 +135,8 @@ VirtualHost "jigasi.meet.jitsi"
       "muc_password_check";
     }
     authentication = "token"
-    signature_algorithm = "{{ $JWT_SIGN_TYPE }}"
+    {{ if .Env.JWT_SIGN_TYPE }}
+       signature_algorithm = "{{ $JWT_SIGN_TYPE }}"
     app_id = "jitsi";
     
     asap_key_server = "https://jaas-public-keys.jitsi.net/jitsi-components/prod-8x8"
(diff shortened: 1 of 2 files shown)


prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -134,7 +135,9 @@ VirtualHost "jigasi.meet.jitsi"
       "muc_password_check";
     }
     authentication = "token"
+    signature_algorithm = "{{ .Env.JWT_SIGN_TYPE }}"
```
saghul, line 138: Use the variable you have declared, not .Env

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -13,6 +13,7 @@
 {{ $JWT_ALLOW_EMPTY := .Env.JWT_ALLOW_EMPTY | default "0" | toBool -}}
 {{ $JWT_AUTH_TYPE := .Env.JWT_AUTH_TYPE | default "token" -}}
 {{ $JWT_ENABLE_DOMAIN_VERIFICATION := .Env.JWT_ENABLE_DOMAIN_VERIFICATION | default "false" | toBool -}}
+{{ $JWT_SIGN_TYPE := .Env.JWT_SIGN_TYPE | default "RS512" -}}
```
saghul, line 16: This will break existing setups. Default to null so the module picks the right default, you can still choose RS512 is you want to.

saghul (changes requested):

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
DevelopingEntitiesWithFuntations: Thanks, fixed it!

saghul (commented):
I'm not sure "null" as a string will work, we need to check that.

DevelopingEntitiesWithFuntations:
Ok, please do.

DevelopingEntitiesWithFuntations:
Is possible to run a CI/CD pipeline to test that?

DevelopingEntitiesWithFuntations:
@saghul what if we take `nil` instead of `null`? 🤔

DevelopingEntitiesWithFuntations:
@saghul then remove the quotes?...

Alternatively, perhaps error out if no value is set otherwise continue?

saghul:
Just add a condition and set it to whatever value the env var has, if it's set.

DevelopingEntitiesWithFuntations:
Let me get this straight - so I need to open up another PR on the Jitsi Meet GitHub to fix this and after that is merged you may merge this PR afterwards?

saghul:
No, you can fix it right here. Just do (pseudocode): if .Env.FOO then the_setting = {{ .Env.FOO }}

So don't set it unconditionally, thus preserving the default value, if the user didn't set it.

DevelopingEntitiesWithFuntations:
Don't if conditions require unnecessary computation power and should be avoided for such cases?

I think I did what you wanted - `{{ $JWT_SIGN_TYPE := .Env.JWT_SIGN_TYPE | default "RS256" -}}`

DevelopingEntitiesWithFuntations:
Or you want smth like:

```
{{- if .Env.JWT_SIGN_TYPE }}
{{ $JWT_SIGN_TYPE := .Env.JWT_SIGN_TYPE }}

```

DevelopingEntitiesWithFuntations:
or you want an or case with a default value?

saghul:
No. DONT declare the auxiliary variable now, you can just if the declaration of the actual prosody variable.

DevelopingEntitiesWithFuntations:
I don't quite get you. Like this on Line 138 (my version)?:
```
{{ if .Env.JWT_SIGN_TYPE }}
    signature_algorithm = "{{ $JWT_SIGN_TYPE }}"
else
    signature_algorithm = "RS256"
```

saghul:
Almost. Don't set signature_algorithm at all if the env car is not defined.

DevelopingEntitiesWithFuntations:
So just the first two lines from the example above then?:
```
{{ if .Env.JWT_SIGN_TYPE }}
    signature_algorithm = "{{ $JWT_SIGN_TYPE }}"
```

saghul:
Yes!

DevelopingEntitiesWithFuntations:
Please review this and merge if it's correct ^^

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -134,7 +135,10 @@ VirtualHost "jigasi.meet.jitsi"
       "muc_password_check";
     }
     authentication = "token"
+    {{ if .Env.JWT_SIGN_TYPE }}
+       signature_algorithm = "{{ $JWT_SIGN_TYPE }}"
```
saghul, line 139: You need the endif

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -134,7 +135,10 @@ VirtualHost "jigasi.meet.jitsi"
       "muc_password_check";
     }
     authentication = "token"
+    {{ if .Env.JWT_SIGN_TYPE }}
```
saghul, line 138: Use the variable

saghul (changes requested):

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -13,6 +13,7 @@
 {{ $JWT_ALLOW_EMPTY := .Env.JWT_ALLOW_EMPTY | default "0" | toBool -}}
 {{ $JWT_AUTH_TYPE := .Env.JWT_AUTH_TYPE | default "token" -}}
 {{ $JWT_ENABLE_DOMAIN_VERIFICATION := .Env.JWT_ENABLE_DOMAIN_VERIFICATION | default "false" | toBool -}}
+{{ $JWT_SIGN_TYPE := .Env.JWT_SIGN_TYPE -}}
```
saghul, line 16:
```suggestion
```

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -134,7 +134,11 @@ VirtualHost "jigasi.meet.jitsi"
       "muc_password_check";
     }
     authentication = "token"
+    {{ if .Env.JWT_SIGN_TYPE }}
+       signature_algorithm = "{{ .Env.JWT_SIGN_TYPE }}"
+       {{ end }}
```
saghul, line 139:
```suggestion
   {{ end -}}
```

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -134,7 +134,11 @@ VirtualHost "jigasi.meet.jitsi"
       "muc_password_check";
     }
     authentication = "token"
+    {{ if .Env.JWT_SIGN_TYPE }}
+       signature_algorithm = "{{ .Env.JWT_SIGN_TYPE }}"
+   {{ end -}}
```
saghul, line 139:
```suggestion
     {{ end -}}
```

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -134,7 +134,11 @@ VirtualHost "jigasi.meet.jitsi"
       "muc_password_check";
     }
     authentication = "token"
+    {{ if .Env.JWT_SIGN_TYPE }}
+       signature_algorithm = "{{ .Env.JWT_SIGN_TYPE }}"
+     {{ end -}}
```
saghul, line 139:
```suggestion
    {{ end -}}
```

prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua
```diff
@@ -134,7 +134,11 @@ VirtualHost "jigasi.meet.jitsi"
       "muc_password_check";
     }
     authentication = "token"
+    {{ if .Env.JWT_SIGN_TYPE }}
+       signature_algorithm = "{{ .Env.JWT_SIGN_TYPE }}"
+    {{ end -}}
     app_id = "jitsi";
+
```
saghul, line 141:
```suggestion
```

PR #1777, opened 2024-04-02
feat(prosody): Switches lua gc back to incremental.

damencho:


saghul (approved):
Pl adjust the commit message to match when merging

PR #1792, opened 2024-04-25
Update acme.sh Version to 3.0.7

zhenguang:
Upgrade acme.sh version to 3.0.7

Having ACME error on cert creation when using the older version, upgrade to 3.0.7 official as of 25 April 2024

[3.0.7](https://github.com/acmesh-official/acme.sh/releases/tag/3.0.7)

### DIFF
--- a/web/Dockerfile
+++ b/web/Dockerfile
@@ -8,7 +8,7 @@ LABEL org.opencontainers.image.url="https://jitsi.org/jitsi-meet/"
 LABEL org.opencontainers.image.source="https://github.com/jitsi/docker-jitsi-meet"
 LABEL org.opencontainers.image.documentation="https://jitsi.github.io/handbook/"
 
-ADD https://raw.githubusercontent.com/acmesh-official/acme.sh/3.0.1/acme.sh /opt
+ADD https://raw.githubusercontent.com/acmesh-official/acme.sh/3.0.7/acme.sh /opt
 COPY rootfs/ /
 
 RUN apt-dpkg-wrap apt-get update && \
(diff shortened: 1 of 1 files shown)


saghul:
Are no other changes required?

zhenguang:
Tested on docker version running with no issue, didn't test on renewal etc, i suppose this would be the same since the script is to run acme.sh

saghul:
Why 3.0.1 and not 3.0.7?

zhenguang:
My intention was to update it to 3.0.7 (the latest version which was released as of today), i was testing on 3.0.1 to ensure that when i upgrade from 2.8.8 to 3.0.1 it's working fine, before proceeding to it's latest version 3.0.7

I've update the PR code to 3.0.7

saghul:
Excellent!

saghul (approved):

saghul:
I'll give it a try before merging.

saghul:
So I tried it and it did not refresh the existing cert properly :-/ Perhaps it has something to do with switching LE engines, in version 2 Let's Encrypt was the default engine, and in version 3 it's ZeroSSL.

I think we should have a parameter controlling that default.

Deleting ~/.jitsi-meet-cfg/web/acme* and restarting made the setup work. Not sure this is an acceptable solution though.

zhenguang:
If you start a server anew, the whole docker will not load up because of the acme failure

saghul:
Not sure what you mean, can you clarify? Is that with the old or the new version?

zhenguang:
With the older version

saghul (approved):

jplandry908:
Thanks @zhenguang / @saghul. This should resolve the issue I created last week. My conclusion was also that acme.sh needed to be upgraded. I'll do some testing and close out that issue once verified.

https://github.com/jitsi/docker-jitsi-meet/issues/1795

jplandry908:
FYI - I confirmed that this resolved https://github.com/jitsi/docker-jitsi-meet/issues/1795.

saghul:
Wonderful!

PR #1480, opened 2023-02-22
whiteboard support

tgrld:

PR #1794, opened 2024-04-25
whiteboard.yml: add collaborative whiteboard addon

loli10K:
This is just an updated version of #1480 since the original author doesn't seem to be active anymore on GitHub.

 There are still some things to be done:

- [x] open a PR to the handbook repository (https://github.com/jitsi/handbook/pull/504)
- [ ] publish an official `jitsi/excalidraw-backend` docker image
- [ ] understand why moving the excalidraw location to a different path doesn't seem to work (https://github.com/jitsi/docker-jitsi-meet/pull/1480#issuecomment-2016579181)

### DIFF
--- a/web/rootfs/defaults/meet.conf
+++ b/web/rootfs/defaults/meet.conf
@@ -134,6 +134,21 @@ location ^~ /etherpad/ {
 }
 {{ end }}
 
+{{ if .Env.WHITEBOARD_COLLAB_SERVER_URL_BASE }}
+# whiteboard (excalidraw-backend)
+location = /socket.io/ {
+    proxy_buffering off;
+    proxy_cache_bypass $http_upgrade;
+
+    proxy_http_version 1.1;
+    proxy_set_header Upgrade $http_upgrade;
+    proxy_set_header Connection "upgrade";
+    proxy_set_header X-Forwarded-For $remote_addr;
+
+    proxy_pass {{ .Env.WHITEBOARD_COLLAB_SERVER_URL_BASE }}/socket.io/?$args;
+}
+{{ end }}
+
 location ~ ^/([^/?&:'"]+)$ {
     try_files $uri @root_path;
 }
--- a/web/rootfs/defaults/settings-config.js
+++ b/web/rootfs/defaults/settings-config.js
@@ -62,8 +62,7 @@
 {{ $ENABLE_LOCAL_RECORDING_SELF_START := .Env.ENABLE_LOCAL_RECORDING_SELF_START | default "false" | toBool -}}
 {{ $DISABLE_PROFILE := .Env.DISABLE_PROFILE | default "false" | toBool -}}
 {{ $ROOM_PASSWORD_DIGITS := .Env.ROOM_PASSWORD_DIGITS | default "false" -}}
-{{ $WHITEBOARD_COLLAB_SERVER_PUBLIC_URL := .Env.WHITEBOARD_COLLAB_SERVER_PUBLIC_URL | default "" -}}
-{{ $WHITEBOARD_ENABLED := .Env.WHITEBOARD_ENABLED | default "false" | toBool -}}
+{{ $WHITEBOARD_ENABLED := or (.Env.WHITEBOARD_COLLAB_SERVER_PUBLIC_URL | default "" | toBool) (.Env.WHITEBOARD_COLLAB_SERVER_URL_BASE | default "" | toBool) }}
 {{ $TESTING_AV1_SUPPORT := .Env.TESTING_AV1_SUPPORT | default "false" | toBool -}}
 
 // Video configuration.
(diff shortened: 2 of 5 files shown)


web/rootfs/defaults/settings-config.js
```diff
@@ -554,7 +554,11 @@ config.e2eping.maxMessagePerSecond = {{ .Env.E2EPING_MAX_MESSAGE_PER_SECOND }};
 // Settings for the Excalidraw whiteboard integration.
 config.whiteboard = {
     enabled: {{ $WHITEBOARD_ENABLED }},
-    collabServerBaseUrl: '{{ $WHITEBOARD_COLLAB_SERVER_PUBLIC_URL }}'
+{{ if .Env.WHITEBOARD_COLLAB_SERVER_PUBLIC_URL -}}
+    config.whiteboard.collabServerBaseUrl = '{{ $WHITEBOARD_COLLAB_SERVER_PUBLIC_URL }}';
```
saghul, line 558: This syntax is incorrect, needs to be likle the other URL.

web/rootfs/defaults/meet.conf
```diff
@@ -134,6 +134,20 @@ location ^~ /etherpad/ {
 }
 {{ end }}
 
+{{ if $WHITEBOARD_ENABLED }}
+# whiteboard (excalidraw-backend)
+location = /socket.io/ {
```
saghul, line 139: Let's make this /excalidraw/socket-io

web/rootfs/defaults/settings-config.js
```diff
@@ -554,7 +554,11 @@ config.e2eping.maxMessagePerSecond = {{ .Env.E2EPING_MAX_MESSAGE_PER_SECOND }};
...
+{{ else }}
+    collabServerBaseUrl: '{{ $PUBLIC_URL }}'
```
saghul, line 560: adjust this to the suggested URL change.

saghul (changes requested):

web/rootfs/defaults/settings-config.js
loli10K: Which other URL? Isn't this the same syntax used for Etherpad?

https://github.com/jitsi/docker-jitsi-meet/blob/09bbfaf01cc400d85d21006ccb33ac40da21366b/web/rootfs/defaults/settings-config.js#L143-L147

saghul: Not the same, here we are declaring an object literal.

loli10K: Sorry, i'm not following you: i just tried to run this locally and the syntax in this file seem to be correct, there was an unrelated issue in another file (`meet.conf`) but with that fixed the whiteboard is running fine on my end.

Was this something that was requested and then addressed in the original PR and i am now reverting it accidentally in this new PR?

saghul: The case that doesn't work is the previous one, not the one you are adding.

Looks perhaps like a merge error. Just make sure the collab... property is assigned the same way not with config.foo.bar =. While that works in other scenarios, it doesn't in an object literal as is the case here.

loli10K: Understood, this is due to things being moved around in https://github.com/jitsi/docker-jitsi-meet/commit/1dbe5cda68038403b417caf641ce476f0fa7d571#diff-8fc682fc9b342e54e76c53511199952847429c949bca05b186ae167f883f18d0L480, should be fixed now.

web/rootfs/defaults/meet.conf
saghul: That is because the URL in the setting needs to end in /excalidraw/

loli10K: Which settings exactly? I changed both `meet.conf` and `settings-config.js`, the latter being the file i thought was responsible for the Excalidraw javascript frontend making HTTP requests?

saghul: In both.

loli10K: I did: it doesn't work. Please read my full comment here https://github.com/jitsi/docker-jitsi-meet/pull/1480#issuecomment-2016579181

> However as soon as i move the excalidraw location to a different path as requested in the code review things start to break:
```diff
diff --git a/web/rootfs/defaults/meet.conf b/web/rootfs/defaults/meet.conf
index 986e9da..6ade033 100644
--- a/web/rootfs/defaults/meet.conf
+++ b/web/rootfs/defaults/meet.conf
@@ -136,7 +136,7 @@ location ^~ /etherpad/ {
 
 {{ if $WHITEBOARD_ENABLED }}
 # whiteboard (excalidraw-backend)
-location = /socket.io/ {
+location = /excalidraw-collab/socket.io/ {
     tcp_nodelay on;
 
     proxy_http_version 1.1;
diff --git a/web/rootfs/defaults/settings-config.js b/web/rootfs/defaults/settings-config.js
index 9fae33a..a828a2d 100644
--- a/web/rootfs/defaults/settings-config.js
+++ b/web/rootfs/defaults/settings-config.js
@@ -557,7 +557,7 @@ config.whiteboard = {
 {{ if .Env.WHITEBOARD_COLLAB_SERVER_PUBLIC_URL -}}
     config.whiteboard.collabServerBaseUrl = '{{ $WHITEBOARD_COLLAB_SERVER_PUBLIC_URL }}';
 {{ else }}
-    collabServerBaseUrl: '{{ $PUBLIC_URL }}'
+    collabServerBaseUrl: '{{ $PUBLIC_URL }}/excalidraw-collab/'
 {{ end }}
 };

```
> For some reason my browser keeps trying to connect to $PUBLIC_URL/socket.io instead of $PUBLIC_URL/excalidraw-collab/socket.io.

saghul: Ok, we can look into that later then.

web/rootfs/defaults/meet.conf
```diff
@@ -134,6 +135,20 @@ location ^~ /etherpad/ {
...
+    tcp_nodelay on;
+
+    proxy_http_version 1.1;
+    proxy_set_header Upgrade $http_upgrade;
+    proxy_set_header Connection "upgrade";
+    proxy_set_header Host $http_host;
+
+    proxy_pass http://whiteboard.meet.jitsi/socket.io/?$args;
```
saghul, line 148: Note to self: In Etherpad we control this with ETHERPAD_URL_BASE, we shuld try to be consistent, but at the same time I'm not sure happy with how the Etherpad settings are done...

saghul (approved):

saghul:
@loli10K I pushed this over the finish line. I'll land it once the CI turns 💚  Sorry for the delay.

Issue #1797, opened 2024-05-01
LDAP Auth: Filtering for group membership + Bind user


le-firehawk:
Hello,
I have recently been attempting to set up LDAP authentication for my Jitsi Meet instance with mild success.
I am running the stable branch for all docker images:
![image](https://github.com/jitsi/docker-jitsi-meet/assets/70988152/cb402b5e-311e-4c23-b561-13c85e3c3c5f)

I am connecting to a FreeIPA instance (v4.11.1)

The configuration I have confirmed as working is the following (extracted from .env)
`LDAP_URL=ldap://ipa01.my.domain
LDAP_BASE=CN=accounts,DC=my,DC=domain
LDAP_FILTER='(uid=%u)'
LDAP_VERSION=3
LDAP_USE_TLS=1`

In the above case, all other LDAP related values are commented out, and authentication works globally.
I however want to only allow login from users of group 'fs-jitsi'. I have attempted this by doing the following:

- Setting `LDAP_BASE` to `CN=fs-jitsi,CN=accounts,DC=my,DC=domain`
- Setting `LDAP_FILTER` to `LDAP_FILTER='(&(memberOf=CN=fs-jitsi,CN=groups,CN=accounts,DC=my,DC=domain)(uid=%u))'` <- This was mimicking the functional LDAP filter I use in Nextcloud

Perhaps I was missing something in the docs, or this feature is not supported via docker env?

Error Message:
![image](https://github.com/jitsi/docker-jitsi-meet/assets/70988152/cd66861d-1f18-4af2-8988-9526cdc48f94)

I get a similar message when defining a bind user for LDAP lookup, which makes me think I may be using an incorrect schema value somewhere
`LDAP_BINDDN=UID=jitsigroup,CN=users,CN=accounts,DC=my,DC=domain
LDAP_BINDPW=password`

This is my first time posting an issue, so please let me know if there's any additional info I need to provide.

aaronkvanmeerten:
This looks more like an LDAP issue than a docker-jitsi-meet issue.  Can you try also posting this question on the community forums?  https://community.jitsi.org/ has a lot of helpful folks.

damencho:
Try logging in with test-user@my.domain. not just the username, but add and the correct domain.

le-firehawk:
> This looks more like an LDAP issue than a docker-jitsi-meet issue. Can you try also posting this question on the community forums? https://community.jitsi.org/ has a lot of helpful folks.

[...]

PR #1798, opened 2024-05-03
prosody/rootfs/defaults/saslauthd.conf: Append extended LDAP variables

le-firehawk:
Include variables from saslauthd documentation [here](https://github.com/winlibs/cyrus-sasl/blob/d933c030ce12ec0668469d79ab8378e347a1b3ba/saslauthd/LDAP_SASLAUTHD) to achieve basic LDAP group membership filtering.

Changes + necessary syntax for FreeIPA compatibility included in env.example.

Tested on stable release, with saslauthd v2.1.28. LDAP tested against FreeIPA v4.11.1. See https://github.com/jitsi/docker-jitsi-meet/issues/1797 for more details.

Peripheral changes:
- LDAP_START_TLS check was outside PROSODY_AUTH_TYPE eq "LDAP" check. Relocate within
- Adjust indentation of {{ if }} and {{ end }} statements to reflect their nested depth
- Cluster variables by function (LDAP bind authentication, LDAP filtering, etc)


env.example
```diff
@@ -143,14 +143,18 @@ ETHERPAD_SKIN_VARIANTS="super-light-toolbar super-light-editor light-background
 # LDAP authentication (for more information see the Cyrus SASL saslauthd.conf man page)
 #
 
-# LDAP url for connection
-#LDAP_URL=ldaps://ldap.domain.com/
+# LDAP url:port for connection
```
duburcqa, line 146: as far as I know specifying the port as no effect. Did you check ?

env.example
```diff
@@ -143,14 +143,18 @@ ETHERPAD_SKIN_VARIANTS="super-light-toolbar super-light-editor light-background
...
+#LDAP_URL=ldaps://ldap.domain.com:636
```
duburcqa, line 147: `ldaps:` can only be specified if `ldap_start_tls` is not. Otherwise, `ldap` must be used instead.

env.example
```diff
@@ -143,14 +143,18 @@ ETHERPAD_SKIN_VARIANTS="super-light-toolbar super-light-editor light-background
...
 
 # LDAP base DN. Can be empty
```
duburcqa, line 149: It would be nice the explain the effect of using anymous bind vs authenticated bind

le-firehawk: Added L:163:165

env.example
le-firehawk: Added L:147

env.example
le-firehawk: cyrus' docs are somewhat indecisive, see below. I changed L:146 back to its original state

```
ldap_servers: <ldap://localhost/>
	Specify URI(s) refering to LDAP server(s), e.g. ldaps://10.1.1.2:999/.
	You can specify multiple servers separated by a space.
```

prosody/rootfs/defaults/saslauthd.conf
```diff
@@ -4,27 +4,47 @@
...
+
+  {{ if .Env.LDAP_BINDDN | default "" }}
+ldap_bind_dn: {{ .Env.LDAP_BINDDN }}
+ldap_bind_pw: {{ .Env.LDAP_BINDPW | default "MISSING PASSWORD" }}
+  {{ end }}
+
+ldap_filter: {{ .Env.LDAP_FILTER | default "uid=%u" }}
+
+ldap_group_attr: {{ .Env.LDAP_GROUP_ATTR | default "uniqueMember" }}
+  {{ if .Env.LDAP_GROUP_DN | default "" }}
```
saghul, line 21: I don't think the default in the check does anything here.

prosody/rootfs/defaults/saslauthd.conf
```diff
@@ -4,27 +4,47 @@
...
+ldap_group_dn: {{ .Env.LDAP_GROUP_DN }}
+ldap_group_match_method: {{ .Env.LDAP_GROUP_MATCH_METHOD | default "attr" }}
+  {{ end }}
+  {{ if .Env.LDAP_GROUP_SEARCH_BASE | default "" }}
+ldap_group_search_base: {{ .Env.LDAP_GROUP_SEARCH_BASE | default .Env.LDAP_BASE }}
+  {{ end }}
+  {{ if .Env.LDAP_GROUP_FILTER | default "" }}
```
saghul, line 28: same

prosody/rootfs/defaults/saslauthd.conf
```diff
@@ -4,27 +4,47 @@
...
+  {{ end }}
+
+ldap_filter: {{ .Env.LDAP_FILTER | default "uid=%u" }}
+
+ldap_group_attr: {{ .Env.LDAP_GROUP_ATTR | default "uniqueMember" }}
+  {{ if .Env.LDAP_GROUP_DN | default "" }}
+ldap_group_dn: {{ .Env.LDAP_GROUP_DN }}
+ldap_group_match_method: {{ .Env.LDAP_GROUP_MATCH_METHOD | default "attr" }}
+  {{ end }}
+  {{ if .Env.LDAP_GROUP_SEARCH_BASE | default "" }}
```
saghul, line 25: same

prosody/rootfs/defaults/saslauthd.conf
```diff
@@ -4,27 +4,47 @@
 
 {{ if eq $PROSODY_AUTH_TYPE "ldap" }}
 ldap_servers: {{ .Env.LDAP_URL }}
+
 ldap_search_base: {{ .Env.LDAP_BASE }}
-{{ if .Env.LDAP_BINDDN | default "" }}
```
le-firehawk, line 8: Pin

prosody/rootfs/defaults/saslauthd.conf
le-firehawk: Happy to remove, I took note of it being used already (see pin), I think it might be used to prevent unexpected behavior when the variable is undefined?

le-firehawk: Done

PR #1801, opened 2024-05-10
Add ENV for defaultLogoUrl

kamyabzad:
Added a new environment variable: `DEFAULT_LOGO_URL` to set config.defaultLogoUrl


web/rootfs/defaults/settings-config.js
```diff
@@ -557,6 +558,9 @@ config.whiteboard = {
     collabServerBaseUrl: '{{ $WHITEBOARD_COLLAB_SERVER_PUBLIC_URL }}'
 };
 
+// Configure application logo url
+config.defaultLogoUrl = {{ $DEFAULT_LOGO_URL }}
```
saghul, line 562: Don't you need quotes here?

PR #1809, opened 2024-05-21
jvb: adds autoscaler sidecar hook to signal final shutdown

aaronkvanmeerten:


sawall (approved):

sawall (approved):

jvb/rootfs/opt/jitsi/shutdown.sh
```diff
@@ -1,4 +1,8 @@
 #!/usr/bin/with-contenv bash
+# notify the sidecar of imminent shutdown
+[ -z "$AUTOSCALER_SIDECAR_PORT" ] && export AUTOSCALER_SIDECAR_PORT="6000"
```
saghul, line 3: No need to export it since there is no subshell. How about `PORT=${AUTOSCALER_SIDECAR_PORT:-6000}` and then use $PORT.

aaronkvanmeerten: agreed

saghul (approved):

PR #1838, opened 2024-06-27
feat(prometheus): Prometheus Scraper service 

24kushang:
### Added Prometheus Container for Metrics Collection

**Summary:**
- Introduced a Prometheus container to the Docker Compose setup.
- The Prometheus configuration is managed via an external `prometheus.yml` file.

**Details:**
- The configuration ensures the Prometheus container starts with `docker-compose -f docker-compose.yml -f prometheus.yml up`.
- Prometheus fetches its configuration from `prometheus/prometheus.yml`, which includes scrape configurations defining targets and the service name.

**Next Steps:**
- The initial configuration sets up basic scrape targets and service names.
- Future updates will enhance the Prometheus configuration with additional targets and detailed metrics.

**Testing:**
- Ensure Prometheus starts correctly with the specified configurations.
- Verify that metrics are being scraped and displayed in the Prometheus web interface accessible at `http://localhost:9090`.

---

Contributor - @24kushang


prometheus/prometheus.yml
```diff
@@ -0,0 +1,15 @@
...
+    scrape_interval: 5s
+    static_configs:
+      - targets: ["prosody:5280"]
+      
+  - job_name: "prometheus-jvb"
+    scrape_interval: 5s
+    static_configs:
+      - targets: ["jvb:8080"]
+
+  - job_name: "prometheus-jicofo"
```
aaronkvanmeerten, line 12: Is there a reason to have separate jobs for each target?

24kushang: This makes separation of concern and better individual service configuration. Suppose, we want to add authorization/authentication for one of the components or change the `scrape_interval` then it would be easier to maintain. 

We can have a single job named `prometheus` and have all the targets listed in there. Then we are required to create a common config with all the features listed for each and every service.

---

Whatever way we choose, it should be best for future use and should have ease of modification.

prometheus.yml
ltorje1, line 1: Allow setting restart policy from outside like using an env variable - see etherpad.yml - I would go with on-failure:2 by default https://docs.docker.com/config/containers/start-containers-automatically/

24kushang: updated the restart policy

ltorje-8x8 (approved):

prometheus/prometheus.yml
```diff
@@ -0,0 +1,9 @@
+scrape_configs:
+  # - job_name: "prometheus"
```
ltorje-8x8, line 2: why is the code commented?

prometheus.yml
ltorje-8x8: I cannot see it in code

prometheus/prometheus.yml
24kushang: Sorry I pushed the code that I was testing. Would change it and push it again.

aaronkvanmeerten (approved):

Issue #1836, opened 2024-06-27
Wrong param for ffmpeg stable-9457


SkyTaurus:
I update jitsi from stable-9258 to stable-9457 from dockerhub
After update JIBRI not working.
jibri stable-9258 ffmpeg version 4.3.6-0+deb11u1 Copyright (c) 2000-2023 the FFmpeg developers

stable-9457 ffmpeg version 5.1.4-0+deb12u1 Copyright (c) 2000-2023 the FFmpeg developers

The environment in the new version of Jibri is different, a newer version of X, a new version of the Debian base image
It seems that ffmpeg does not understand the --i argument at all for grabbing from X -f x11grab.

![image](https://github.com/jitsi/docker-jitsi-meet/assets/3510675/afa63b51-40ae-4f79-bd1e-1895dd2e4fd7)

if I move the -i argument immediately after -f x11grab, I can see that ffmpeg understands this argument.

![image](https://github.com/jitsi/docker-jitsi-meet/assets/3510675/2b9a166c-4bd9-4d82-8155-198a573f28da)

saghul:
Damn, this will require a change in Jibri... Ping @aaronkvanmeerten have we observed this?

damencho:
But if there are differences in ffmpeg, there can be any version on the server ...

emrahcom:
I remember this issue. It should be related with changed `xorg` config. The old config is not supported by Debian Bookworm.

https://github.com/jitsi/jibri/blob/master/resources/debian-package/etc/jitsi/jibri/xorg-video-dummy.conf#L108-L113

https://github.com/jitsi/docker-jitsi-meet/blob/master/jibri/rootfs/defaults/xorg-video-dummy.conf#L113-L118

emrahcom:
https://github.com/jitsi/jibri/pull/522

saghul:
Oh, have we released that Jibri to stable @damencho ?

emrahcom:
But Dockerized Jibri generates its own [xorg-video-dummy.conf](https://github.com/jitsi/docker-jitsi-meet/blob/master/jibri/rootfs/defaults/xorg-video-dummy.conf). It doesn't use the config comes from the deb package.

PR #1840, opened 2024-06-28
fix(jibri): display modes in xorg config

emrahcom:
Fix https://github.com/jitsi/docker-jitsi-meet/issues/1836


emrahcom:
See https://github.com/jitsi/docker-jitsi-meet/pull/1846

PR #1841, opened 2024-06-28
Add new environment variables for jigasi configuration

denzs:
As mentioned in https://community.jitsi.org/t/jigasi-sip-dialin-enable-sound-notification/132331/2

We are using 
* org.jitsi.jigasi.JVB_INVITE_TIMEOUT (needed for more than 30 seconds waiting time in the lobby)
* org.jitsi.jigasi.LOCAL_REGION (Jigasi should use the internal JVBs in our setup)
* org.jitsi.jigasi.xmpp.acc.BOSH_URL_PATTERN (to have Jigasi use Bosh to get the right shard via HAProxy)

I am open for discussions - of course everythin could be handled with .Env.JIGASI_CONFIGURATION but i think exposing some useful switches makes sense anyhow.. But let me know what you are thinking :)

### DIFF
--- a/jigasi/rootfs/defaults/sip-communicator.properties
+++ b/jigasi/rootfs/defaults/sip-communicator.properties
@@ -1,6 +1,10 @@
 {{ $ENABLE_TRANSCRIPTIONS := .Env.ENABLE_TRANSCRIPTIONS | default "0" | toBool -}}
 {{ $JIGASI_BREWERY_MUC := .Env.JIGASI_BREWERY_MUC | default "jigasibrewery" -}}
 {{ $JIGASI_XMPP_USER := .Env.JIGASI_XMPP_USER | default "jigasi" -}}
+{{ $JIGASI_JVB_TIMEOUT := .Env.JIGASI_JVB_TIMEOUT | default "30000" -}}
+{{ $JIGASI_LOCAL_REGION := .Env.JIGASI_LOCAL_REGION | default "" -}}
+{{ $BOSH_URL_PATTERN := .Env.BOSH_URL_PATTERN | default "" -}}
+{{ $USE_TRANSLATOR_IN_CONFERENCE := .Env.USE_TRANSLATOR_IN_CONFERENCE | default "0" | toBool -}}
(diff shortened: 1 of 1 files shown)


jigasi/rootfs/defaults/sip-communicator.properties
```diff
@@ -187,3 +191,24 @@ org.jitsi.jigasi.MUC_SERVICE_ADDRESS={{ $XMPP_MUC_DOMAIN }}
...
+{{ end }}
+
+{{ if $USE_TRANSLATOR_IN_CONFERENCE }}
+org.jitsi.jigasi.xmpp.acc.USE_TRANSLATOR_IN_CONFERENCE=true
+net.java.sip.communicator.impl.protocol.sip.acc1.USE_TRANSLATOR_IN_CONFERENCE=true
+# Should be enabled when using translator mode
+net.java.sip.communicator.impl.neomedia.audioSystem.audiosilence.captureDevice_list=["AudioSilenceCaptureDevice:noTransferData"]
+{{ end }}
+
+{{ if .Env.JIGASI_CONFIGURATION -}}
```
damencho, line 212: What is this?

denzs: Same thing as https://github.com/jitsi/docker-jitsi-meet/blob/master/prosody/rootfs/defaults/conf.d/jitsi-meet.cfg.lua#L267 - we are setting some more custom options to finetune Jigasi in our setup, like SIP_PREFERRED_CLEAR_PORT but i thought it wouldnt make sense to expose every possible config option seperated, so JIGASI_CONFIGURATION can be used to render custom snippets inside the config in addition to the predefined options..

aaronkvanmeerten (approved):

damencho (approved):

emrahcom:
The added environmnt variables don't exist in https://github.com/jitsi/docker-jitsi-meet/blob/master/jigasi.yml

saghul:
True that @denzs . Can you please add them?

denzs:
Thanks for pointing that out! see https://github.com/jitsi/docker-jitsi-meet/pull/1856

PR #1845, opened 2024-07-01
Add healtcheck.sh scripts for jigasi, jicofo, jvb and prosody, add cu…

denzs:
coming from https://github.com/jitsi/jigasi/issues/539

We are using the container images with podman and are going to leverage the integrated healthcheck features..

This PR adds some basic healtcheck.sh scripts for jigasi, jicofo, jvb and prosody but does not add them as healthcheck in the Dockerfiles - because we are not using docker and i cant estimate if there are any sideeffects.

Anyhow i think having general healtcheck scripts available in the images would be a good thing :)

### DIFF
--- a/jicofo/rootfs/healthcheck.sh
+++ b/jicofo/rootfs/healthcheck.sh
@@ -1,3 +1,0 @@
-#!/bin/bash
-
-curl --fail-with-body -v localhost:8888/about/health
--- a/jicofo/rootfs/healthcheck.sh
+++ b/jicofo/rootfs/healthcheck.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+curl --fail-with-body -v localhost:8888/about/health
--- a/prosody/Dockerfile
+++ b/prosody/Dockerfile
@@ -44,6 +44,7 @@ RUN wget -qO /etc/apt/trusted.gpg.d/prosody.gpg https://prosody.im/files/prosody
       lua-ldap \
       lua-luaossl \
       lua-sec \
+      curl \
       lua-unbound && \
     apt-dpkg-wrap apt-get -d install -y jitsi-meet-prosody && \
     dpkg -x /var/cache/apt/archives/jitsi-meet-prosody*.deb /tmp/pkg && \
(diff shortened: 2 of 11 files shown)


prosody/Dockerfile
```diff
@@ -44,6 +44,7 @@ RUN wget -qO /etc/apt/trusted.gpg.d/prosody.gpg https://prosody.im/files/prosody
       lua-ldap \
       lua-luaossl \
       lua-sec \
+      curl \
```
saghul, line 47: Let's put curl in the base image instead please.

jicofo/rootfs/healthcheck.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
saghul, line 1: Don't put these files directly in the root please. /usr/local/bin feels like the right palce for them.

jicofo/rootfs/healthcheck.sh
```diff
@@ -0,0 +1,3 @@
...
+
+curl --fail-with-body -v localhost:8888/about/health
```
saghul, line 3: Why the verbose flag? Also, any reason not to use 127.0.0.1 and skip a lookup?

saghul (changes requested):
Left some comments, PTAL! Also @aaronkvanmeerten WDYT?

jicofo/rootfs/healthcheck.sh
denzs: good point, it is removed :+1:

jicofo/rootfs/healthcheck.sh
denzs: done

prosody/Dockerfile
denzs: done :+1:

denzs:
@saghul i think i've considered every suggestion, thanks for the input! :)
Anyhow, Github still shows "Changes requested" - but i dont see any open discussions :shrug:

saghul (approved):
LGTM! LEt's wait for @aaronkvanmeerten 's feedback.

prosody/Dockerfile
```diff
@@ -57,6 +57,7 @@ RUN wget -qO /etc/apt/trusted.gpg.d/prosody.gpg https://prosody.im/files/prosody
     rm -rf /usr/share/lua/{5.1,5.2,5.3} && \
     wget -qO /prosody-plugins/mod_auth_cyrus.lua https://hg.prosody.im/prosody-modules/raw-file/65438e4ba563/mod_auth_cyrus/mod_auth_cyrus.lua && \
     wget -qO /prosody-plugins/sasl_cyrus.lua https://hg.prosody.im/prosody-modules/raw-file/65438e4ba563/mod_auth_cyrus/sasl_cyrus.lua  && \
+    wget -qO /prosody-plugins/mod_http_health.lua https://hg.prosody.im/prosody-modules/raw-file/2b80188448d1/mod_http_health/mod_http_health.lua && \
```
aaronkvanmeerten, line 60: This one has a different hash than the rest, not urgent but I'm curious if we can either have the same hash or update the others to a newer one?  Doesn't have to be in this PR.

aaronkvanmeerten (approved):
LGTM, curious about the difference in hash for the mercurial source fetch, but not enough to block merging.

prosody/Dockerfile
denzs: Thanks for merging! :)

Let me try to satisfy your curiosity.. ;)

2b80188448d1/5924 refers to the current tip (June 2024) of the hg repository.
65438e4ba563/4929 refers to a commit from Apr 18 2022.

mod_http_health was added in 5667:9bcd257dea4e (Oct 2023).

That is why 65438e4ba563 was not feasible for mod_http_health.

After diffing 65438e4ba563 and 2b80188448d1 i can confirm, there are no differences in mod_auth_cyrus..

So there is nothing to be said against using the current tip 2b80188448d1 for mod_auth_cyrus too.

Want a PR? :)

PR #1852, opened 2024-07-03
Copy mod_c2s.lua from prosody container and update default_port

### LINKED ISSUE #1812: Hardcoded prosody port in mod_c2s.lua
After changing the default `XMPP_PORT` to '5223', prosody was still trying to use 5222 in addition to 5223:

```
jitsi-prosody-1   jitsi/prosody:stable-9457-2   "/init"   prosody   16 minutes ago   Up 8 minutes    5222-5223/tcp, 5269/tcp, 5280/tcp, 5347/tcp
```

After inspecting the container, it looks like the jitsi prosody container has a hardcoded value for the default port, which does not use `XMPP_PORT` at all:
```
root@1b30451b8a46:/# rgrep 5222 /lib/prosody/
/lib/prosody/modules/mod_c2s.lua:       default_port = 5222;
```

In this setup, there's a Jitsi stack already running in parallel, using port 5222 for prosody, which is why I need to change it.

This results in jvb and jicofo being very unhappy:
```
jvb-1      | 2024-05-28T08:40:08.847079060Z org.jivesoftware.smack.SmackException$EndpointConnectionException: The following addresses failed: 'RFC 6120 A/AAAA Endpoint + [xmpp.meet.jitsi:5223] (xmpp.meet.jitsi/172.31.0.2:5223)' failed because: java.net.ConnectException: Connection refused
jicofo-1   | 2024-05-28T08:40:08.852750162Z org.jivesoftware.smack.SmackException$EndpointConnectionException: The following addresses failed: 'RFC 6120 A/AAAA Endpoint + [xmpp.meet.jitsi:5223] (xmpp.meet.jitsi/172.31.0.2:5223)' failed because: java.net.ConnectException: Connection refused
```

Am I doing anything wrong, or is it really a hardcoded value for 5222/tcp?

teyfix:
Closes #1812 

This pull request includes the following changes:

- Copied mod_c2s.lua from the prosody container to the appropriate location.

- Updated the default_port in mod_c2s.lua to use the XMPP_PORT environment variable.

- Added mod_c2s.lua to cont-init.d

_P.S I'm not sure this is the right way to add mod_c2s.lua_

### DIFF
--- a/prosody/rootfs/defaults/lib/prosody/modules/mod_c2s.lua
+++ b/prosody/rootfs/defaults/lib/prosody/modules/mod_c2s.lua
@@ -1,0 +1,488 @@
+-- Prosody IM
+-- Copyright (C) 2008-2010 Matthew Wild
+-- Copyright (C) 2008-2010 Waqas Hussain
+--
+-- This project is MIT/X11 licensed. Please see the
+-- COPYING file in the source package for more information.
+--
+
+module:set_global();
+
+local add_task = require "util.timer".add_task;
+local new_xmpp_stream = require "util.xmppstream".new;
+local nameprep = require "util.encodings".stringprep.nameprep;
+local sessionmanager = require "core.sessionmanager";
+local statsmanager = require "core.statsmanager";
+local st = require "util.stanza";
+local sm_new_session, sm_destroy_session = sessionmanager.new_session, sessionmanager.destroy_session;
+local uuid_generate = require "util.uuid".generate;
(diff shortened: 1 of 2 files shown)


damencho:
Nope, it is not acceptable to copy prosody code here. That port should be configurable through the config.
https://prosody.im/doc/ports
You can do that via `c2s_ports = { 5222, 5322 } -- Listen on 5322`

PR #1853, opened 2024-07-03
Change: Drive image compilation with JITSI_IMAGE_VERSION

Berbe:
The `JITSI_IMAGE_VERSION` environment variable, part of `env.example`, is already made for driving images tag selection at runtime.

This proposal aims at:
* Using `JITSI_IMAGE_VERSION` environment variable also at build time, making image generation and stack deployment coherent
* Ensuring compilation of all images is done with that same tag


saghul:
This would break the current workflow of `make prepare` creating `latest` images for testing.

saghul:
The release.sh script would be broken.

PR #1864, opened 2024-07-19
Fix registry jigasi brewery with ENABLE_TRANSCRIPTIONS

marcoadasilvaa:
Ref: https://community.jitsi.org/t/docker-jigasi-config-error-for-transcription/132505/8?u=marcoadasilvaa


jicofo/rootfs/defaults/jicofo.conf
```diff
@@ -203,7 +204,7 @@ jicofo {
     }
     {{ end }}
 
-    {{ if and .Env.JIGASI_SIP_URI $JIGASI_BREWERY_MUC }}
+    {{ if or .Env.JIGASI_SIP_URI .Env.ENABLE_TRANSCRIPTIONS }}
```
saghul, line 207:
```suggestion
    {{ if or .Env.JIGASI_SIP_URI $ENABLE_TRANSCRIPTIONS }}
```

saghul: Use the variable you just declared.

saghul (changes requested):

saghul (approved):

jicofo/rootfs/defaults/jicofo.conf
marcoadasilvaa: Thanx!

PR #1876, opened 2024-08-12
Enhancement: Improve Access to Recorded Videos by Organizing Saved Files

MR-IOErr:
I'm using Jitsi with Docker, and I encountered some challenges when trying to manage recorded conference videos. The videos are saved in a directory mounted on the Docker container, but they are stored in folders with hash string names. This makes it difficult to identify and locate specific recordings.

Currently, after recording a conference, the videos are saved in a directory with a hashed folder name. Administrators or users who need to access these videos must manually navigate through the hash-named directories, which can be time-consuming and confusing, especially as the number of recordings grows.

To address this, I developed a script that organizes the recorded videos in a more user-friendly manner. The script automatically:
- Renames the videos to more descriptive names based on the conference name and timestamp.
- Moves the videos to a structured directory hierarchy for easier access.

This enhancement will simplify the process of managing recorded videos, making it more intuitive and less error-prone.

Users can quickly find and access recorded videos without dealing with hash strings.
Administrators will spend less time managing the video storage, reducing overhead.
 As the number of recordings increases, this approach will help keep the file system organized.

I would love to contribute this script to the Jitsi project if you find it useful. Please let me know your thoughts on this enhancement.

### DIFF
--- a/backup/README.md
+++ b/backup/README.md
@@ -17,7 +17,7 @@ This script is designed to help administrators and users of Jitsi, who are runni
 - A CronJob for executing script in case you need to have a frequent execution process.
 
 
-## Configuration:
+## Configuration
 
 Open the script "main.go" and configure the following variables:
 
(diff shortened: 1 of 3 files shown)


aaronkvanmeerten:
Hi and thank you for your contribution, this looks really useful indeed!

We actually have a separate github group for community contributions that don't quite fit into the core product/service but are still of use to the global community.  This feels like it could be in a repo there, which we'd grant you commit access to, so you could maintain it at your leisure.  If this is of interest to you, I'll get a repo set up for you and get you an invite.  Let me know!  and sorry again for the delayed response to this PR, it really is a nice contribution!

MR-IOErr:
Hey Aaron,
Thanks for your reply.

Sure, I'll add it to the repo. It'd be great if you could send me an invite.

Cheers,
Arian Moafizad

PR #1882, opened 2024-08-21
jaas: add ability to run the web container targetting JaaS

saghul:
If JAAS_APP_ID is set, the config will be updated to point to the production JaaS deployment for the given App ID.


saghul:
@damencho @aaronkvanmeerten PTAL! The next update will fix the CORS problem so this would be ready then.

damencho (approved):

PR #1883, opened 2024-08-21
web: remove unnecessary option

saghul:


saghul:
@damencho I see we don't use this in prod and get it from the conference request. Does it make sense to remove the option altogether from lib-jitsi-meet?

damencho:
Well ... I think people who do not run https://github.com/jitsi/jitsi-meet/blob/master/resources/prosody-plugins/mod_client_proxy.lua and prosody as their XMPP server (the Openfire guys) use this setting so they can communicate with jicofo.

saghul:
But this Docker, it uses prosody here, so we can assume that, can't we?

damencho:
Yes, we can drop it. But in the other places we set config.hosts.focus and I do not see it here. Does it work without it?

saghul:
Yep it does.

PR #1737, opened 2024-02-07
Unify hidden domain usage for recording and transcription

zobadaniel:
More generic approach to proper transcriber inclusion into docker.

I made a new pull request, based on the info & feedback from here: https://github.com/jitsi/docker-jitsi-meet/pull/1663 

The patch is bigger because I renamed XMPP_RECORDER_DOMAIN. The name would be confusing if it is not the sole purpose of that domain.

I'm still not 100% sure this is the right way to do it. Please let me know your comments.

Patch has been tested vs release 9220.

PR #1945, opened 2024-10-20
Rename recorder domain to hidden domain

M4GNV5:
This is bacially the leftovers of #1737.
Since the addition of `transcriber.yaml` by @aaronkvanmeerten the main functionality from the original PR already exist.
This PR simply renames the "recorder" domain to a more generic "hidden" domain, since it is used not only by the recorder, but also by the transcriber.

### DIFF
--- a/jicofo/rootfs/defaults/jicofo.conf
+++ b/jicofo/rootfs/defaults/jicofo.conf
@@ -6,6 +6,7 @@
 {{ $JICOFO_AUTH_LIFETIME := .Env.JICOFO_AUTH_LIFETIME | default "24 hours" -}}
 {{ $ENABLE_SCTP := .Env.ENABLE_SCTP | default "1" | toBool -}}
 {{ $ENABLE_RECORDING := .Env.ENABLE_RECORDING | default "0" | toBool -}}
+{{ $ENABLE_TRANSCRIPTIONS := .Env.ENABLE_TRANSCRIPTIONS | default "0" | toBool -}}
 {{ $ENABLE_OCTO := .Env.ENABLE_OCTO | default "0" | toBool -}}
 {{ $ENABLE_OCTO_SCTP := .Env.ENABLE_OCTO_SCTP | default $ENABLE_SCTP | toBool -}}
 {{ $ENABLE_AUTO_LOGIN := .Env.ENABLE_AUTO_LOGIN | default "1" | toBool -}}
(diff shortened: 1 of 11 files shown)


prosody/rootfs/defaults/conf.d/visitors.cfg.lua
```diff
@@ -187,18 +187,23 @@ Component '{{ $VISITORS_MUC_PREFIX }}.v{{ $VISITOR_INDEX }}.{{ $VISITORS_XMPP_DO
...
-	-- The time in seconds, after which the limit for an IP address is lifted.
-	rate_limit_timeout = {{ $RATE_LIMIT_TIMEOUT }};
-	-- List of regular expressions for IP addresses that are not limited by this module.
-	rate_limit_whitelist = {
-      "127.0.0.1";
-      {{ range $index, $cidr := (splitList "," $RATE_LIMIT_ALLOW_RANGES | compact) -}}
-      "{{ $cidr }}";
-      {{ end -}}
-    };
+    rate_limit_login_rate = {{ $RATE_LIMIT_LOGIN_RATE }};
```
saghul, line 190: Can you please avoid the unnecessary formatting changes here?

saghul (approved):
LGTM with a minor nit. @aaronkvanmeerten PTAL too!

jicofo/rootfs/defaults/jicofo.conf
```diff
@@ -6,6 +6,7 @@
 {{ $JICOFO_AUTH_LIFETIME := .Env.JICOFO_AUTH_LIFETIME | default "24 hours" -}}
 {{ $ENABLE_SCTP := .Env.ENABLE_SCTP | default "1" | toBool -}}
 {{ $ENABLE_RECORDING := .Env.ENABLE_RECORDING | default "0" | toBool -}}
+{{ $ENABLE_TRANSCRIPTIONS := .Env.ENABLE_TRANSCRIPTIONS | default "0" | toBool -}}
```
aaronkvanmeerten, line 9: I believe this is being moved for alphabetizing (which I approve of), but also duplicates an existing entry for this variable further down in the file.  I suggest remove line 16 below if we're adding this one here.

aaronkvanmeerten (approved):
I approve, excepting the duplicated line I referenced i my comments should be removed.

prosody/rootfs/defaults/conf.d/visitors.cfg.lua
M4GNV5: I do agree this is mostly formatting, but before this PR there was mixed indentation (tabs/spaces) in this file.
For next time I assume the better approach would be to create a seperate PR for this?

saghul: Generally speaking, yes. But since you are fixing spacing, let's leave it!

M4GNV5:
I rebased this PR and added the one remaining change request. I hope this makes everything ready to be merged

saghul (approved):

PR #1949, opened 2024-10-22
feat(prosody): move external_services module and config to global prosody.cfg.lua

aaronkvanmeerten:


damencho (approved):

saghul:
No objection, just curious: why?

aaronkvanmeerten:
> No objection, just curious: why?

It removes duplicated code between visitors and standard, and also makes it more inline with what external_services does as a module, which doesn't really related to any specific vhost.

saghul:
Great!

PR #1953, opened 2024-10-23
feat(transcriber): flag to enable translations

aaronkvanmeerten:


jigasi/rootfs/defaults/transcriber-sip-communicator.properties
```diff
@@ -2,6 +2,7 @@
 {{ $JIGASI_TRANSCRIBER_FILTER_SILENCE := .Env.JIGASI_TRANSCRIBER_FILTER_SILENCE | default "0" | toBool -}}
 {{ $XMPP_RECORDER_DOMAIN := .Env.XMPP_RECORDER_DOMAIN | default "recorder.meet.jitsi" -}}
 {{ $JIGASI_TRANSCRIBER_ENABLE_SAVING := .Env.JIGASI_TRANSCRIBER_ENABLE_SAVING | default "1" | toBool -}}
+{{ $JIGASI_TRANSCRIBER_ENABLE_TRANSLATION := .Env.JIGASI_TRANSCRIBER_ENABLE_TRANSLATION | default "1" | toBool -}}
```
damencho, line 5: Shouldn't this be off by default? It requires extra configuration in any of the supported translations.

aaronkvanmeerten: good question, I set it on by default since it was the previous default behavior.  I agree it probably does need to be off unless configured.

PR #1955, opened 2024-10-25
feat(transcriber): fix exposing transcript path in web

### LINKED ISSUE #1427: When I click on the JIGASI TRANSCRIBER ADVERTISE URL, I get 404 error
When I try to access the txt files I recorded with jigasi and google speech to text services, I get the following error. Actually, I understood the cause of the error, but I couldn’t find how to solve it. Waiting for your valuable answers.
Ekran Resmi 2022-10-29 15 14 14

In the picture above, jitsi is sending the link where he recorded the meeting chat. But when I click on the link I get 404 error. also when I look at the logs of the web container
`{"log":"2022/10/26 20:07:12 [error] 260#260: *62 open() \"/usr/share/jitsi-meet/2022-10-26T20:07:00.591530Z_35c4790f- 7f4a-46cf-854d-b302f880d7f4/transcript_2022-10-26T20:07:00.592183Z_dc3d59c5-a967-4759-9dd0-5dad7e3181de.txt\" failed (2: No such file or directory)`

I see log. As far as I understand, the txt files are saved in the /usr/share/jitsi-meet/transcripts path, but it searches for the files directly in the /usr/share/jitsi-meet path. how can i solve this problem?

bmagistro:
This addes the transcipts path to the list of static items that should be allowed by nginx.

Fixes #1427

### DIFF
--- a/web/rootfs/defaults/meet.conf
+++ b/web/rootfs/defaults/meet.conf
@@ -60,7 +60,7 @@ location = /_api/room-info {
 {{ end }}
 
 # ensure all static content can always be found first
-location ~ ^/(libs|css|static|images|fonts|lang|sounds|connection_optimization|.well-known)/(.*)$ {
+location ~ ^/(libs|css|static|images|fonts|lang|sounds|connection_optimization|.well-known|transcripts)/(.*)$ {
     add_header 'Access-Control-Allow-Origin' '{{ $CORS_HEADER_ACCESS_CONTROL_ALLOW_ORIGIN }}';
     alias /usr/share/jitsi-meet/$1/$2;
 


saghul:
Have you tested this? I don't think ti would work since the web container is not where the transcript files are stored, are they?

bmagistro:
Very valid question as I expected it to be more complicated too.  I did and have it running on a test instance right now.  

Looking now at the why/how, it looks like transcripts are volume mounted in via the web compose https://github.com/jitsi/docker-jitsi-meet/blob/master/docker-compose.yml#L14C1-L15C1 .  I didn't look to see why when developing this, just exec'd in and started looking around based on the logs but also knew our compose files should be pretty close to the stock ones provided.

saghul:
Ah looks like we have them mounted here: https://github.com/jitsi/docker-jitsi-meet/blob/74d913f58acfe01573aa335ea363b8487137078d/transcriber.yml#L9

LGTM @aaronkvanmeerten any objections?

aaronkvanmeerten (approved):

saghul (approved):

PR #1964, opened 2024-11-12
feat(web): allow comment at the top of config.js to be overridden

aaronkvanmeerten:


damencho (approved):

saghul (approved):
Why do we want to make this configurable? Can't the custom file add comments to the bottom?

aaronkvanmeerten:
To duplicate the comment at the top of 8x8.vc/config.js to reduce hackerone spam reports

PR #1969, opened 2024-11-21
feat(web): include custom nginx conf files

emrahcom:
PR adds an include line into Nginx config to allow custom configuration in it.

Like `include /etc/jitsi/meet/jaas/*.conf` line in [jitsi-meet.example](https://github.com/jitsi/jitsi-meet/blob/master/doc/debian/jitsi-meet/jitsi-meet.example).

In my use-case, I will add my OIDC backend as proxy by creating a config file in this folder.
Not having this folder (`/config/nginx-custom/`) by default is not a problem for Nginx.

### DIFF
--- a/web/rootfs/defaults/meet.conf
+++ b/web/rootfs/defaults/meet.conf
@@ -38,6 +38,8 @@ add_header X-Jitsi-Shard {{ .Env.DEPLOYMENTINFO_SHARD }};
 # Opt out of FLoC (deprecated)
 add_header Permissions-Policy "interest-cohort=()";
 
+include /config/nginx-custom/*.conf;
+
 location = /config.js {
     alias /config/config.js;
 }


saghul (approved):

atlashour:
If I modify that line myself at meet.conf, nothing happens when I run: docker compose up -d.

Why when you change something at rootfs it won't work in the docker container?

Need some help here, thank you so much for you contribution. I hope your feature will be integrated soon.

emrahcom:
> If I modify that line myself at meet.conf, nothing happens when I run: docker compose up -d.
> 
> Why when you change something at rootfs it won't work in the docker container?

Hi @atlashour, this change has no effect if you don't have custom config files in `/config/nginx-custom` but if you want to inject some additional configs into Nginx then you may put them into this folder as files.

This type of customization is needed in some setups.

PR #1985, opened 2024-12-20
Add Env for Shared document random name

daimoc:
Hi,
Since many years, Jicofo has a parameter to use a random name in Etherpad : https://github.com/jitsi/jicofo/blob/5803269e775b84aa53d18cade50e81c33a26cc47/jicofo-selector/src/main/resources/reference.conf#L223.
 
It makes the Etherpad link only known by conference participant and more secure.
This pull request adds this option in the docker version.

### DIFF
--- a/jicofo/rootfs/defaults/jicofo.conf
+++ b/jicofo/rootfs/defaults/jicofo.conf
@@ -188,10 +188,12 @@ jicofo {
       enable-multi-stream-backward-compat = {{ .Env.JICOFO_MULTI_STREAM_BACKWARD_COMPAT | toBool }}
       {{ end }}
 
+      {{ if .Env.ENABLE_SHARED_DOCUMENT_RANDOM_NAME }}
       shared-document {
         use-random-name = {{ .Env.ENABLE_SHARED_DOCUMENT_RANDOM_NAME | toBool }}
       }
-      
+      {{ end }}
+
     }
 
     {{ if .Env.JICOFO_ENABLE_HEALTH_CHECKS }}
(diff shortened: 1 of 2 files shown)


jicofo/rootfs/defaults/jicofo.conf
```diff
@@ -187,6 +188,12 @@ jicofo {
       enable-multi-stream-backward-compat = {{ .Env.JICOFO_MULTI_STREAM_BACKWARD_COMPAT | toBool }}
       {{ end }}
 
+      {{ if .Env.ENABLE_SHARED_DOCUMENT_RANDOM_NAME }}
```
aaronkvanmeerten, line 191:
```suggestion
      {{ if $ENABLE_SHARED_DOCUMENT_RANDOM_NAME }}
```
Since we have a variable defined we ought to use it.  Otherwise looks great to me!

jicofo/rootfs/defaults/jicofo.conf
```diff
@@ -187,6 +188,12 @@ jicofo {
       enable-multi-stream-backward-compat = {{ .Env.JICOFO_MULTI_STREAM_BACKWARD_COMPAT | toBool }}
       {{ end }}
 
+      {{ if $ENABLE_SHARED_DOCUMENT_RANDOM_NAME }}
+      shared-document {
+        use-random-name = {{ .Env.ENABLE_SHARED_DOCUMENT_RANDOM_NAME | toBool }}
```
saghul, line 193: Use the variable here too, that way you don't need to use toBool.

daimoc: That's done, thanks.

saghul (approved):

daimoc:
Hi, just checking in on this PR. Let me know if anything is needed from my side to help with the validation. Thanks!

saghul:
Opsie, merged now!

PR #1989, opened 2024-12-31
feat(prosody) - Added TURN_USERNAME and TURN_PASSWORD

lalayueh:
# Fix: Add support for TURN servers that don't implement draft-uberti-behave-turn-rest-00

## Problem
Cloudflare TURN servers do not support the draft-uberti-behave-turn-rest-00 specification, which is the default method used by mod_external_services to generate TURN credentials (username and password).

## Solution
Added two new environment variables:
- `TURN_USERNAME`: Define a static TURN username
- `TURN_PASSWORD`: Define a static TURN password

When `TURN_CREDENTIALS` is not set, the module will not attempt to generate credentials using the draft-uberti-behave-turn-rest-00 method. Instead, it will use the static credentials if provided through `TURN_USERNAME` and `TURN_PASSWORD`.

This change allows Prosody to work with TURN servers that don't implement the expired draft spec, such as Cloudflare TURN.

## Reference
- [Cloudflare TURN FAQ](https://developers.cloudflare.com/calls/turn/faq/#does-cloudflare-calls-turn-support-the-expired-ietf-rfc-draft-draft-uberti-behave-turn-rest-00)

### DIFF
--- a/prosody/rootfs/defaults/prosody.cfg.lua
+++ b/prosody/rootfs/defaults/prosody.cfg.lua
@@ -331,7 +331,7 @@ log = {
 {{ end }}
 }
 
-{{ if $PROSODY_ENABLE_METRICS }} 
+{{ if $PROSODY_ENABLE_METRICS }}
 -- Statistics Provider configuration
 statistics = "internal"
 statistics_interval = "manual"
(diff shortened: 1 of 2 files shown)


prosody/rootfs/defaults/prosody.cfg.lua
```diff
@@ -351,15 +351,53 @@ external_services = {
...
         {{- if or $STUN_HOST $TURN_HOST $idx -}},{{- end }}
-        { type = "turns", host = "{{ $host }}", port = {{ $TURNS_PORT }}, transport = "tcp", secret = true, ttl = {{ $TURN_TTL }}, algorithm = "turn" }
+        {
+            type = "turns",
+            host = "{{ $host }}",
+            port = {{ $TURNS_PORT }},
+            transport = "tcp",
+            ttl = {{ $TURN_TTL }},
+
+            {{ if $.Env.TURN_CREDENTIALS -}}
```
saghul, line 388: Shouldn't this just be `.Env.TURN_CREDENTIALS` ?

prosody/rootfs/defaults/prosody.cfg.lua
```diff
@@ -351,15 +351,53 @@ external_services = {
...
+            secret = true,
+            algorithm = "turn",
+            {{- end }}
+
+            {{ if $.Env.TURN_USERNAME -}}
```
saghul, line 393: Ditto.

prosody/rootfs/defaults/prosody.cfg.lua
```diff
@@ -351,15 +351,53 @@ external_services = {
...
+            username = "{{$.Env.TURN_USERNAME}}",
+            {{- end }}
+
+            {{ if $.Env.TURN_PASSWORD -}}
```
saghul, line 397: Ditto.

prosody/rootfs/defaults/prosody.cfg.lua
lalayueh: Because `.Env` tries to access the Env field from the current scope rather than the root context.

In Go templates, the `$` symbol refers to the root context. When accessing top-level variables like `Env`, we need to use `$.Env` instead of `.Env` to ensure we're accessing from the root scope rather than the current context.

saghul: Ops, I forgot about that! Thanks for the reminder!

saghul (approved):

Issue #1956, opened 2024-10-28
Grafana/log-analyser not working after fresh setup with docker on Ubuntu24


pzystorm:
Hi,
to install I have followed these instructions: https://jitsi.github.io/handbook/docs/devops-guide/devops-guide-log-analyser/

My system:

```
# docker -v
Docker version 27.3.1, build ce12230
# docker compose version
Docker Compose version v2.29.7
# lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 24.04.1 LTS
Release:        24.04
Codename:       noble
```

After starting I first had the problem, that Grafana complained about not having permissions to write to /var/lib/grafana. But after `chmod -R a+w log-analyser` this problem is gone. I have realized that grafana is not runnning under root in the docker container but rather under a local user with the id 472.

Now it starts, but after doing a lot of migration and import stuff with the output of "Migration successfully executed" it stated, that there is some error with the admin user and some dashboards and a database lock:

```grafana-1  | logger=ngalert.multiorg.alertmanager t=2024-10-28T01:47:51.446368221Z level=info msg="Starting MultiOrg Alertmanager"
grafana-1  | logger=ngalert.migration t=2024-10-28T01:47:51.448727439Z level=info msg=Starting
grafana-1  | logger=ngalert.migration t=2024-10-28T01:47:51.449037184Z level=info msg="Starting legacy migration"
grafana-1  | logger=ngalert.migration orgID=1 t=2024-10-28T01:47:51.449198965Z level=info msg="Migrating alerts for organisation"
grafana-1  | logger=ngalert.migration orgID=1 t=2024-10-28T01:47:51.449499582Z level=info msg="Alerts found to migrate" alerts=0
grafana-1  | logger=ngalert.migration orgID=1 t=2024-10-28T01:47:51.449852528Z level=warn msg="No available receivers"
grafana-1  | logger=dashboard-service t=2024-10-28T01:47:51.451198337Z level=error msg="Could not make user admin" dashboard="Jitsi Web Dashboard" namespaceID=user userID=0 error="identifier is not initialized"
grafana-1  | logger=dashboard-service t=2024-10-28T01:47:51.452762025Z level=error msg="Could not make user admin" dashboard="JVB Dashboard" namespaceID=user userID=0 error="identifier is not initialized"
[...]

PR #1991, opened 2025-01-05
Grafana/log-analyser not working after fresh setup with docker on Ubuntu24

pzystorm:
As complained in this bug https://github.com/jitsi/docker-jitsi-meet/issues/1956 I have now found a solution what needs to be changed:

- it seems the datasource ids are random ids. The current grafana version can handle plaintext names of the datsources
- the datasources were not imported, because ./log-analyser/grafana-provisioning/datasources was not mounted
- after one day I have always reached some max_send_msg_size limit and Grafana could not get data from loki, to the limit has been set from 4mb to 32mb and works now much much better than before
- the otel component has complained about "loglevel: debug" which seems to be some old style format


saghul:
Ping @24kushang :-)

PR #1999, opened 2025-01-17
Making whiteboard integrating a bit simple to understand

yajatkaul:
Since i have been through a lot due to lack of documentation this boilerplate change might be really helpful for people trying to integrate whiteboard with this docker jitsi


docker-compose.yml
```diff
@@ -1,500 +1,512 @@
-version: '3.5'
+version: "3.5"
```
saghul, line 1: Please revert to the existing style.

docker-compose.yml
```diff
@@ -1,500 +1,512 @@
...
+      - XMPP_AUTH_DOMAIN
+      - XMPP_INTERNAL_MUC_DOMAIN
+      - XMPP_SERVER
+      - XMPP_PORT
+    depends_on:
+      - prosody
+    networks:
+      meet.jitsi:
+  
+  # Whiteboard seamless integrating with docker boiler plate
```
saghul, line 498: We already have https://github.com/jitsi/docker-jitsi-meet/blob/master/whiteboard.yml

env.example
```diff
@@ -89,7 +89,7 @@ TZ=UTC
 #WHITEBOARD_COLLAB_SERVER_URL_BASE=http://whiteboard.meet.jitsi
 
 # Set the excalidraw-backend public URL (uncomment to enable)
-#WHITEBOARD_COLLAB_SERVER_PUBLIC_URL=https://whiteboard.meet.my.domain
+#WHITEBOARD_COLLAB_SERVER_PUBLIC_URL=http://127.0.0.1:3002
```
saghul, line 92: This is for the external version, if you want to use the existing one in whiteboard.yml then uncometing the one above is enough.

saghul (commented):
Thanks for the attempt at improving things! Not sure exactly how this helps, can you clarify?

yajatkaul:
its not a really impactful change but i find it to be quiet helpful since there are not many resources out there and integrating these 2 backends i found was a little tricky with this docker yml change you can get all of it running in one go and the env.example also hints on the settings. This allows people with less docker knowledge also work it out

saghul:
Yeah, the docs are lacking. I'd suggest a PR to the handbook clarifying what to do: https://jitsi.github.io/handbook/docs/devops-guide/devops-guide-docker

yajatkaul:
yep had a look seems to documentation fix

PR #2028, opened 2025-02-21
feat: Add an env var for jicofo ice failure detection.

bgrozev:


aaronkvanmeerten (approved):

saghul:
Is there a reason for not having this default to true?

bgrozev:
Just an abundance of caution. I turned it on for all of our environments.

PR #2034, opened 2025-02-27
fix(jibri): streaming max bitrate needs units

aaronkvanmeerten:


aaronkvanmeerten (commented):
this is bringing the param in line with the reference.conf

https://github.com/jitsi/jibri/blob/3325e37a1de3936d3cd09d52b0cf59f5c6e861a1/src/main/resources/reference.conf#L55

saghul (approved):

PR #2036, opened 2025-02-28
fix(prosody) simplify build

saghul:
lua-cjson is now provided for arm64 by Debian so we no longer need to do deb package dissection.


prosody/Dockerfile
```diff
@@ -41,21 +40,15 @@ RUN set -x && \
...
       libsasl2-modules-ldap \
       lua-cyrussasl \
-      lua-inspect \
       lua-ldap \
-      lua-luaossl \
-      lua-sec \
       lua-unbound && \
-    apt-dpkg-wrap apt-get -d install -y jitsi-meet-prosody && \
-    dpkg -x /var/cache/apt/archives/jitsi-meet-prosody*.deb /tmp/pkg && \
-    rm /tmp/pkg/usr/share/jitsi-meet/prosody-plugins/mod_smacks.lua && \
```
damencho, line 51: We still need the remove, to use the prosody provided module.

saghul: Re-added the removal.

saghul:
Arg FML. Our post-inst script has no regard for non-interactive environments and tries to do a bunch of stuff it won't be able to.

PR #2038, opened 2025-02-28
fix(web) fix generated config syntax

saghul:
Arrow functions don't work on the JS interpreter which evaluates config.js on Android.


saghul:
Example of previously failing: https://github.com/jitsi/docker-jitsi-meet/actions/runs/13595798766/job/38012303929

```
 SyntaxError: parse error (line 77)
    at [anon] (/tmp/config.js:77) internal
    at [anon] (duk_js_compiler.c:3715) internal
error in executing file /tmp/config.js
```

That was without the 2nd commit. Now config is tested through Duktape, not Node, like it runs on Android.

Issue #2022, opened 2025-02-17
Failed to obtain a certificate from the Let's Encrypt CA.


ovchinnikov:
Hello!
I'm trying to set up Jitsi using the [Self-Hosting Guide](https://jitsi.github.io/handbook/docs/devops-guide/devops-guide-docker/). I followed all the instructions carefully, but when I start the service, I encounter the following error:

```
web-1      | [Mon Feb 17 07:09:59 UTC 2025] Extracting master.tar.gz
web-1      | [Mon Feb 17 07:09:59 UTC 2025] Installing to /opt
web-1      | [Mon Feb 17 07:09:59 UTC 2025] Installed to /opt/acme.sh
web-1      | [Mon Feb 17 07:09:59 UTC 2025] Installing alias to '/root/.profile'
web-1      | [Mon Feb 17 07:09:59 UTC 2025] Close and reopen your terminal to start using acme.sh
web-1      | [Mon Feb 17 07:09:59 UTC 2025] Installing cron job
web-1      | 13 17 * * * "/config/acme.sh"/acme.sh --cron --home "/config/acme.sh" > /dev/null
web-1      | [Mon Feb 17 07:09:59 UTC 2025] bash has been found. Changing the shebang to use bash as preferred.
web-1      | [Mon Feb 17 07:10:00 UTC 2025] OK
web-1      | [Mon Feb 17 07:10:00 UTC 2025] Install success!
web-1      | [cont-init.d] 09-update-acme: exited 0.
web-1      | [cont-init.d] 10-config: executing... 
web-1      | /opt /
web-1      | [Mon Feb 17 07:10:00 UTC 2025] Installing to /config/acme.sh
web-1      | [Mon Feb 17 07:10:00 UTC 2025] Installed to /config/acme.sh/acme.sh
web-1      | [Mon Feb 17 07:10:00 UTC 2025] Installing alias to '/root/.profile'
web-1      | [Mon Feb 17 07:10:00 UTC 2025] Close and reopen your terminal to start using acme.sh
web-1      | [Mon Feb 17 07:10:00 UTC 2025] Installing cron job
web-1      | 13 17 * * * "/config/acme.sh"/acme.sh --cron --home "/config/acme.sh" > /dev/null
web-1      | [Mon Feb 17 07:10:00 UTC 2025] bash has been found. Changing the shebang to use bash as preferred.
web-1      | [Mon Feb 17 07:10:01 UTC 2025] OK
web-1      | /
[...]

PR #2042, opened 2025-03-06
web: Added LETSENCRYPT_ACME_DIRECTORY environment variable

HighwayStar:
* you can set it when default zerossl doesnt work for you. For example yor server located in the country blocked by ZeroSSL like described in #2022

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -121,6 +121,7 @@ services:
             - LETSENCRYPT_DOMAIN
             - LETSENCRYPT_EMAIL
             - LETSENCRYPT_USE_STAGING
+            - LETSENCRYPT_ACME_SERVER
             - MATOMO_ENDPOINT
             - MATOMO_SITE_ID
             - MICROSOFT_API_APP_CLIENT_ID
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -121,6 +121,7 @@ services:
             - LETSENCRYPT_DOMAIN
             - LETSENCRYPT_EMAIL
             - LETSENCRYPT_USE_STAGING
+            - LETSENCRYPT_ACME_DIRECTORY
```
saghul, line 124: I think this variable name is very misleading. acme.sh may implement the feature using a directory, but that's not what the user really wants, they want to change the server. Let's call this LETSENCRYPT_ACME_SERVER.

web/rootfs/etc/cont-init.d/10-config
```diff
@@ -19,10 +19,18 @@ if [[ $DISABLE_HTTPS -ne 1 ]]; then
         if [[ $LETSENCRYPT_USE_STAGING -eq 1 ]]; then
             STAGING="--staging"
         fi
+
+        ACME_DIRECTORY=""
```
saghul, line 23: ACME_SERVER

saghul (commented):
Thanks for doing this! Left some naming comments.

saghul (approved):

saghul:
Thank you!

PR #2043, opened 2025-03-06
prosody: Added XMPP_PERSISTENT_CHATS environment variable

przemub:
When set to true, it will cause MUC component of prosody not to drop chat history when the last participant leaves.

Chat history will be also written to the disk instead of keeping it in memory, so it will also be kept after the container is restarted.

Useful for private conferencing rooms when future reference is needed.

Solution inspired by @saghul suggestion here:
https://github.com/jitsi/lib-jitsi-meet/issues/1049

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -329,6 +329,7 @@ services:
             - XMPP_MUC_CONFIGURATION
             - XMPP_INTERNAL_MUC_MODULES
             - XMPP_HIDDEN_DOMAIN
+            - XMPP_PERSISTENT_CHATS
             - XMPP_PORT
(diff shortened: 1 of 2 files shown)


damencho:
Have you tested this? This will just store some muc state on disk and restore it when you join second time in the same room, not sure it may store some history, but that will be just the last 20 messages.
There are some prosody community modules that actually does storing of chat history.

przemub:
Hi, thanks for pointing it out - it worked pretty well (the messages were re-sent to the joining client) until I restarted the server, then only the last message from before the restart was sent. Pretty interesting but obviously not good enough! Thanks for the suggestions as well.

PR #2049, opened 2025-03-17
FEAT Prosody mod_muc_moderation

arminfelder:
adds mod_muc_moderation , to allow moderators to redact illegal content using XEP-0425 once reported

https://modules.prosody.im/mod_muc_moderation


saghul:
What's your take here @damencho dhall we perhaps get the whole prosody-plugins repo?

damencho:
Well it's not food to have all modules from there, there are some that are for backward compatibility of 0.11 and 0.12 and names may clash with some modules in 13. 
Like smacks and external services are examples.

damencho:
Other than that, this looks ok.

damencho (approved):

PR #2050, opened 2025-03-18
feat(jvb): add ports to advertised ips

1fabi0:
I initially described my problem [here](https://github.com/jitsi/jitsi-videobridge/issues/2305)
>We are using a Kubernetes StatefulSet to host Jitsi video bridges. A load balancer acts as the ingress to the video bridges.
>
>From my perspective, it would be ideal if all video bridges could use the same UDP port internally, with the load balancer differentiating them by using different external ports. This approach would save me from having to allow all possible UDP port values in the video bridges' pod template.

This is my solution to this problem, it ensures old advertised IPs also remain functional.

I welcome any comments or suggestions.

Thanks in advance.

Best regards,
Fabian

### DIFF
--- a/jvb/rootfs/defaults/jvb.conf
+++ b/jvb/rootfs/defaults/jvb.conf
@@ -138,11 +138,22 @@ ice4j {
             }
             static-mappings = [
 {{ range $index, $element := $JVB_IPS -}}
+{{ if contains ":" $element -}}
+{{ $element_ip_port := splitn ":" 2 $element -}}
+                {
+                    local-address = "{{ $ENV.LOCAL_ADDRESS }}"
+                    public-address = "{{ $element_ip_port._0 }}"
+                    local-port = {{ $ENV.JVB_PORT | default 10000 }}
+                    public-port = {{ $element_ip_port._1 }}
+                    name = "ip-{{ $index }}"
+                },
+{{ else -}}
                 {
                     local-address = "{{ $ENV.LOCAL_ADDRESS }}"
                     public-address = "{{ $element }}"
                     name = "ip-{{ $index }}"
                 },
+{{ end -}}
 {{ end -}}
             ]
         }


aaronkvanmeerten:
How are you planning to guarantee that the JVB that jicofo selects/signals for a particular endpoint is the JVB that receives the UDP stream from your load balancer?  Is it that on the outside you have multiple ports, and differentiate the JVB by letting it know which port to advertise?  but on the inside of the container you wish to simply have JVB listen on 10000?

1fabi0:
@aaronkvanmeerten 
> Is it that on the outside you have multiple ports, and differentiate the JVB by letting it know which port to advertise? but on the inside of the container you wish to simply have JVB listen on 10000?

Exactly, currently I expose the whole range of possible ports for a container/pod and at start up of a statefulset I have a startup script/inline bash script that first adds the pod number to the JVB_PORT in future this script should set the port on the advertised ip script.

Regarding the required changes in the Helm chart there will be another pull request in the other repos later on, that builds on this change and will allow a scaleable jvb with loadbalancer services.

jvb/rootfs/defaults/jvb.conf
```diff
@@ -138,11 +138,22 @@ ice4j {
             }
             static-mappings = [
 {{ range $index, $element := $JVB_IPS -}}
+{{ if contains ":" $element -}}
```
saghul, line 141: This breaks for IPv6. We could use # as the delimiter. A bit weird to read but would also work for IPv4 and IPv6.

saghul: It should go without.

InDieTasten (approved):

jvb/rootfs/defaults/jvb.conf
1fabi0: Updated it, I know that invalid ip addresses are allowed with the provided regex but I think it is not part of the configuration to validate the addresses

saghul:
Not a fan of using regexes for this. @aaronkvanmeerten WDYT?

Sprout (which we use in the tempalting engine) supports parsing an IP address, but it doesn't seem to support splitting IP and port. https://docs.atom.codes/sprout/registries/network

Perhaps it would be worth it to add to tpl?

1fabi0:
@saghul 
I can understand this, I could also think of a sultion without regex e.g. 

``` tpl
{{ if contains "]:" $element -}}
{{ $element_ip_port := splitn "]:" 2 $element  -}}
                {
                    local-address = "{{ $ENV.LOCAL_ADDRESS }}"
                    public-address = "{{ trimPrefix "[" $element_ip_port._0 }}"
                    local-port = {{ $ENV.JVB_PORT | default 10000 }}
                    public-port = {{ $element_ip_port._1 }}
                    name = "ip-{{ $index }}"
                },
{{ else if and (contains "." $element) (contains ":" $element) -}}
{{ $element_ip_port := splitn ":" 2 $element -}}
                {
                    local-address = "{{ $ENV.LOCAL_ADDRESS }}"
                    public-address = "{{ $element_ip_port._0 }}"
                    local-port = {{ $ENV.JVB_PORT | default 10000 }}
                    public-port = {{ $element_ip_port._1 }}
                    name = "ip-{{ $index }}"
                },
{{ else -}}
                {
                    local-address = "{{ $ENV.LOCAL_ADDRESS }}"
                    public-address = "{{ $element }}"
                    name = "ip-{{ $index }}"
                },
{{ end -}}
```

saghul:
Honestly at that point I'd rather do the `IP#port` syntax since it's unambiguous.

aaronkvanmeerten:
> Honestly at that point I'd rather do the `IP#port` syntax since it's unambiguous.

I think upon reflection this is likely the right way despite it being non-standard.  We may look at other places where we split on `:` and review whether we should use `#` instead there as well.

1fabi0:
Ok, I will change it then, thanks 👍

jvb/rootfs/defaults/jvb.conf
```diff
@@ -138,11 +138,22 @@ ice4j {
             }
             static-mappings = [
 {{ range $index, $element := $JVB_IPS -}}
+{{ if contains "#" $element -}}
```
saghul, line 141: Can you add an example to env.example please?

saghul (approved):

saghul:
@aaronkvanmeerten are you still ok to land this?

aaronkvanmeerten:
sorry for the delay on this one, I moved continents and it fell off my radar.

PR #2057, opened 2025-03-22
fix(prosody): Fixes auth type internal with prosody 13.

damencho:
When that mechanism is enabled we end up in endless loop:
```
prosody-1  | 2025-03-22T21:01:26.925692034Z 2025-03-22 21:01:26 meet.jitsi:tokenauth                                         warn	Failed to verify access token: (nil)
prosody-1  | 2025-03-22T21:01:27.043438860Z 2025-03-22 21:01:27 meet.jitsi:tokenauth                                         warn	Failed to verify access token: (nil)
.....
```


saghul (approved):
How on earth did you figure this one out?!

damencho:
Well, I managed to repro on non docker, but I saw the endless cycle in prosody logs ... prosody code a little and here we are.

PR #2061, opened 2025-03-28
feat(prosody): use 0.12 nightly

aaronkvanmeerten:


damencho (approved):

saghul (approved):

azhebko:
Hi! Could you please tell the reason of downgrade from prosody13?

saghul:
We are experiencing some instability that we are looking into.

PR #2067, opened 2025-04-02
fix self-host screen share issues

Jacques-z:
I self-hosted [aae3756](https://github.com/jitsi/docker-jitsi-meet/releases/tag/stable-10133-1) version on my laptop, and cross-device screen share doesn't work, saying "Video for user's screen has been turned off to save bandwidth".

This is resolved after changing these default values.


saghul:
@jallamsetty1 whsts your take on this?

jallamsetty1:
> @jallamsetty1 whsts your take on this?

@Jacques-z I don't understand how changing the capture rate will have an affect downstream from the bridge to the endpoint? On the other hand, changing the defaults for max fps will affect the capture resolution for everyone. Maybe it was a transient network issue? Have you tested it couple of times with the defaults?  Even otherwise, you should be able to update the env variables on your deployment right?

Jacques-z:
@jallamsetty1 I just realized that I haven't even had set `DESKTOP_SHARING_FRAME_AUTO` false. So it does seem nothing to do with these two values.

BTW, I've tested several times that cross-device users can only view existing screen share when they join, which makes mutual screen share impossible. This doesn't happen with meet.jit.si. Anyway, that's another story and I'll try to debug on that.

PR #2096, opened 2025-05-26
fix(ci) avoid OOM when building with QEMU

saghul:
We seem to be close to the QEMU memory limit in GHA, that's why actions fail sporadically.

Switch to native builds, aka build amd64 in an amd64 machine and do the same for arm64.

This means we need to push 2 extra tags: foo-amd64 and foo-arm64 and then create a manifest to unite them both.

Ref: https://github.com/docker/build-push-action/issues/621


saghul:
We need this, despite the other PRs that just landed, since this fixes a different problem:

```
28.22 E: Could not read from /var/lib/apt/lists/partial/download.***.org_unstable_InRelease - getline (12: Cannot allocate memory)
```

PR #2099, opened 2025-05-27
fix(docker) fix Docker warnings

saghul:


saghul:
Unrelated build error, fixed in #2098

svigerske:
The removal of `version: '3.5'` broke the use of the `docker-compose.yml` for me:
```
ERROR: Invalid interpolation format for "web" option in service "services": "jitsi/web:${JITSI_IMAGE_VERSION:-unstable}"
```
I don't know enough about docker-compose to guess about the consequences of that change. Probably it is also my docker-compose being old (version 1.25.0).

But just wanted to mention it here, in case something simple could be done to make it compatible again. (I've just put `version: '3.5'` back.)

saghul:
Your compose version is really really old. You should update.

PR #2137, opened 2025-08-14
fix(log-analyser): pin otel/opentelemetry-collector-contrib image version

Taukon:
The Loki exporter, which is used by the log-analyser, has been removed from recent versions of the `otel/opentelemetry-collector-contrib` image. Using the `latest` tag breaks the logging functionality.

This PR pins the image to `otel/opentelemetry-collector-contrib:0.130.0` to ensure that a version with the Loki exporter is used.

Reference: 
- https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.131.0
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/41413

### DIFF
--- a/log-analyser.yml
+++ b/log-analyser.yml
@@ -17,7 +17,7 @@ services:
 
   otel-collector:
     container_name: otel
-    image: otel/opentelemetry-collector-contrib
+    image: otel/opentelemetry-collector-contrib:0.130.0
     user: "0" # required for reading docker container logs
     volumes:
       - ./log-analyser/otel-collector-config.yaml:/etc/otelcol-contrib/config.yaml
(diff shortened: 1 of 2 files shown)


saghul:
Doesn't us lock us into something that will not be maintained further?

saghul (approved):

Taukon:
> Doesn't us lock us into something that will not be maintained further?

Thank you for your feedback. I agree with your point about the maintenance risks of pinning to an old version.

I believe migrating to the `otlphttp exporter` is the right long-term solution. However, I've noticed this will also require changes to LogQL queries, so it might take some time to implement correctly.

Given this, would you prefer that I close this PR and create a new one for the full migration, or would you consider merging this as a temporary fix?

saghul:
Let's have this one for now, but I do like your plan to migrate!

PR #2142, opened 2025-08-22
refactor(jvb): add DISABLE_AWS_HARVESTER

emrahcom:
PR adds `DISABLE_AWS_HARVESTER` as environment variable to set/unset `ice4j.harvest.mapping.aws.enabled` in `jvb.conf`

Related with [this topic](https://community.jitsi.org/t/service-unavailable-and-failed-conference/140678) in the community forum.


jvb/rootfs/defaults/jvb.conf
```diff
@@ -137,6 +138,13 @@ jmt {
 ice4j {
     harvest {
         mapping {
+            aws {
+{{ if $DISABLE_AWS_HARVESTER -}}
```
saghul, line 142: Since it's a boolean, and we have a default value, why not set it unconditionally?

saghul (approved):

PR #2145, opened 2025-08-27
fix(grafana): enable data source provisioning to fix dashboards

Taukon:
In the current docker-compose setup, Grafana data sources were not being created automatically. This caused an issue where provisioned dashboards would fail to render graphs, showing a `Datasource <UID> was not found` error in the panels.

This PR resolves this issue by enabling Grafana's data source provisioning and updating the relevant dashboard configurations. This ensures that the Prometheus and Loki data sources are created automatically on container startup, allowing dashboards to correctly recognize them and display data reliably.

Changes
- Updated the existing `log-analyser/datasources/datasource_loki.yml` to assign static uids to the Prometheus and Loki data sources.
- Added a volume mount to `grafana.yml` to load the data source definition directory, enabling automatic data source creation.
- Updated `Prosody-Dashboard.json` and `prosody.json`.

### DIFF
--- a/grafana.yml
+++ b/grafana.yml
@@ -9,6 +9,7 @@ services:
     volumes:
       - ./log-analyser/grafana:/var/lib/grafana
       - ./log-analyser/grafana-provisioning/dashboards/:/etc/grafana/provisioning/dashboards/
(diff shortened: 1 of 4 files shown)


saghul:
Cheers!

PR #2159, opened 2025-09-30
fix(log-analyser): migrate from loki exporter to otlphttp exporter

Taukon:
This PR provides a permanent solution to the issue temporarily addressed in #2137. It migrates the log-analyser from the deprecated loki exporter to the recommended otlphttp exporter, ensuring it works with the latest `otel/opentelemetry-collector-contrib` images.

Changes:
-   Updated `otel-collector-config.yaml` to use the `otlphttp` exporter and added the `service.name` resource attribute for better indexing.
-   Updated Grafana dashboard JSON files with queries compatible with the OTLP log format.
-   Removed the pinned image version in `log-analyser.yml`.

### DIFF
--- a/log-analyser.yml
+++ b/log-analyser.yml
@@ -5,7 +5,7 @@ services:
 
   loki:
     container_name: loki
-    image: grafana/loki:3.0.0
+    image: grafana/loki
     command: -config.file=/conf/loki-config.yaml
     volumes:
       - ./log-analyser/loki/data:/data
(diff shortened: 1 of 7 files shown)


log-analyser.yml
```diff
@@ -5,7 +5,7 @@ services:
 
   loki:
     container_name: loki
-    image: grafana/loki:3.0.0
+    image: grafana/loki
```
saghul, line 8: Pin it to some version please.

log-analyser.yml
```diff
@@ -17,7 +17,7 @@ services:
 
   otel-collector:
     container_name: otel
-    image: otel/opentelemetry-collector-contrib:0.130.0
+    image: otel/opentelemetry-collector-contrib
```
saghul, line 20: Pin it sto some version please.

Taukon:
I've pinned the image versions. Please take a look.

saghul (approved):

PR #2184, opened 2025-12-04
jibri: bump Chrome release version in Dockerfile

saghul:


saghul:
@aaronkvanmeerten Looke like this was the cause for the recent buidl failure, I guess they stop serving old versions at one point.

PR #2187, opened 2025-12-08
Preventing default GC setting from eating all node RAM

alekseyl:
Having 400 as a default GC threshold ( in combination with speed equal 250 as a second default ) makes lua anable to recover memory at all! Eating out every single byte on node cluster:

image


alekseyl:
400 breaking through the roof almost instanlty: 
image

damencho:
We have being doing stress testing with 500 participants in a meeting and 10k for visitors to come to these values. The current settings use more RAM but makes GC to run less frequent which in stress times uses less CPU.

PR #2195, opened 2026-01-07
fix(prosody) pin version to avoid bug with ceratin room names

saghul:


saghul:
No luck, the prosody Debian repos don't seem to keep old versions around.

PR #2197, opened 2026-01-07
fix: Patch prosody 13.0.3 to disable domain verification.

bgrozev:
We allow `_` in tenant names, which get translated to
conferencename@conference.TANANT.meet.jitsi. With this change in 13.0.3
prosody throws errors as domains with "_" are not valid according to the
spec.
https://hg.prosody.im/trunk/rev/863dd118f8e8


damencho (approved):

emrahcom:
I think the last part of the added command (`|| true`) may cause some problem. This command will be always successful even one of the previous step fails.

PR #2198, opened 2026-01-10
Fix image pull failures by defaulting to stable Jitsi image tags

Gautam-aman:
### Problem
docker-jitsi-meet defaults to pulling Docker image tags that are no longer
published on Docker Hub, causing fresh installations to fail.

### Solution
- Default all services to a stable Jitsi image tag
- Preserve JITSI_IMAGE_VERSION override behavior
- Document the change for users

### Testing
- Clean install pulls stable-* images
- JITSI_IMAGE_VERSION override pulls unstable images
- Full docker-compose stack starts successfully

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,14 @@ Tag | Description
 `unstable-YYYY-MM-DD` | Daily unstable release
 `latest` | Deprecated, no longer updated (will be removed)
 
+> ⚠️ **Note**
+>
+> Docker Hub no longer publishes or updates the `latest` Jitsi images.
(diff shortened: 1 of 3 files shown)


saghul:
As per the docs, you should be using release tags, which do have the right versions.

Master defaults to unstable, for testing, and JITSI_IMAGE pointing to the nonexistent "latest" is there for development. If you build the images locally they will be tagged with latest.

Gautam-aman:
Understood, thanks for the clarification. I appreciate the explanation around release tags vs. latest and the intended behavior on master.

PR #2202, opened 2026-01-21
fix(prosody): Moves to using shell for roster creation.

damencho:


saghul:
What's the rationale here?

damencho:
That it fails to successfully create the focus roster. https://community.jitsi.org/t/room-creation-is-restricted/141695?u=damencho
The same we have seen using ansible on the shards.
I have started the changes 10 months ago in a branch docker-jitsi-meet/tree/move-to-shell but there was a bug in prosody which is now fixed in 13.0.3 and we can move to use shell, we did that to the ansible part.

PR #2212, opened 2026-01-29
feat(prosody): Validates the installed prosody version.

damencho:


aaronkvanmeerten:
Yes, this will also break on the next release of prosody. I think if we want to stick to a specific version we should be installing via URL.  Alternately, we should be checking for just the major version being 13 for now.  This will make us be able to upgrade.

I think installing alternate packages is going to be more complicated. Maybe we could put instructions on how to build an alternate container. But I think having our own opinionated container build makes the most sense instead of trying to be open ended for everybody.

PR #2217, opened 2026-02-09
Remove unused environment variables from templates

emrahcom:
Some templates contain unused environment variables. Probably because of removed codes. So, they are not needed anymore. This PR removes these unused envs to make codes more clean.


emrahcom:
Manual observation... I catched one of them while working on Prosody. I checked the others manually.

PR #2229, opened 2026-02-25
fix: enable Sentry logging support for jvb, jicofo, and jigasi

hluaces:
The logging.properties templates used toBool on the SENTRY_DSN variable, which is a URL, making it impossible to configure Sentry properly. The SENTRY_DSN default was also set to "0" which is truthy in Go templates when toBool is removed.

Additionally, the sentry-jul JAR (containing the SentryHandler class) was missing from all three Docker images, causing a ClassNotFoundException at startup.

I believe we are not trying to move this upstream (see https://github.com/jitsi/jitsi videobridge/pull/1785) but if we wanted to make this better we should be changing upstream so that the jul jars are also included.

This won't work as-is until the docker images are built and published. An override file like this can be used to force to use local images to test it:

```yaml
services:
    jicofo:
        build:
            context: ./jicofo
            args:
                BASE_TAG: unstable
    jvb:
        build:
            context: ./jvb
            args:
                BASE_TAG: unstable
```

For: 

- https://github.com/jitsi/docker-jitsi-meet/issues/2228

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -413,7 +413,7 @@ services:
             - MAX_BRIDGE_PARTICIPANTS
             - OCTO_BRIDGE_SELECTION_STRATEGY
             - PROSODY_VISITORS_MUC_PREFIX
-            - SENTRY_DSN="${JICOFO_SENTRY_DSN:-0}"
+            - SENTRY_DSN=${JICOFO_SENTRY_DSN:-}
             - SENTRY_ENVIRONMENT
             - SENTRY_RELEASE
             - TZ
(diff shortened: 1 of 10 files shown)


hluaces:
Managed to get it working and shipping messages:

image

damencho:
If we are downloading it here, can we drop it from here: https://github.com/jitsi/jitsi-videobridge/blob/4e32a93889406114ab0b5d9410f7471ee60fc1e0/jvb/pom.xml#L214?

hluaces:
> If we are downloading it here, can we drop it from here: https://github.com/jitsi/jitsi-videobridge/blob/4e32a93889406114ab0b5d9410f7471ee60fc1e0/jvb/pom.xml#L214?

Not the same thing. That's `sentry`, this downloads `sentry-jul`. Upstream packages the former, but doesn't do it for the latter, which this pr fixes here.  I believe we need both.

I'm not entirely sure on what upstream wants to do here. Provided that Sentry wants to be supported, the long term fix is to build `sentry-jul` in them. But after seeing https://github.com/jitsi/jitsi-videobridge/pull/1785 I'm not entirely sure if that's the case, so that's why I baked the jar file here.

If we prefer open PRs in the rest, it should be straightforward.

damencho:
What about adding and the other dependency here and getting rid of those dependencies in jicofo and jvb?

hluaces:
Probably doable. We'd need to download both artifacts here. Also worth noting jigasi also uses the dependency.

PR #2234, opened 2026-03-01
Add support for custom transcription languages config option

SSpirate11:
Added support for the [customLanguages](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) configuration option in the transcription settings, enabling Docker deployments to extend the list of supported transcription languages.

### DIFF
--- a/env.example
+++ b/env.example
@@ -115,6 +115,21 @@ TZ=UTC
 
 
 #
+# Transcription configuration
+#
+
+# Enable transcription (requires Jigasi)
+#ENABLE_TRANSCRIPTIONS=0
+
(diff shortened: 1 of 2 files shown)


zobadaniel:
I'd like to have this merged.

zobadaniel:
Please let me know what you need to have this merged. It's not a big diff and I don't see any side effects for the rest of the system, so it would be great to not patch every release manually.

SSpirate11:
@damencho  Can i get a review for this PR ?

damencho (approved):

PR #2257, opened 2026-06-15
base: update to Debian Trixie

saghul:
Migrate all Debian-based images from bookworm to trixie (Debian 13).

- base: debian:trixie-slim and trixie-backports
- base-java: openjdk-17 -> openjdk-21. Trixie dropped openjdk-17; the
  jitsi packages already declare openjdk-21 as a supported alternative.
- prosody: use the trixie suite from packages.prosody.im, and drop the
  dead 'rm -rf /usr/{lib,share}/lua/{5.1,5.2,5.3}' cleanup. The brace
  expansion never ran under /bin/sh (dash), and on trixie
  /usr/share/lua/5.4 is a symlink farm into 5.1/5.2, so actually
  deleting those trees would break the prosody install.
- base: drop apt-transport-https, a no-op transitional package in trixie.
- ci: prosody trixie suite in unstable.yml.

### DIFF
--- a/.github/workflows/unstable.yml
+++ b/.github/workflows/unstable.yml
@@ -28,7 +28,7 @@ jobs:
         run: echo 'APT::Key::Assert-Pubkey-Algo ">=rsa2048,ed25519,ed448,dsa1024";' | sudo tee /etc/apt/apt.conf.d/99weakkey-warning
       - name: Prosody repo
         id: prosody_repo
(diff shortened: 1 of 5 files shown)


saghul:
@damencho I did a local 3 way test and all works as expected AFAICT.

damencho:
Looks good to me, we have moved jibri and jigasi to support java 21 so everything should be fine.

damencho (approved):

PR #2258, opened 2026-06-16
feat: Rootless containers

saghul:
Run all Jitsi containers as an unprivileged "s6" user (uid/gid 1000)
instead of root, and upgrade s6-overlay from v1.22.1.0 to v3.2.0.2
(services.d/cont-init.d -> s6-rc.d).

- base/base-java: download s6-overlay v3 (noarch + arch tarballs), add
  the "s6" user, switch APT to deb822 .sources files, end with USER s6.
- All service containers migrated to /etc/s6-overlay/s6-rc.d with a
  config oneshot + longrun services; runtime config is rendered into
  /run/<service>/config from the read-only /config input.
- Writable runtime data split into dedicated volumes: prosody data in
  /var/lib/prosody (auto-migrated from /config/data), jibri recordings
  and logs in /storage, transcripts in storage/transcripts, web TLS state
  in storage/web, web cron/load-test in tmp/.
- jibri no longer requires the SYS_ADMIN capability; Chrome runs with
  --no-sandbox (added to every chrome flags branch).
- prosody: prosody/saslauthd/register-setup run as s6; existing accounts
  and certs are migrated on first start.

Adapted to the current tree (Debian Trixie, OpenJDK 21, prosody 13) and
fixed a few issues found while porting the original change:
- finish scripts forward the exit code via a bash wrapper (execlineb -P
  drops arguments), so graceful shutdown handling keeps working.
- --no-sandbox is rendered in all jibri chrome flag branches, including
  the default one.
- jicofo copies /config into /run so custom-jicofo.conf is honored.
- the config oneshot fails fast with an actionable message (check-writable)
  if a required writable volume is not writable by the container user, so
  a misconfigured host volume stops the container loudly instead of
  silently losing data.
- web persists its TLS material (self-signed keys and the Let's Encrypt
  account + issued certificates) and the JaaS marker in a dedicated
  storage/web volume, migrated from /config on first start, so they
  survive restarts; the bundled acme.sh is also made readable by the
  unprivileged user (it shipped 0600 root).

Backwards-incompatible changes
==============================

You can keep an existing ~/.jitsi-meet-cfg directory, but note:

- Unprivileged user. The processes run as uid/gid 1000 with no effective
  Linux capabilities. They can bind ports 80/443 (allowed on the default
  Docker configuration) but cannot write to host directories owned by
  another user. Any bind mount the containers write to must be writable
  by uid 1000.

- New writable directories - create these before starting. If Docker
  auto-creates a missing bind-mount source it does so as root:root, which
  the uid-1000 process cannot write to, so create them yourself and make
  them writable by uid 1000 (chmod 777, or chown -R 1000:1000):

    mkdir -p ~/.jitsi-meet-cfg/storage/{jibri,prosody,transcripts,web}
    mkdir -p ~/.jitsi-meet-cfg/tmp/{web-crontabs,web-load-test}
    chmod 777 ~/.jitsi-meet-cfg/storage/{jibri,prosody,transcripts,web} \
              ~/.jitsi-meet-cfg/tmp/{web-crontabs,web-load-test}

  Containers that need these will refuse to start (with a clear message)
  if the directory is missing or not writable by uid 1000.

- Moved paths (docker-compose.yml / jibri.yml / transcriber.yml are
  already updated):
    * Prosody data: /config/data -> /var/lib/prosody
      (${CONFIG}/storage/prosody). Existing accounts and certificates are
      copied over automatically on first start.
    * Jibri recordings and logs: /config/recordings and /config/logs ->
      /storage/recordings and /storage/logs (${CONFIG}/storage/jibri).
      The JIBRI_RECORDING_DIR default changed accordingly.
    * Transcripts: ${CONFIG}/transcripts -> ${CONFIG}/storage/transcripts.
    * Web TLS material (self-signed keys, Let's Encrypt account + certs)
      and JaaS marker -> /storage (${CONFIG}/storage/web).
    * Web cron/load-test: ${CONFIG}/web/crontabs and
      ${CONFIG}/web/load-test -> ${CONFIG}/tmp/web-crontabs and
      ${CONFIG}/tmp/web-load-test.

- Read-only /config, runtime config in /run. Each container treats its
  mounted /config as read-only input and renders the effective config
  into /run/<service>/config on every start. custom-* override files in
  /config are still picked up; the rendered files no longer appear in the
  mounted config directory.

- Web TLS material is persisted in ${CONFIG}/storage/web and survives
  restarts. Self-signed keys and the Let's Encrypt account + issued
  certificates are kept there; material from older images under
  ${CONFIG}/web (keys, acme.sh, acme-certs) is migrated on first start. A
  certificate supplied by the operator in ${CONFIG}/web/keys still takes
  precedence and is copied into the persistent volume.

- Jibri no longer needs SYS_ADMIN. The cap_add: SYS_ADMIN entry was
  removed from jibri.yml; Chrome is launched with --no-sandbox instead.

- Autoscaler sidecar key. When using AUTOSCALER_URL, the ASAP signing key
  must be readable by uid 1000 (the sidecar no longer runs as a dedicated
  user).

- Timezone. The TZ-based /etc/localtime symlink step was removed (an
  unprivileged user cannot write /etc). Applications still honour the TZ
  environment variable directly.

Co-authored-by: Saúl Ibarra Corretgé <s@saghul.net>


damencho (approved):

saghul:
Let's go, the only way forward is through!

Huge thanks @emrahcom for all the work here! ❤️

PR #2261, opened 2026-06-20
refactor(web): use unprivileged ports 8000 and 8443

### LINKED ISSUE #2260: Proposal to use unprivileged ports 8080 and 8443 by default for the web image to support non-root execution in Kubernetes.
Binding to default ports 80/443 requires root privileges in Linux. Changing the defaults to 8080/8443 allows the container to run securely under a non-root account out of the box in Kubernetes. What do you think?

emrahcom:
PR updates the ports of the web image to run it in Kubernetes under a non-root account.

Closes https://github.com/jitsi/docker-jitsi-meet/issues/2260

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,8 +8,8 @@ services:
             - /run:size=16M,mode=1750,exec
             - /tmp:size=16M,mode=1777,noexec
         ports:
-            - '${HTTP_PORT}:80'
-            - '${HTTPS_PORT}:443'
+            - '${HTTP_PORT}:8000'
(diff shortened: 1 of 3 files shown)


saghul:
Ah damn I just remembered something.

The outside ports must be 80/443 for Let's Encrypt to work.

I think we're good here, right?

saghul (approved):

emrahcom:
Yes, the outside ports (`HTTP_PORT` and `HTTPS_PORT`) can be 80/443. These updates are only for the internal ports.

PR #2266, opened 2026-06-22
feat(jicofo): map JIBRI_BUSY_RETRIES env vars to jicofo.conf

aaronkvanmeerten:
## What

Maps two new optional env vars in the jicofo `jicofo.conf` template's `jibri` block:

| Env var | jicofo.conf key |
|---|---|
| `JIBRI_BUSY_RETRIES` | `busy-retries` |
| `JIBRI_BUSY_RETRY_DELAY` | `busy-retry-delay` (rendered as `"<n> seconds"`) |

Both are guarded with `{{ if .Env.* }}`, so when unset jicofo falls back to its built-in default (`busy-retries = 0`, i.e. off). Mirrors the existing `JIBRI_REQUEST_RETRIES` → `num-retries` mapping right above it.

## Why

When all Jibri instances are momentarily busy (e.g. a demand spike while the autoscaler spins up more, or two requests racing for the last free instance), jicofo currently fails the recording start immediately. The new jicofo `busy-retries` feature (jitsi/jicofo#1288) lets it wait briefly for the pool to free up. This change lets deployments configure it via env vars.

## Related
- jicofo: jitsi/jicofo#1288

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -425,6 +425,8 @@ services:
             - JICOFO_TRUSTED_DOMAINS
             - JIBRI_BREWERY_MUC
             - JIBRI_REQUEST_RETRIES
+            - JIBRI_BUSY_RETRIES
+            - JIBRI_BUSY_RETRY_DELAY
             - JIBRI_PENDING_TIMEOUT
             - JIGASI_BREWERY_MUC
             - JIGASI_SIP_URI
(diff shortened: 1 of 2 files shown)


aaronkvanmeerten:
Review fix (commit 2dfa7d5): render `busy-retry-delay` verbatim instead of appending " seconds". `JIBRI_PENDING_TIMEOUT` already carries its unit in the value (default "90 seconds"); following that convention avoids a "5 seconds seconds" HOCON parse failure if an operator sets the value with a unit. The unit is now supplied by the producer (see infra-provisioning#1098).

saghul (approved):

PR #2274, opened 2026-06-29
Grafana on sub uri

amg-web:
This PR is to allow running Grafana on `/monitoring` suburi
extra die is required: ${CONFIG}/grafana
Grafana should be able to write to
`chmod 777 ${CONFIG}/grafana`
or 
`chown -R 472:472 ${CONFIG}/grafana `
472 is grafana pid

Also need to activate the respective vars in configs

Grafana version is 116.16, as newer version should have some fixes in dashboards

### DIFF
--- a/env.example
+++ b/env.example
@@ -243,3 +243,19 @@ JIBRI_XMPP_PASSWORD=
 
 # Jitsi image version (useful for local development)
 #JITSI_IMAGE_VERSION=latest
+
+# Grafana
+
+# Set servers external URL
+#GF_SERVER_ROOT_URL= '${PUBLIC_URL}/monitoring'
+
+# Set defaoult admin user
+#GF_SECURITY_ADMIN_USER='admin'
+
+# Set provisioning grafana password (can be changed later)
+#GF_SECURITY_ADMIN_PASSWORD='new-admin-password'
+
+# Set security settings
+#GF_SECURITY_CONTENT_SECURITY_POLICY='true'
+#GF_SECURITY_COOKIE_SECURE='true'
+#GF_SECURITY_COOKIE_SAMESITE='true'
--- a/env.example
+++ b/env.example
@@ -249,6 +249,7 @@ JIBRI_XMPP_PASSWORD=
 # Enable Grafana configs
 #GRAFANA=1
 
+# Set Grafana sub uri
 #GRAFANA_SUB_URI="_monitoring"
 
 # Set defaoult admin user
--- a/env.example
+++ b/env.example
@@ -246,8 +246,10 @@ JIBRI_XMPP_PASSWORD=
 
 # Grafana
 
-# Set servers external URL
-#GF_SERVER_ROOT_URL= '${PUBLIC_URL}/monitoring'
+# Enable Grafana configs
+#GRAFANA=1
+
+#GRAFANA_SUB_URI="_monitoring"
 
 # Set defaoult admin user
 #GF_SECURITY_ADMIN_USER='admin'
--- a/env.example
+++ b/env.example
@@ -246,17 +246,17 @@ JIBRI_XMPP_PASSWORD=
 
 # Grafana
 
-# Enable Grafana configs
+# Enable Grafana configs if required
 #ENABLE_GRAFANA=1
 
-# Set Grafana sub uri
+# Set Grafana sub uri (required if ENABLE_GRAFANA is enabled)
 #GRAFANA_SUB_URI="_monitoring"
 
 # Set defaoult admin user
 #GF_SECURITY_ADMIN_USER='admin'
 
 # Set provisioning grafana password (can be changed later)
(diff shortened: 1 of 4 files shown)


web/rootfs/defaults/meet.conf
```diff
@@ -277,3 +277,21 @@ location @root_path {
         rewrite ^/([^/?&:'"]+)/(.*)$ /$2;
     }
 {{ end }}
+
+{{ if .Env.GF_SERVER_ROOT_URL }}
+# Grafana settings
+location ^~ /monitoring/ {
```
aaronkvanmeerten: This value also needs to be reflected within the GF_SERVER_ROOT_URL above.  Perhaps instead this can also be configurable?

web/rootfs/defaults/meet.conf
```diff
@@ -277,3 +277,21 @@ location @root_path {
...
+    proxy_set_header Host $http_host; 
+    proxy_pass http://grafana.meet.jitsi:3000/;
+  }
+
+  # Proxy Grafana Live WebSocket connections.
+location ^~ /monitoring/api/live {
```
aaronkvanmeerten: Similarly, I expect that this would need to be made configurable

web/rootfs/defaults/meet.conf
amg-web: Value is taken from env. I think I missed main docker compose variable

saghul:
> Grafana version is 116.16, as newer version should have some fixes in dashboards

Please don't mix this with the rest of the changes.

web/rootfs/defaults/meet.conf
aaronkvanmeerten: My point was: this forces us to always use /monitoring, which may not be the intention of the end user.  They could potentially want `monitoring.example.com/` for instance, or `/grafana` or `/_gf` or some other path.  If that's not possible to accomplish easily, then we can force /monitoring, but perhaps we can consider making it flexible.

amg-web:
> > Grafana version is 116.16, as newer version should have some fixes in dashboards
> 
> Please don't mix this with the rest of the changes.

I do not remember from which grafana version nginx config changed
I tested with version 11.6

saghul:
> > > Grafana version is 116.16, as newer version should have some fixes in dashboards
> > 
> > 
> > Please don't mix this with the rest of the changes.
> 
> I do not remember from which grafana version nginx config changed I tested with version 11.6

Make a sewparate PR with the Grafana image bump please.

amg-web:
> > > > Grafana version is 116.16, as newer version should have some fixes in dashboards
> > > 
> > > 
> > > Please don't mix this with the rest of the changes.
> > 
> > 
> > I do not remember from which grafana version nginx config changed I tested with version 11.6
> 
> Make a sewparate PR with the Grafana image bump please.

version reverted.
you can do version change yourself.
grafana v12 should work too. grafana 13 will work too.
any newer version will update dashboards itself. 
but v13 has biggest changes.

I will not do any PR about version change

amg-web:
is anything else needed?

saghul:
GF_SERVER_ROOT_URL is used as a boolean, but it really isn't. Also, I agree with @aaronkvanmeerten here, exposing this might not be what others want to do and right now it's all or nothing.

Perhaps prefixing it with _ is the right thing to do, since that cannot be used as a meeting name.

amg-web:
updated

web/rootfs/defaults/meet.conf
```diff
@@ -277,3 +277,21 @@ location @root_path {
         rewrite ^/([^/?&:'"]+)/(.*)$ /$2;
     }
 {{ end }}
+
+{{ if .Env.GRAFANA }}
```
saghul: Use toBool, so setting it to zero or false doesn't turn it on

web/rootfs/defaults/meet.conf
```diff
@@ -277,3 +277,21 @@ location @root_path {
...
+# Grafana settings
+location ^~ /{{ .Env.GRAFANA_SUB_URI }}/ {
```
saghul, line 283: Check if this is set above too, with `and`

env.example
```diff
@@ -243,3 +243,22 @@ JIBRI_XMPP_PASSWORD=
 
 # Jitsi image version (useful for local development)
 #JITSI_IMAGE_VERSION=latest
+
+# Grafana
```
saghul, line 247: Please document these in the handbook, we stopped adding everything to the sample file because it can feel overwhelming otherwise.

env.example
```diff
@@ -243,3 +243,22 @@ JIBRI_XMPP_PASSWORD=
...
+
+# Enable Grafana configs
+#GRAFANA=1
```
saghul: ENABLE_GRAFANA?

saghul (changes requested):

amg-web:
added PR to handbook

PR #2287, opened 2026-07-31
feat: add optional Jibri recording library

imanYazizi:
## Summary\n- add an optional, read-only recording-library sidecar for completed Jibri MP4 files\n- provide a lightweight page with playback and download links\n- bind the service to loopback only and document authenticated reverse-proxy deployment\n\n## Usage\n\n```bash\ndocker compose -f docker-compose.yml -f jibri.yml -f recordings.yml up -d --build\n```\n\nThe sidecar reads the default `/storage/jibri/recordings` path and does not write to it. It is deliberately not exposed directly to the Internet.\n\n## Validation\n- `docker compose -f docker-compose.yml -f jibri.yml -f recordings.yml config --quiet`\n- `node --check recordings/site/app.js`\n\nA local Docker image smoke test could not run because the local Docker daemon is unavailable.

### DIFF
--- a/env.example
+++ b/env.example
@@ -24,6 +24,10 @@ HTTP_PORT=8000
 # Exposed HTTPS port
 HTTPS_PORT=8443
 
+# Loopback-only port for the optional Jibri recording library (recordings.yml)
+# Put this behind an authenticated reverse proxy before exposing it to users.
+#RECORDINGS_PORT=8090
+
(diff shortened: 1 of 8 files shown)


saghul:
Why does this need to live here? It can be a standalone repo with an extra yaml file and you can PR the handbook how to set it up.

Issue #187, opened 2020-02-09
How do you change the logo?


0187773933:
https://github.com/jitsi/jitsi-meet/issues/797

I can't find any of the html or css files after installing via docker

0187773933:
Modifying interface_config.js didn't change anything. Even after 
docker-compose up -d --force-recreate

saghul:
You need to change the interface_config.js in the $CONFIG directory.

jabiertxof:
how you do it?

saghul:
Oops, I was wrong. You need to replace the /usr/share/jitsi-meet/images/watermark.png file in the web container.

jabiertxof:
thanks also found in my home folder the hidden $CONFIG directory, for help configuring further people

ovimano:
You can create a new volume in docker-compose.yml for the png.

volumes:
               - .....
               - .......
               - /home/user/watermark.png:/usr/share/jitsi-meet/images/watermark.png

And recreate the containers.

Put the png in /home/user/watermark.png and that's it!

jabiertxof:
thanks!

TheoOliveira:
I've tried to add like you guys did, but it doesn't work. What it could be? Says image not found. Not sure why.

JacBra:
I think there was a change in interface_config.js:

    DEFAULT_LOGO_URL: 'images/watermark.png',
    DEFAULT_WELCOME_PAGE_LOGO_URL: 'images/watermark.png',

was changed to 

    DEFAULT_LOGO_URL: 'images/watermark.svg',
    DEFAULT_WELCOME_PAGE_LOGO_URL: 'images/watermark.svg',

so by default you now need an svg file (or change it in interface_config.js back to png)

hypocinose:
Hi!
I had the same problem and I changed the config file to
```
DEFAULT_LOGO_URL: 'images/mywatermark.png',
DEFAULT_WELCOME_PAGE_LOGO_URL: 'images/mywatermark.png',
```
and copied mywatermark.png into the container using docker cp
That worked quite well, but the new watermark is extremely tiny, because there is a size limit on it. You can see that when you inspect that element in your browser:
```
background-image: url("images/mywatermark.png"); max-width: 140px; max-height: 70px;
```
Where does that come from?
BR

TonyMugwump:
[...]

Issue #78, opened 2019-03-15
Can't able to view video conference for more than two persons 


remin23:

saghul:
Aare you running your setup on the local LAN or on a public server? Did you set DOCKER_HOST_ADDRESS?

remin23:
running on LAN. nope, not set any host address...

saghul:
Then please read the documentation. It's excplicitly mentioned.

remin23:
i have set the DOCKER_HOST_ADDRESS. Now it seems fine on LAN... Thanks for your info!!!

HichamBali:
Hi, It's running in a LAN when i set a Docker Host Adress, but can't access from outside(Public) when publish it on the firewall

PR #59, opened 2019-02-07
jigasi,web: add transcriber options

netaskd:
The PR adds opportunity to enable transcribing in a conference.

PR #2299, opened 2026-08-06
feat(jibri) add experimental opt-in Wayland capture

saghul:
Adds a second capture pipeline alongside the existing one: a headless Cage
compositor (wlroots, software rendering) with Chrome on Ozone/Wayland, and
wf-recorder capturing via wlr-screencopy instead of ffmpeg polling an Xorg
dummy framebuffer with x11grab.

Off by default. Set JIBRI_USE_WAYLAND=true to opt in; both paths ship in one
image and the inactive side's s6 service parks, so the two can be A/B'd on the
same host. X11 remains the default until this is validated against a real
meeting.

jibri-wf-capture translates args to wf-recorder and speaks back just enough of
ffmpeg's output for Jibri's health checks. That check is a state machine with
no stall timeout: a capturer that is alive but frozen would stay Running
forever, so failure can only be signalled by exiting. Hence the shim carries a
stall watchdog that kills a wedged capture, and wf-recorder's stderr is
forwarded to stdout so RTMP errors reach Jibri's BadRtmpUrl classifier rather
than being swallowed. Stopping uses SIGTERM, not SIGINT, because bash makes
background jobs ignore SIGINT and a swallowed signal leaves an mp4 with no moov
atom; a bounded wait escalates to SIGKILL so the shim cannot hang.

Damage tracking is decided by mode. It is the main reason this is cheaper than
x11grab, but on a static screen wlr-screencopy emits zero frames rather than
merely fewer, and --framerate cannot backfill them. A recording tolerates the
resulting VFR; RTMP ingest does not, so streaming passes --no-damage.

### DIFF
--- a/jibri.yml
+++ b/jibri.yml
@@ -40,6 +40,9 @@ services:
             - JIBRI_RECORDING_QUEUE_SIZE
             - JIBRI_RECORDING_STREAMING_MAX_BITRATE
             - JIBRI_USAGE_TIMEOUT
+            - JIBRI_USE_WAYLAND
+            - JIBRI_WAYLAND_OUTPUT
+            - JIBRI_WAYLAND_STALL_TIMEOUT
             - JIBRI_XMPP_USER
             - JIBRI_XMPP_PASSWORD
             - JIBRI_XORG_HORIZ_SYNC
--- a/jibri/Dockerfile
+++ b/jibri/Dockerfile
@@ -20,6 +20,7 @@ COPY rootfs/ /
 RUN \
   apt-dpkg-wrap apt-get update && \
   apt-dpkg-wrap apt-get install -y -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" \
+    cage \
     dbus \
     dbus-x11 \
     fonts-noto \
@@ -35,6 +36,8 @@ RUN \
     rclone \
     rtkit \
     unzip \
+    wf-recorder \
+    wlr-randr \
     && \
   /usr/bin/install-chrome.sh && \
   apt-get purge -y \
@@ -69,6 +72,22 @@ RUN \
   esac && \
   wget -qO /usr/bin/shm-check https://github.com/saghul/shm-check/releases/download/v1.0.0/shm-check-${SC_ARCH} && \
   chmod +x /usr/bin/shm-check
+
+# Wayland capture path (JIBRI_USE_WAYLAND). These live here rather than in the
+# config script because s6 oneshots cannot export into sibling services, and
+# Chrome only inherits them by being a grandchild of the jibri service.
+# The headless wlroots backend needs no DRM device, seat or logind, which is
+# what makes this work in a rootless container with no capabilities.
+# WAYLAND_DISPLAY must match what cage actually creates: it calls
+# wl_display_add_socket_auto(), which takes the first free name in a freshly
+# mounted XDG_RUNTIME_DIR, i.e. wayland-0. Chrome and wf-recorder are not cage
+# children so they do not inherit the name from it -- if this disagrees, they
+# silently fail to connect and capture falls back to a blank 1280x720.
+ENV XDG_RUNTIME_DIR=/run/tmp/xdg \
+    WAYLAND_DISPLAY=wayland-0 \
+    WLR_BACKENDS=headless \
+    WLR_LIBINPUT_NO_DEVICES=1 \
+    WLR_RENDERER=pixman
 
 VOLUME /config
 
--- a/jibri/rootfs/defaults/jibri.conf
+++ b/jibri/rootfs/defaults/jibri.conf
@@ -10,6 +10,8 @@
 {{ $JIBRI_SINGLE_USE_MODE := .Env.JIBRI_SINGLE_USE_MODE | default "false" -}}
 {{ $STATSD_HOST := .Env.JIBRI_STATSD_HOST | default "localhost" -}}
 {{ $STATSD_PORT := .Env.JIBRI_STATSD_PORT | default "8125" -}}
+{{ $USE_WAYLAND := .Env.JIBRI_USE_WAYLAND | default "false" | toBool -}}
+{{ $WAYLAND_OUTPUT := .Env.JIBRI_WAYLAND_OUTPUT | default "HEADLESS-1" -}}
 
 jibri {
     // A unique identifier for this Jibri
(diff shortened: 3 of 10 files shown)


aaronkvanmeerten:
I did a deep review of this comparing the Wayland pipeline against the existing Xorg/ffmpeg path, verifying the PR's assumptions against the jibri source (`OutputParser.kt`, `FfmpegStatusStateMachine.kt`, `ProcessWrapper.kt`, `JibriSubprocess.kt`, `reference.conf`) and against the wf-recorder/cage/wlroots versions actually shipped in Debian trixie. Overall this is well-engineered for an experimental opt-in — genuinely off by default, X11 path untouched — and the claims it makes about Jibri's health-check machinery are accurate. The risk is concentrated in **streaming mode**, plus a few watchdog blind spots and silent-degradation paths in recording mode.

## Assumptions I verified as correct

- `jibri.ffmpeg.ffmpeg-binary`, `command-linux-recording`, and `command-linux-streaming` exist in jibri's `reference.conf`, and Jibri appends the sink as exactly one trailing argument. The substitutions the template relies on but doesn't define (`gop-size`, `streaming-buf-size`, `recording-format`, `streaming-format`) all resolve from reference.conf defaults (60, 5952k, mp4, flv).
- The health check really has no stall detection: any whole line of `key=value` tokens classifies as ENCODING (field extraction is literally commented out in `OutputParser.kt`, so the shim's fabricated `frame=`/`fps=` values are never inspected), and a live-but-frozen capturer stays `Running` forever. There is also no `StartingUp` timeout. The stall watchdog addresses a real gap.
- Jibri stops the capturer with `kill -s SIGINT`, waits 10s, then `destroyForcibly` (SIGKILL). The trap on INT/TERM, the `Exiting normally, received signal 2.` line (regex: `Exiting.*signal\s*2`), and the 8s internal escalation all fit inside that window. Exit code is logged but ignored by Jibri.
- Debian trixie ships wf-recorder 0.5.0 (has every flag used here: `--framerate`, `--overwrite`, `--codec-param`, `--muxer`, `--no-damage`, `--audio=dev`) and cage 0.2.0 (implements wlr-output-management, so `wlr-randr --custom-mode` works).
- The reasoning about `--framerate` being an input-driven fps *filter* that cannot fabricate frames during a zero-damage gap (it backfills in a burst when the next frame arrives) is exactly right per wf-recorder 0.5.0 source (`frame-writer.cpp`), and `--no-damage` for streaming is the correct mitigation.

⚠️ **Dependency floor worth a comment**: none of this works on bookworm — wf-recorder 0.3.0 lacks `--framerate`/`--overwrite` (and *interactively prompts* on existing files), and cage 0.1.4 doesn't implement wlr-output-management at all, so `wlr-randr` is a no-op there. The feature is silently trixie-only.

## Highest-risk area: RTMP streaming

1. **`BadRtmpUrl` will likely never be classified.** Jibri's classifier requires a whole line matching `rtmp://.*Input/output error` — that format is printed by the *ffmpeg CLI's* error handler (`print_error` in cmdutils), not by libavformat. wf-recorder uses libav as a library and emits its own error text, so a bad stream key surfaces as `QuitUnexpectedly` (retriable) instead of `BadRtmpUrl` (non-retriable, `shouldRetry() = false`) — Jibri may retry a permanently broken URL. Since the shim already synthesizes ffmpeg output for the happy path, it should synthesize this line too when wf-recorder dies in streaming mode.
2. **The stall watchdog does not cover streaming at all.** `dest_size()` returns empty for an `rtmp://` destination, so the watchdog never arms. A wf-recorder that is alive but wedged mid-stream (blocked socket, frozen encoder) stays `Running` forever — the exact failure mode the shim was built to prevent, unguarded in the mode most likely to hit it.
3. **wf-recorder's RTMP path is barely exercised upstream.** It mechanically works (`avio_open` on the URL with `-m flv`), but the upstream tracker has a history of streaming problems (ammen99/wf-recorder#60, #40, #187), and wf-recorder does no real-time pacing or packet flushing.
4. **Audio differences vs the ffmpeg command.** Jibri's ffmpeg uses AAC 44.1 kHz @ 128 kbps with `-af aresample=async=1` continuously healing drift. wf-recorder 0.5.0 defaults to AAC 48 kHz with *no explicit bitrate* and no async resampling, and has a documented A/V desync history under encode pressure (#59, #78, #144). 48 kHz AAC is fine for YouTube; long sessions under CPU pressure are the drift scenario.

## Recording-mode risks

- **The watchdog only arms after the first byte exists.** Until the file is created, `dest_size()` returns empty and that case is treated as "streaming, no file to measure" — a wf-recorder that starts but never opens the output emits healthy encoding lines forever, and Jibri has no startup timeout to catch it.
- **Audio-driven file growth masks video wedges.** The file grows from audio packets alone, so a capture whose video side wedged passes the growth check indefinitely. (Rough parity with x11grab recording a frozen screen, but the watchdog gives a false sense of coverage.)
- **VFR output is a downstream-compatibility change.** Damage-driven mp4s play fine, but anything in a finalize/post-processing chain that assumes CFR (naive ffprobe fps math, some editors/transcoders, fixed-interval thumbnailing) can mishandle long static gaps. Worth testing against real post-processing pipelines.
- **H.264 profile drift.** The jibri recording command pins `-profile:v main -level 3.1`; the shim doesn't, so libx264 emits High profile with auto level. Fine for modern players, potentially not for older decoders downstream.
- **`--resolution` is parsed but never used.** Actual capture size is whatever the cage output happens to be. If the backgrounded `wlr-randr` fails (it logs and *continues*), every recording silently comes out 1280x720 regardless of `JIBRI_RECORDING_RESOLUTION`. In the Xorg path the modeline plus ffmpeg `-s` make resolution a hard guarantee. A wrong-resolution recording is arguably worse than a failed one — suggest verifying the output mode and failing loudly on mismatch.

## Process-lifecycle risks

- **SIGKILL orphans wf-recorder.** If Jibri's `destroyForcibly` ever fires, SIGKILL hits the *shim*; wf-recorder reparents to s6 init and keeps running — holding the Pulse source and screencopy session while a subsequent session starts a second wf-recorder. The ffmpeg path has no such layer. The 8s bound usually wins the race, but consider a process group or a defensive `pkill` before starting a new capture.
- **Startup readiness is best-effort.** cage is a longrun with no readiness notification, so "cage is up" means the process exists, not that `wayland-0` exists. JVM startup time wins this race in practice, but a crash-looping cage produces confusing chromedriver failures rather than a clear "compositor not ready". s6 `notification-fd` with a socket poll would make it deterministic.
- **Chrome dialogs render centered over the recording.** cage centers secondary toplevels on top of the primary view (`view.c`) — a permission/crash-restore dialog sits mid-frame until dismissed. Roughly parity with icewm, but no WM scripting escape hatch.

## Resource consumption vs Xorg/ffmpeg

Encoding dominates both paths and is identical (same libx264 preset/CRF/GOP/maxrate). The differences:

| Aspect | Xorg + x11grab | cage + wf-recorder |
|---|---|---|
| Capture | Full-frame `XShmGetImage` memcpy at fixed 30 fps, always | shm screencopy on damage (recording); compositor-paced with `--no-damage` (streaming) |
| Compositing | X server + icewm | cage with pixman (CPU) per frame |
| Idle/static content | Full cost regardless | Near-zero video cost (recording) |
| Memory | Xorg dummy reserves 192 MB VideoRam + server + icewm | cage/pixman is small; shim is a bash process |

- **Recording a live meeting**: modest savings. Active meeting video means near-continuous damage, so the damage-tracking advantage mostly evaporates during video; the real wins are the removed Xorg/icewm overhead (likely 100–200 MB RSS) and the static portions (lobby, static screenshare).
- **Streaming can cost *more* than x11grab.** `--no-damage` re-requests frames as fast as the compositor presents them; the headless output defaults to 60 Hz, so you'd composite and copy at 60 fps and discard half through the `fps=30` filter — versus ffmpeg's clean 30 fps poll. The `--custom-mode WxH@${FRAMERATE}Hz` attempt is what neutralizes this, but the fallback path (size-only → 60 Hz) silently doubles capture work. Worth logging/asserting the effective refresh rate.
- **Chrome is a wash**: software raster either way; `--in-process-gpu` saves one process.

## Buffer underrun/overrun

The ffmpeg pipeline is meaningfully more battle-hardened here:

- **ffmpeg**: `-thread_queue_size 4096` buffers both inputs against encoder hiccups (overrun → visible warnings + graceful drops), and `aresample=async=1` heals accumulated drift. Note **`JIBRI_RECORDING_QUEUE_SIZE` is silently ignored in Wayland mode** — no equivalent knob exists.
- **wf-recorder**: a small frame queue between screencopy and encoder; under encode pressure the documented failure mode is A/V desync rather than clean drops. Audio is a blocking `pa_simple_read` on its own thread — if the writer stalls, the Pulse buffer overruns into audio gaps. No drift-healing resampler.
- **Underrun on static content**: recording tolerates zero-frame gaps by design (VFR); `--no-damage` is the right call for RTMP.

## Smaller notes

- `jibri.yml` exposes `JIBRI_WAYLAND_STALL_TIMEOUT` but not `JIBRI_WAYLAND_STOP_TIMEOUT`, which the shim also reads.
- The shim logs its full argv (including the RTMP URL with stream key) to `/storage/logs/jibri-wf-capture.log` and appends forever with no rotation. Jibri itself also logs the command, so parity on secrecy, but the unbounded cross-session log is new.
- In the `CHROMIUM_FLAGS` branch the ozone flags are appended after user flags, so a user-supplied ozone flag gets duplicated (last one wins; benign).
- `expand_rate` using k=1000 matches ffmpeg's SI convention — correct.

## Suggested changes

1. Add a streaming-mode stall detector (current one is file-growth-only, inert for RTMP).
2. Arm the recording watchdog from process start ("no output file after N seconds" = stall).
3. Synthesize an `rtmp://... Input/output error`-shaped line when wf-recorder exits nonzero in streaming mode, so the non-retriable `BadRtmpUrl` classification survives translation.
4. Verify/enforce the output's actual resolution in the shim rather than accepting `--resolution` and ignoring it.
5. Log or assert the effective headless refresh rate (the wlr-randr fallback silently doubles streaming capture cost at 60 Hz).
6. Consider `--audio-codec-param b=128000` (and possibly 44.1 kHz) to match the ffmpeg pipeline, and document that `JIBRI_RECORDING_QUEUE_SIZE` has no effect in Wayland mode.
7. Note the trixie-only dependency floor (wf-recorder ≥ 0.5.0, cage ≥ 0.2.0) in a comment.

Overall: safe to merge as an experiment given the opt-in gating, but I'd treat **recording as the only mode worth A/B-testing initially** — streaming has three independent weaknesses (misclassified errors, no stall coverage, unproven wf-recorder RTMP path) that make it materially riskier than x11grab today.

PR #2300, opened 2026-08-07
Fix standalone acme.sh session using wrong HTTP port

timoschwarzer:
Since the change to use rootless containers, the default HTTP port is 8000 instead of 80.
Retrieving certificates failed because acme.sh was still listening on port 80 so it was never reachable from the internet.

### DIFF
--- a/web/rootfs/etc/s6-overlay/scripts/config
+++ b/web/rootfs/etc/s6-overlay/scripts/config
@@ -54,6 +54,7 @@ if [[ $DISABLE_HTTPS -ne 1 ]]; then
             $ACME_SERVER \
             --issue \
             --standalone \
+            --httpport 8000 \
             --pre-hook "if [[ -d /run/service/nginx ]]; then s6-svc -d /run/service/nginx; fi" \
             --post-hook "if [[ -d /run/service/nginx ]]; then s6-svc -u /run/service/nginx; fi" \
             -d $LETSENCRYPT_DOMAIN


saghul:
Shouldn't it be 8000 so the docker port forward works?

emrahcom:
Hi @timoschwarzer

How do you forward web's https (from 443 to 8443)?
You should do the same for http (from 80 to 8000)

emrahcom:
By the way, hardcoded value may break some deployments which already forward tcp/80 to the internal port

timoschwarzer:
@saghul thank you, yes! I have noticed this mistake and changed it in my testing environment but not in this PR 🤦 will update this shortly. (Edit: fixed)

@emrahcom In the docker-compose.yml port 8000 is hardcoded. https://github.com/jitsi/docker-jitsi-meet/blob/44002b930b044e3bb73af7a3addbe23d639faf12/docker-compose.yml#L11

timoschwarzer:
@emrahcom 

I'm trying to understand what scenario you are describing with

> By the way, hardcoded value may break some deployments which already forward tcp/80 to the internal port

to no avail.

In my deployment I am redirecting port 80 to the hardcoded internal port 8000. Let's Encrypt and other ACME services will *always* use port 80. By default, the standalone server of acme.sh also listens on port 80. But since port 80 of the host is mapped to port 8000 of the container, we need to tell acme.sh that it needs to listen on port 8000.

From what I can see, fetching certificates cannot work under any circumstances on the current master branch since the switch to rootless containers, which changed the internal port from 80 to 8000.

saghul (approved):

garloff:
Just as a sidenote: I ran into this exact same issue and spent more than an hour debugging.
I arrived at the same fix.
Thanks for tracking it down and fixing it, @timoschwarzer.

I think it warrants a stable-11146-2. Just my 0.02€.

PR #2303, opened 2026-08-10
Add distributed tracing setup

JulianGCalderon:
This PR adds support for enabling distributed tracing in a docker environment, by using the features added in:

- jitsi-xmpp-extensions - https://github.com/jitsi/jitsi-xmpp-extensions/pull/146
- jitsi-meet - https://github.com/jitsi/jitsi-meet/pull/17621
- jitsi-meet-lib - https://github.com/jitsi/lib-jitsi-meet/pull/3067
- jicoco - https://github.com/jitsi/jicoco/pull/240
- jitsi-videobridge - https://github.com/jitsi/jitsi-videobridge/pull/2433
- jicofo - https://github.com/jitsi/jicofo/pull/1294.

As there is no stable release with some of these changes, you need to build the images with:
```
JITSI_RELEASE=unstable make
```

Then follow `tracing/README.md` in order to test everything locally.


jicofo/rootfs/defaults/jicofo.conf
```diff
@@ -348,4 +349,12 @@ jicofo {
     }
 }
 
+{{ if $ENABLE_TRACING }}
+tracing {
+    enabled = true
+    otlp-protocol = "grpc"
+    otlp-endpoint = "http://alloy:4317"
```
bgrozev: We want to allow users to plug in their own proto and endpoint. See for example statsd in jibri:
```
{{ $STATSD_HOST := .Env.JIBRI_STATSD_HOST | default "localhost" -}}

```

And the same applies to JVB and prosody.

tracing.yml
```diff
@@ -0,0 +1,48 @@
...
+            - ALLOY_OTLP_UPSTREAM=tempo:4317
+        ports:
+            - "12345:12345"
+            - "4317:4317"
+        depends_on:
+            - tempo
+        networks:
+            meet.jitsi:
+
+    grafana:
```
bgrozev: We already have grafana.yml for the log analyser that uses port 3000, and I think now `grafana.yml` and `tracing.yml` clash. Ideally we'd re-use the existing one and have the tracing resources (dashboard) copied into it. But if that doesn't work for some reason you can just rename and use a new port.

bgrozev (changes requested):
Great stuff, runs successfully! See a couple of comments

tracing.yml
JulianGCalderon: Done!

jicofo/rootfs/defaults/jicofo.conf
JulianGCalderon: Done!

aaronkvanmeerten (commented):
Thanks for this! The core wiring is solid — I verified the generated config against the merged upstream code and the top-level `tracing { enabled / otlp-protocol / otlp-endpoint }` HOCON block matches jicoco's `TracingConfig` exactly, prosody's `otlp_endpoint = ".../v1/traces"` matches `mod_trace`, and enabling `"trace"` only on the internal MUC is correct (it only hooks colibri2 `conference-modify` IQs). The observability overlay has a few dead-on-arrival pieces though — inline comments below.

Two items with no diff line to attach to:

1. **`env.example` isn't updated** — `ENABLE_TRACING`, `TRACING_ENDPOINT`, `TRACING_PROTOCOL`, and `TRACING_OTLP_ENDPOINT` should be documented there (with a note that the feature currently requires `unstable` images; on stable images prosody errors loading the nonexistent `trace` module and jicofo/jvb silently ignore the `tracing{}` block).
2. **Tempo has no persistent volume** — `/var/tempo` (WAL + blocks) lives in the container filesystem, so all traces are lost on any container recreate. The sibling overlays persist their data (`log-analyser.yml`, `grafana.yml`).

I'll follow up with a PR against this branch with suggested fixes for all of the above.

docker-compose.yml
```diff
@@ -232,6 +232,7 @@ services:
             - ENABLE_VISITORS
             - ENABLE_XMPP_WEBSOCKET
             - ENABLE_JAAS_COMPONENTS
+            - ENABLE_TRACING
```
aaronkvanmeerten, line 235: Only `ENABLE_TRACING` is passed through to the containers here (and in the jicofo/jvb lists below), but the templates also read `.Env.TRACING_ENDPOINT` / `.Env.TRACING_PROTOCOL` (jicofo, jvb) and `.Env.TRACING_OTLP_ENDPOINT` (prosody). Since compose only forwards whitelisted variables, setting e.g. `TRACING_ENDPOINT=http://my-collector:4317` in `.env` is silently ignored and the containers always render the hardcoded `http://alloy:4317`/`4318` defaults — the endpoint is effectively unconfigurable. These variables need to be added to the respective services' `environment` lists.

prosody/rootfs/defaults/prosody.cfg.lua
```diff
@@ -49,6 +49,8 @@
 {{ $XMPP_MUC_DOMAIN := .Env.XMPP_MUC_DOMAIN | default "muc.meet.jitsi" -}}
 {{ $XMPP_PORT := .Env.XMPP_PORT | default "5222" -}}
 {{ $XMPP_HIDDEN_DOMAIN := .Env.XMPP_HIDDEN_DOMAIN | default "hidden.meet.jitsi" -}}
+{{ $ENABLE_TRACING := .Env.ENABLE_TRACING | default "0" | toBool -}}
+{{ $TRACING_ENDPOINT := .Env.TRACING_OTLP_ENDPOINT | default "http://alloy:4318" -}}
```
aaronkvanmeerten: Naming inconsistency: prosody reads `TRACING_OTLP_ENDPOINT` while jicofo/jvb read `TRACING_ENDPOINT` for the same concept. An operator pointing the deployment at an external collector has to know to set two differently-named variables, and setting only one silently drops the other component's spans. Since prosody's `mod_trace` only speaks OTLP/HTTP (with `/v1/traces` appended) while the Java stack defaults to gRPC, the two knobs can't be merged into one — but they should at least share a consistent naming scheme and be documented together in `env.example`.

log-analyser/grafana-provisioning/datasources/tempo.yaml
```diff
@@ -0,0 +1,21 @@
...
+  basicAuth: false
+  isDefault: true
+  version: 1
+  editable: false
+  apiVersion: 1
+  uid: tempo
+  jsonData:
+    httpMethod: GET
+    serviceMap:
+      datasourceUid: prometheus
```
aaronkvanmeerten: `serviceMap.datasourceUid: prometheus` references a datasource uid that doesn't exist in this stack — the provisioned Prometheus datasource in `datasource_loki.yml` has uid `d301145e-8c4e-4027-bf6e-43e81f095020`. As shipped, the Tempo datasource's Service Graph tab fails with "data source not found".

Also, `isDefault: true` displaces the org default datasource for existing deployments (none of the currently provisioned datasources claim default) — probably better left unset.

tracing/tempo/tempo.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+metrics_generator:
+  registry:
+    external_labels:
+      source: tempo
+      cluster: docker-compose
+  storage:
+    path: /var/tempo/generator/wal
+    remote_write:
+      - url: http://prometheus:9090/api/v1/write
```
aaronkvanmeerten, line 24: This remote-write target can never work as documented: the README's compose command doesn't include `prometheus.yml` (so the `prometheus` host doesn't resolve), and even when that overlay is added, it starts Prometheus without `--web.enable-remote-write-receiver`, so the write endpoint returns 404. Tempo will log continuous remote-write failures and the span-metrics/service-graphs (which the service map in the Tempo datasource depends on) are never produced. Suggest adding the flag to `prometheus.yml` and including it in the README's compose command.

log-analyser/grafana-provisioning/dashboards/tracing.json
```diff
@@ -0,0 +1,173 @@
...
+      "description": "Rate of participants joining",
+      "gridPos": { "h": 8, "w": 8, "x": 0, "y": 1 },
+      "datasource": { "type": "tempo", "uid": "${datasource}" },
+      "fieldConfig": { "defaults": { "unit": "reqps" }, "overrides": [] },
+      "targets": [
+        {
+          "refId": "A",
+          "queryType": "traceql",
+          "datasource": { "type": "tempo", "uid": "${datasource}" },
+          "query": "{ name = \"conference.member-joined\" } | rate()"
```
aaronkvanmeerten, line 41: All the timeseries panels in this dashboard use TraceQL metrics queries (`| rate()`, `| quantile_over_time(...)`), but `grafana.yml` pins `grafana/grafana:10.2.0`, and TraceQL metrics queries require Grafana ≥ 10.4 (shipped alongside Tempo 2.4, March 2024). Following the README as written provisions a dashboard where every metric panel errors — the Grafana pin needs a bump for this to work.

log-analyser/grafana-provisioning/dashboards/tracing.json
```diff
@@ -0,0 +1,173 @@
...
+    { "id": 101, "type": "row", "title": "Latency", "gridPos": { "h": 1, "w": 24, "x": 0, "y": 17 }, "collapsed": false, "panels": [] },
+
+    {
+      "id": 5,
+      "type": "timeseries",
+      "title": "colibri.allocate duration (p50/p95)",
+      "description": "JVB resource allocation latency",
+      "gridPos": { "h": 8, "w": 12, "x": 0, "y": 18 },
+      "datasource": { "type": "tempo", "uid": "${datasource}" },
+      "fieldConfig": { "defaults": { "unit": "ns" }, "overrides": [] },
```
aaronkvanmeerten: `quantile_over_time(duration, ...)` returns values in **seconds**, not nanoseconds — Grafana's own trace-based-alerting docs threshold these results directly in seconds. With `"unit": "ns"` (here and on the participant-invite panel at line 124), a real p95 of 1.2 s renders as "1.20 ns". Should be `"unit": "s"`.

tracing.yml
```diff
@@ -0,0 +1,33 @@
...
+        networks:
+            meet.jitsi:
+
+
+    alloy:
+        image: grafana/alloy:latest
+        restart: ${RESTART_POLICY:-unless-stopped}
+        volumes:
+            - ./tracing/alloy/:/etc/alloy/
+            - /var/run/docker.sock:/var/run/docker.sock
```
aaronkvanmeerten: This Docker socket mount is unused — `tracing/alloy/config.alloy` contains only an OTLP receiver and exporter, no Docker discovery/logging component. Mounting the socket grants root-equivalent control of the host to the alloy container (on an unpinned `:latest` image) for no functional benefit. Suggest dropping it.

tracing.yml
```diff
@@ -0,0 +1,33 @@
...
+        command:
+            - run
+            - /etc/alloy/config.alloy
+            - --server.http.listen-addr=0.0.0.0:12345
+        environment:
+            - ALLOY_OTLP_UPSTREAM=tempo:4317
+        ports:
+            - "12345:12345"
+            - "4317:4317"
```
aaronkvanmeerten: None of the host port publishes in this file (`3200` for Tempo's query API, `12345` for Alloy's admin UI, `4317` for OTLP ingest) are needed for the documented workflow: Grafana reaches Tempo via the internal `http://tempo:3200`, and all trace producers (jicofo, jvb, prosody) send to alloy over the `meet.jitsi` network. Notably, the browser does **no** OTLP export — lib-jitsi-meet only echoes a `traceparent` Jingle extension — so there's no host-side consumer for 4317 either. On an internet-facing deployment these expose unauthenticated trace query/ingest and an admin UI to the world. Suggest removing the `ports` sections entirely.

Related hygiene: `grafana/tempo:latest` and `grafana/alloy:latest` are unpinned while every neighbouring overlay pins versions — worth pinning for reproducibility.

aaronkvanmeerten:
As promised in the review: I've opened JulianGCalderon/jitsi-docker#2 against the `tracing` branch with suggested fixes for all of the points above (env var pass-through, env.example docs, Grafana version bump, service-map datasource uid, Prometheus remote-write receiver, panel units, and tracing.yml hygiene). Merging that into your branch will update this PR.

aaronkvanmeerten (approved):

PR #2306, opened 2026-08-17
feat(jibri) add support for running Chrome on arm64

### LINKED ISSUE #2298: Switch to Google Chrome arm64
It's almost there: https://issues.chromium.org/issues/374811603?pli=1

Google Chrome 151 is already available as a deb package for Debian / Ubuntu arm64: https://dl.google.com/linux/direct/google-chrome-stable_current_arm64.deb

The matching ChromeDriver is not ready yet, however. It should show up here when ready: https://googlechromelabs.github.io/chrome-for-testing/#stable

## Why switch to it when Debian's Chromium Just Works (TM)?

- Consistency across platforms
- Ability to select the exact upstream version we want for a given image

saghul:
- Switch to installing Chrome-for-Testing versions
- Drop support for version being "latest", one must be specified
- Add support for testing any versions

Fixes: https://github.com/jitsi/docker-jitsi-meet/issues/2298

### DIFF
--- a/jibri/Dockerfile
+++ b/jibri/Dockerfile
@@ -11,9 +11,9 @@ LABEL org.opencontainers.image.source="https://github.com/jitsi/docker-jitsi-mee
 LABEL org.opencontainers.image.documentation="https://jitsi.github.io/handbook/"
 
 ARG USE_CHROMIUM=0
-#ARG CHROME_RELEASE=latest
 # https://googlechromelabs.github.io/chrome-for-testing/
-ARG CHROME_RELEASE=149.0.7827.114
+ARG USE_CHROME_CANARY=1
+ARG CHROME_RELEASE=153.0.8009.0
 
 COPY rootfs/ /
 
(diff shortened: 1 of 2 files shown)


jibri/Dockerfile
```diff
@@ -11,9 +11,9 @@ LABEL org.opencontainers.image.source="https://github.com/jitsi/docker-jitsi-mee
 LABEL org.opencontainers.image.documentation="https://jitsi.github.io/handbook/"
 
 ARG USE_CHROMIUM=0
-#ARG CHROME_RELEASE=latest
 # https://googlechromelabs.github.io/chrome-for-testing/
-ARG CHROME_RELEASE=149.0.7827.114
+ARG USE_CHROME_CANARY=1
```
saghul: Note: right now there is no linux-arm64 stable ChromeDriver released, so this is necessary to test it. My plan is to land this with it defaulting to 0, once 153 is stable. I'll also update CHROME_RELEASE to match the stable release version.

saghul:
This one is now ready to go!

I tested it and it's working great on macOS with an arm64 guest.

I went one step ahead and switched to using CfT instead of the Chrome Debian repo, which now allows us to use stable, beta, dev or canary versions easily and there is no risk of having any drift between the Chrome and ChromeDriver versions.

aaronkvanmeerten (approved):

PR #2320, opened 2026-09-08
jigasi: bind REST API to loopback by default (JIGASI_REST_HOST opt-in)

Sumit-Sabhaya:
Ref: reported to 8x8's HackerOne program as HackerOne #3996571 ("jitsi/jigasi: Unauthenticated REST API enables unauthorized outbound SIP dialing (toll fraud)"), closed as Informative on 2026-09-08. This PR is a defense-in-depth fix for part of that finding.

# jigasi: bind REST API to loopback by default; 0.0.0.0 becomes explicit opt-in

## Summary

When `JIGASI_ENABLE_REST=true` is set, this template writes
`org.jitsi.jigasi.rest.jetty.host=0.0.0.0`, exposing the Jigasi REST API on
the whole docker network. This PR makes loopback the default for that mode
and adds an explicit `JIGASI_REST_HOST` variable for operators who actually
need a non-loopback bind.

Ref: reported to 8x8's HackerOne program as **HackerOne #3996571**
("jitsi/jigasi: Unauthenticated REST API enables unauthorized outbound SIP
dialing (toll fraud)"), closed as Informative on 2026-09-08. This PR is a
defense-in-depth fix for part of that finding.

## Why this matters

- Since jitsi/jigasi commit `31ab705` (PR jitsi/jigasi#601, "fix: Sets
  default host for the rest plugin", 2025-06-03) the upstream default for
  the REST host is `127.0.0.1`.
- jicoco's `JettyBundleActivatorConfig` resolves the legacy property
  `org.jitsi.jigasi.rest.jetty.host` **first** and only falls back to the
  HOCON default when the legacy key is absent. The line this PR removes is
  therefore not just redundant — it silently **overrides** the upstream
  loopback fix in the exact deployment mode the docker README documents.
- The REST API (`/configure/call-control-muc/*`, `/shutdown`, `/about/*`)
  carries no authentication. Combined with a `0.0.0.0` bind, any other
  container on the overlay network can e.g. inject live XMPP call-control
  accounts via unauthenticated `POST /configure/call-control-muc/add`.
- The documented consumer of REST mode (the autoscaler sidecar reading
  `http://localhost:8788/about/stats`) shares the container's network
  namespace and only needs loopback — `0.0.0.0` is not required by that
  use case.

## Change

- Default REST bind in REST-enabled mode becomes `127.0.0.1` (matching the
  upstream default and also pinning pre-31ab705 jigasi builds to loopback).
- Operators who need external reachability can set `JIGASI_REST_HOST`
  (e.g. `JIGASI_REST_HOST=0.0.0.0` to restore current behavior).

## Verification

```bash
JIGASI_ENABLE_REST=true docker compose up -d jigasi

# inside the jigasi container
docker compose exec jigasi ss -ltnp | grep 8788
#   expected: 127.0.0.1:8788  (previously 0.0.0.0:8788)

# from another container on the same network (name may vary by compose project)
docker run --rm --network <project>_meet.jitsi curlimages/curl:latest \
  -s -m 3 http://jigasi:8788/about/version || echo "connection refused (expected)"

# localhost inside the container still works (sidecar/operator use case)
docker compose exec jigasi curl -s http://127.0.0.1:8788/about/version
```

## Compatibility

Backward compatible by default: with no `JIGASI_REST_HOST` set, the only
behavior change is that the REST API is no longer reachable from other
containers unless explicitly opted in — which is the point of the change.

If this repo maintains a public env reference (e.g. `env.example` /
`.env` docs), `JIGASI_REST_HOST` should be documented there as the explicit
opt-in for non-loopback binds.


saghul:
Not sure how much sense it makes to bind to localhost in a container scenario. Nobody will be able to connect, so what's the plan there?

I'd say the container should bind to 0.0.0.0 but then the jigasi.yaml should expose that port only on localhost.

What are you trying to protect yourself against? Is it rogue containers on your own network?!

Sumit-Sabhaya:
Thanks for the review — responding to your three points directly.

**On "nobody will be able to connect":** the shipped compose stack doesn't publish 8788 to the host today (`jigasi.yml` only publishes the UDP media range), so a loopback default removes no existing host-side consumer. The consumers this mode exists for are co-located with the container:
- `docker compose exec` for operator tooling (`reconfigure_xmpp.sh`)
- the autoscaler-sidecar, whose own docs retrieve stats from a *localhost* URL on the local instance (`STATS_RETRIEVE_URL`)

Sidecars share the instance's network namespace, so loopback keeps them working. Non-loopback stays available via the new `JIGASI_REST_HOST`.

**On 0.0.0.0 in-container + restricting in `jigasi.yml`:** restricting the *published* port only shapes host access. On a docker bridge network there's no filtering between containers — every service on `meet.jitsi` reaches every other service's internal ports regardless of `jigasi.yml`, and jigasi can't be moved off that network (XMPP + media). The bind address is the only per-service control that actually scopes who can reach 8788.

The core point: upstream jigasi already chose loopback as the default for this API in `31ab705` ([jigasi#601](https://github.com/jitsi/jigasi/pull/601)), and this template's line *overrides* that default — jicoco resolves the legacy `org.jitsi.jigasi.rest.jetty.host` property before the HOCON value. So with REST enabled, operators relying on the upstream fix silently get `0.0.0.0`. This PR is defense-in-depth: stopping this repo from re-widening a control surface upstream already defaults to loopback.

**On the threat model:** not hypothetical rogue containers; lateral movement. Any single compromised service on the shared network (the web container is internet-facing) can POST `/configure/call-control-muc/add`, which injects a live XMPP call-control account and lets the gateway originate calls through the operator's SIP trunk — plus `/shutdown` and unauthenticated account add/remove/list. That's a write + billing path, not read-only stats.

Agreed that JVB's Colibri REST is a similar exposure class and deserves a separate look — this PR only addresses the template override.

Happy to adjust scope — e.g. a companion issue for real auth on this API, since a bind default is hardening, not the full fix.

saghul:
If I wanted to read AI ouput I'd ask it myself. 

> Any single compromised service on the shared network (the web container is internet-facing) can POST

This makes no sense. If your infra is already compromised you are not gaining anything from protecting jigasi here. This is security theater.

Sumit-Sabhaya:
Fair enough, thanks for the review. I'll close this and open an issue for auth on the REST API instead, since that's the real fix anyway.

PR #2326, opened 2026-09-12
feat(opus-transcriber-proxy): add a new container

emrahcom:
Built from source on the base image rather than using the published `jitsi/opus-transcriber-proxy` image, which is alpine, runs as root and has no s6. The first-party services here are Debian, rootless and read-only. So, the proxy is built the same way.

There is no Debian package for it and the last tag predates the code on main. So, the build tracks the main branch. `OPUS_TRANSCRIBER_PROXY_VERSION` takes a tag instead when that is wanted.

Only the native `libopus` backend is built. The WASM one is upstream's default because they also target Cloudflare Workers, which cannot load a native addon. We only ship the container. So one backend is enough and native is the faster of the two.


emrahcom:
I tested with these additional env variables and it works:

```
ENABLE_TRANSCRIPTIONS=1    
PROVIDERS_PRIORITY=deepgram
DEEPGRAM_API_KEY=my-api-key
XMPP_MUC_MODULES=force_async_transcription
```

emrahcom:
any feedback?

saghul:
Hey @emrahcom sorrybdor the delay, it's been a bit busy. I hope to take a look towards the end of this week.

